PR #63, opened 2026-05-19
Add typed errors for agent core and key runtime paths

Bhavesh-0409:
## Summary

Adds shared typed errors to `@tapflow/agent-core` and replaces key generic `Error` throw sites with `ValidationError` and `PlatformError` so consumers can handle failures with `instanceof`.

## Changes

- Added `ValidationError`, `PlatformError`, and `AuthError`
- Exported them from `@tapflow/agent-core`
- Updated key throw sites in `agent-core`, `android-agent`, `ios-agent`, and `relay`
- Added focused test coverage for typed error handling
- Fixed workspace TypeScript package resolution for local typecheck by adding `types` entries to package export maps

## Verification

- `pnpm typecheck`

## Summary by CodeRabbit

* **New Features**
  * Added base TapflowError and new error classes (ValidationError, PlatformError, AuthError); errors are re-exported. Added a throwing JWT verifier (verifyJwtOrThrow).
  * Package manifests now include explicit TypeScript type entrypoints for improved type resolution.

* **Bug Fixes**
  * Replaced many generic throws with specific error types to improve error clarity across agents, session handling, DB access, and streams.

* **Tests**
  * Updated and added tests to assert the new error types and messages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jo-duchan/tapflow/pull/63?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/agent-core/src/__tests__/AgentRegistry.test.ts
+++ b/packages/agent-core/src/__tests__/AgentRegistry.test.ts
@@ -36,8 +36,16 @@ describe('AgentRegistry', () => {
   })
 
   it('throws when platform is not registered', () => {
-    expect(() => AgentRegistry.get('unknown')).toThrow(PlatformError)
-    expect(() => AgentRegistry.get('unknown')).toThrow('No agent registered for platform: unknown')
+    let thrown: unknown
+
+    try {
+      AgentRegistry.get('unknown')
+    } catch (error) {
+      thrown = error
+    }
+
+    expect(thrown).toBeInstanceOf(PlatformError)
+    expect((thrown as Error).message).toBe('No agent registered for platform: unknown')
   })
 
 
--- a/packages/agent-core/src/__tests__/AgentRegistry.test.ts
+++ b/packages/agent-core/src/__tests__/AgentRegistry.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach } from 'vitest'
+import { PlatformError } from '../index'
 import { AgentRegistry } from '../AgentRegistry'
 import type { DeviceAgent } from '../DeviceAgent'
 import type { Device } from '../types'
@@ -35,10 +36,10 @@ describe('AgentRegistry', () => {
   })
 
   it('throws when platform is not registered', () => {
-    expect(() => AgentRegistry.get('unknown')).toThrow(
-      'No agent registered for platform: unknown'
-    )
+    expect(() => AgentRegistry.get('unknown')).toThrow(PlatformError)
+    expect(() => AgentRegistry.get('unknown')).toThrow('No agent registered for platform: unknown')
   })
+
 
   it('resets cached instance when re-registering', () => {
     AgentRegistry.register('mock', MockAgent)
--- a/packages/agent-core/src/errors.ts
+++ b/packages/agent-core/src/errors.ts
@@ -1,0 +1,11 @@
+class TapflowError extends Error {
+  constructor(message: string, options?: ErrorOptions) {
+    super(message, options)
+    this.name = new.target.name
+    Object.setPrototypeOf(this, new.target.prototype)
+  }
+}
+
+export class ValidationError extends TapflowError {}
+
+export class PlatformError extends TapflowError {}
--- a/packages/agent-core/src/errors.ts
+++ b/packages/agent-core/src/errors.ts
@@ -9,3 +9,5 @@ class TapflowError extends Error {
 export class ValidationError extends TapflowError {}
 
 export class PlatformError extends TapflowError {}
+
+export class AuthError extends TapflowError {}
--- a/packages/android-agent/src/scrcpy/ScrcpySession.ts
+++ b/packages/android-agent/src/scrcpy/ScrcpySession.ts
@@ -8,7 +8,7 @@ import { promisify } from 'util'
 import { ScrcpyVideo } from './ScrcpyVideo.js'
 import { ScrcpyControl } from './ScrcpyControl.js'
 import type { ScrcpyDeviceInfo } from './ScrcpyVideo.js'
-import { createLogger } from '@tapflowio/agent-core'
+import { createLogger, PlatformError, ValidationError } from '@tapflowio/agent-core'
 
 const logger = createLogger('android-agent:scrcpy')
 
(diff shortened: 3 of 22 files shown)


Bhavesh-0409:
Please review my PR

packages/android-agent/src/scrcpy/ScrcpySession.ts
```diff
@@ -54,12 +54,12 @@ export class ScrcpySession {
   }
 
   get video(): ScrcpyVideo {
```
jo-duchan, line 56: `ValidationError` is for user-input validation failures. Accessing `.video` or `.control` before `start()` is an internal state invariant violation — `PlatformError` is the right type here. Same applies to the `.control` getter below. The whole point of typed errors is that callers can distinguish causes via `instanceof`, so mixing semantics breaks that.

packages/agent-core/src/__tests__/AgentRegistry.test.ts
```diff
@@ -35,9 +36,25 @@ describe('AgentRegistry', () => {
...
+    }
+
+    throw new Error('expected AgentRegistry.get() to throw')
+  })
+
+  it('exports all typed errors from the public package entry', () => {
+    expect(new ValidationError('bad input')).toBeInstanceOf(Error)
+    expect(new PlatformError('platform failed')).toBeInstanceOf(Error)
+    expect(new AuthError('unauthorized')).toBeInstanceOf(Error)
   })
```
jo-duchan, line 51: This test is redundant — `toThrow(PlatformError)` in the test above already performs an `instanceof` check internally. The try/catch pattern here doesn't add coverage. Can remove this entire `it` block.

packages/agent-core/src/errors.ts
```diff
@@ -0,0 +1,13 @@
...
+    this.name = new.target.name
+    Object.setPrototypeOf(this, new.target.prototype)
+  }
+}
+
+export class ValidationError extends TapflowError {}
+
+export class PlatformError extends TapflowError {}
+
+export class AuthError extends TapflowError {}
```
jo-duchan, line 13: `AuthError` is exported but has no throw sites anywhere in the codebase. An unused public export is confusing for consumers — would you be up for following this up with a separate ticket for typed errors in the relay's JWT/auth validation paths? That seems like a natural next step and would give `AuthError` a real home. For now, consider removing it from this PR to keep the scope clean.

jo-duchan (changes requested):
Overall direction is solid — the base class implementation in `errors.ts` is correct (`new.target.name`, `Object.setPrototypeOf`, `cause` support). A few things to address before merging.

jo-duchan:
Hey @Bhavesh-0409 — heads up that we're merging a scope rename PR (#64) before this one: all internal packages are moving from `@tapflow/*` to `@tapflowio/*` (npm org name conflict). Once that lands, please rebase this branch and update any `@tapflow/agent-core` imports to `@tapflowio/agent-core`. Thanks!

Bhavesh-0409:
I tried pushing after making necessary changes , can you review

jo-duchan:
## Code Review

Thanks for the PR! The `Object.setPrototypeOf` pattern and `new.target.name` usage are correct — these are required for proper `instanceof` behavior in transpiled environments, and the error chaining looks good too.

A few things to address before merge:

---

### 🔴 Critical — `AuthError` is missing

The PR description and ROADMAP both list `AuthError` as one of the error classes to introduce, but it's not in the diff. The `TapflowError` base + `ValidationError` / `PlatformError` / `NetworkError` are present, but `AuthError` is absent entirely.

Please add `AuthError` (e.g. for unauthorized / token-expired scenarios) before merge so the typed-error surface is complete.

---

### 🟡 Unintended change — em dash regression in `ScreenCaptureStreamer.ts`

The original code uses an em dash (`—`) in a comment or string, but this PR changes it to a hyphen (`-`). This looks unintentional and should be reverted to stay within the minimal-change principle.

---

### 🟡 Wrong error type — `Device not found` in `AndroidAgent.ts`

```ts
throw new ValidationError('Device not found')
```

`ValidationError` implies bad user input. "Device not found" is a runtime/platform state issue — `PlatformError` is the semantically correct type here (same as how iOS agent handles it).

---

### 🟡 Redundant call in tests

One test calls the same function twice in sequence without asserting intermediate state. The second call is unnecessary — remove it to keep tests focused.

---

### Summary

| | |
|---|---|
| `AuthError` missing | Must add before merge |
| em dash → hyphen | Revert unintended change |
| `ValidationError` for "Device not found" | Change to `PlatformError` |
| Duplicate call in test | Remove redundant call |

jo-duchan:
Hi @Bhavesh-0409 — thanks for the contribution!

We're proceeding with the stable 0.1.0 release now, so this PR won't make the cut for that milestone. Once you've addressed the review feedback (especially the missing `AuthError`), please request a re-review and we'll target it for **0.1.1**. Looking forward to getting it merged then!

Bhavesh-0409:
Addressed the requested feedback:

- restored `AuthError` and added a real auth/JWT throw path
- changed Android `Device not found` to `PlatformError`
- restored the em dash in `ScreenCaptureStreamer`
- removed the redundant extra invocation in the `AgentRegistry` test

Requesting re-review for `0.1.1`.

jo-duchan:
Good progress — `AuthError` is now present and `verifyJwtOrThrow` is a clean addition. Most of the other feedback has been addressed.

Two items still remaining:

---

### 🟡 `IOSAgent.ts` — `Device not found` inconsistency

```ts
// IOSAgent.ts
if (!target) throw new ValidationError(`Device not found: ${deviceId}`)

// AndroidAgent.ts (same scenario)
if (!target) throw new PlatformError(`Device not found: ${avdId}`)
```

"Device not found at runtime" is a platform/environment state issue, not a user input validation error — `PlatformError` is the correct type here, same as AndroidAgent uses. Please align IOSAgent to match.

---

### 🟡 `SessionManager.test.ts` — duplicate `sm.join()` calls

```ts
expect(() => sm.join('bad-id', mockSocket())).toThrow(ValidationError)
expect(() => sm.join('bad-id', mockSocket())).toThrow('Session not found: bad-id')
```

This calls `sm.join()` twice. Combine into one:

```ts
let thrown: unknown
try { sm.join('bad-id', mockSocket()) } catch (e) { thrown = e }
expect(thrown).toBeInstanceOf(ValidationError)
expect((thrown as Error).message).toBe('Session not found: bad-id')
```

Same pattern applies to the "session busy" test below it.

---

Fix these two and it's ready for 0.1.1!

Bhavesh-0409:
Addressed the last two comments:

- aligned iOS Device not found to PlatformError
- removed duplicate sm.join() invocations in SessionManager tests

Requesting final re-review for 0.1.1.

jo-duchan (approved):
All feedback addressed — looks good to merge! 🎉

Also, welcome and thank you for your first contribution to tapflow! The typed error system you added makes the library significantly more useful for consumers who need to handle failures programmatically.

If you find tapflow useful or interesting, a ⭐ on the repo would mean a lot to us and helps others discover the project!

PR #275, opened 2026-06-12
Harden relay configuration for public and proxied exposure

jo-duchan:
## Summary

Strengthens the relay for deployments that are exposed off-host or run behind a reverse proxy. Configuration-level hardening only — no protocol or public API changes.

## Changes

- **JWT secret**: generate and persist a per-install secret when `JWT_SECRET` is unset, replacing the shared development default. No action needed for a single relay.
- **Authentication**: rate-limit login attempts with exponential backoff (per IP + account); fail safely on malformed stored password hashes.
- **Bootstrap**: restrict `auth/init` to localhost. Headless servers run `tapflow admin init` on the relay host (existing fallback path).
- **Trusted proxy**: new `TAPFLOW_TRUSTED_PROXIES` resolves the real client IP from `X-Forwarded-For` when behind a same-host reverse proxy. Disabled by default — no behavior change unless set.

## Compatibility

- No WebSocket protocol, public API signature, CLI flag, or DB schema changes. New handler parameters are optional with defaults.
- One user-visible change: remote-browser `/setup` bootstrap is now localhost-only. Documented in the self-hosting guide; `tapflow admin init` covers headless servers.

## Tests

New: clientAddress (14), init-localhost (8), rateLimit (8), login-hardening (7), jwtSecret (6). Full relay suite: 238 passing.

## Docs

configuration.md and self-hosting.md (EN + KO) updated for the per-install secret and `TAPFLOW_TRUSTED_PROXIES`.

## Release Notes

* **New Features**
  * Automatic per-install JWT secret generation and persistence when unset on first boot
  * Rate limiting with exponential backoff on authentication endpoints to prevent brute force attacks
  * Localhost-only restriction for the initialization endpoint
  * Support for trusted reverse proxy configuration via TAPFLOW_TRUSTED_PROXIES environment variable

* **Documentation**
  * Updated configuration reference and guides for JWT secret behavior and trusted proxy setup

### DIFF
--- a/.changeset/relay-exposure-hardening.md
+++ b/.changeset/relay-exposure-hardening.md
@@ -1,0 +1,11 @@
+---
+"tapflow": patch
+"@tapflowio/relay": patch
+---
+
+Harden the relay for public and proxied exposure:
+
+- A per-install JWT secret is generated and persisted automatically when `JWT_SECRET` is unset, replacing the shared development default.
+- Authentication endpoints apply rate limiting with exponential backoff.
+- Bootstrap (`auth/init`) is restricted to localhost — on headless servers, run `tapflow admin init` on the relay host.
+- New `TAPFLOW_TRUSTED_PROXIES` resolves the real client IP from `X-Forwarded-For` when the relay runs behind a same-host reverse proxy.
--- a/.gitignore
+++ b/.gitignore
@@ -45,8 +45,8 @@ tapflow.config.json
 *.db-shm
 *.db-wal
 
-# playground runtime data (recordings, uploads, db)
-playground/.tapflow-data/
+# runtime data (recordings, uploads, db, generated jwt-secret) — any package/cwd
+.tapflow-data/
 
 # playground test builds
 playground/test-builds/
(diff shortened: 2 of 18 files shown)


jo-duchan:
## CodeRabbit review responses

Status of each finding:

- **`clientAddress.ts` — parse X-Forwarded-For from the right** (Major): Fixed in 66c0d66. XFF is now parsed right-to-left, stripping trusted-proxy hops and using the first untrusted hop. A trusted-proxy socket with no XFF is treated as a direct connection, so host-local `admin init` / CLI / agent still work.
- **`rateLimit.ts` — bound/expire limiter entries** (Major): Fixed in 66c0d66 (per-entry expiry + max-size cap) and dac4776 (LRU re-insertion so actively-locked keys survive overflow eviction).
- **`jwtSecret.test.ts` — gate the 0600 assertion to POSIX** (Minor): Fixed in 9bbfa32.

### Skipping: `RelayServer` `this.options.trustedProxies ?? []`

Intentionally left as-is:

- **Not a real bug** — `server.ts` always injects `new RelayServer({ trustedProxies: config.local.trustedProxies })`, so the production path matches the config value.
- **Falling back to `config.local.trustedProxies` inside the router would break test isolation** — tests that construct `new RelayServer({ port: 0 })` deliberately omit trusted proxies and would start depending on the global `TAPFLOW_TRUSTED_PROXIES` env. `?? []` gives an explicit, isolated default ("not injected → empty").
- `RelayServer` intentionally does not import `config`; `server.ts` is the single wiring point.

Issue #232, opened 2026-06-08
LAN HTTPS (Secure Context) for WebCodecs — DNS-01 primary, local-CA fallback


jo-duchan:
## Background

WebCodecs (HW video decoding) only works in a **Secure Context** (HTTPS or `localhost`). The Mac running tapflow is fine over `localhost`, but **teammates accessing the LAN** (`:4000`) from their browsers are not in a secure context — so they can't use WebCodecs and miss the HW-decode performance win.

We want to make LAN HTTPS as frictionless as possible. **Low priority** — not starting now, just recording the decision.

## Decision

**Primary: DNS-01 mode.**
- Map a subdomain of the team's domain (`tapflow.company.com`) to the LAN private IP, plus a DNS API token.
- Caddy issues + auto-renews certs via Let's Encrypt **DNS-01 challenge**.
- **Zero client friction** — no teammate installs a root CA, no browser warnings.
- Startup teams almost always already own a company domain, making this the best fit for the target user (multiple teammates).

**Why this does not violate the "no external cloud" principle:**
DNS-01 only sends **domain-ownership proof (a TXT record)** outbound. No app stream/data ever leaves the network. Cert issuance is one-time metadata; the video stream stays purely on the LAN.

## Scope (convert to a plan when picked up)

- [ ] Design the Helper as a **thin onboarding wizard + Caddy orchestrator** (heavy CA logic isolated to the fallback path only)
  - [ ] Pick DNS provider → enter API token once
  - [ ] Guide subdomain / LAN IP mapping
  - [ ] Generate + start Caddy config with the DNS plugin (reuse Caddy's built-in issuance/renewal)
- [ ] **Fallback: local-CA mode** — for teams without a domain or air-gapped networks. Only here does the full Helper path run (create CA → register in OS Trust Store → issue leaf certs), i.e. what mkcert does.
- [ ] Docs for hostname resolution: public DNS A-record pointing at a private IP (simplest; minor IP exposure) vs internal DNS

## Out of scope

- **Real devices (QA phones) excluded** for this topic — desktop-browser teammates only. iOS profile / Android user-CA constraints are left as a separate topic.

## Reference tools

- **mkcert** — local CA + trust store. Single-machine; doesn't scale to N clients → the model for the fallback mode.
- **Caddy** — built-in Let's Encrypt (DNS-01) when a domain is available → the primary mode's engine.
[...]

PR #281, opened 2026-06-14
feat(relay): LAN HTTPS cert issuance for secure-context / WebCodecs (#232)

jo-duchan:
## What & why

Browsers on the LAN that open the dashboard over `http://192.168.x.x:4000` are **not a secure context**, so `WebCodecs` (hardware-accelerated decode, #232) is unavailable. This PR adds the prerequisite: a way to serve the relay over **real, publicly-trusted HTTPS** so any teammate's browser gets a secure context.

Key constraint honored: **the media/control stream stays LAN-direct.** The cloud is only touched for DNS resolution and cert issuance/renewal (off the hot path) — never a reverse-proxy hop. So WebCodecs' speed gains are preserved.

## How it works

A pluggable cert/TLS layer in `relay` (`packages/relay/src/lib/cert/`), kept orthogonal to the tunnel/reachability axis:

- **`CertProvider` / `DnsProvider`** interfaces (DIP, hot-updater-style "we provide the structure, you bring your own account").
- **Issuance paths (v1):**
  - `byo-api-token` — ACME DNS-01 via `acme-client`; the user's own DNS account API sets the `_acme-challenge` TXT. Cert private key is generated locally and never leaves the Mac.
  - `import-cert` — bring your own cert/key.
- **`CloudflareDnsProvider`** (DNS-01 TXT/A·AAAA, zone discovery). Token read from env, never config.
- **`DiskCertStore`** (key `0600`) + **renewal scheduler** (renew only within 30 days of expiry; no re-issue on every boot).
- **relay terminates HTTPS + WSS in-process** (`https.createServer` when a cert is present; `updateTlsContext` hot-swaps renewed certs). Bootstrap wired in `server.ts` and `tapflow relay start`.
- **`tapflow init` wizard**: for the LAN path, asks Streaming performance (Standard HTTP/WASM vs High HTTPS/WebCodecs) → cert method (Cloudflare API token | import). Writes the `tls` block to `tapflow.config.json`; tokens stay in env. **Domain-less users pick Standard and use tapflow fully over HTTP — HTTPS is a performance upgrade, not a requirement.**

## Validated

Issued a **real Let's Encrypt _staging_ certificate** end-to-end via DNS-01 (subject `CN=...`, issuer `(STAGING) ...`, ~90-day validity), exercising ACME order → DNS-01 → propagation → LE validation → cert. Unit suites: relay 323 + cli 207 green; lint + typecheck pass across all packages.

A `DesecDnsProvider` exists as a plugin used by the staging smoke test, but **free dynDNS services are intentionally not offered in the wizard** (deletion/inactivity/ToS fragility) — domain-less users use HTTP/WASM instead.

## Out of scope (follow-ups)

- The WebCodecs decoder itself (#232 body) — this PR only enables its prerequisite.
- Dashboard fallback UX (DNS-rebinding guidance) and QA router matrix.
- Additional DNS adapters (Vercel/Route53) and CNAME-delegation path (plugin/community).
- HTTPS for the Tailscale / rathole reachability paths.

## Test plan

See the PR comment / description for manual steps (wizard config, relay TLS via self-signed import-cert, optional real staging issuance).

## Release Notes

* **New Features**
  * Initialization wizard adds interactive HTTPS/TLS setup (Cloudflare or deSEC DNS-01, Vercel DNS-01, or importing an existing certificate/key).
  * `tapflow.config.json` now supports an optional top-level `tls` block; the Relay server can run over HTTPS/WSS and hot-reload certificates on renewal.
* **Tests**
  * Added unit, contract, and integration tests covering TLS termination, ACME issuance/renewal, and DNS provider TXT/A/AAAA behaviors.
* **Chores**
  * Added automated certificate issuance/renewal support for TLS configurations.


jo-duchan:
Thanks @coderabbitai — addressed in `63ef7a4`:

**Fixed**
- **Single-flight issuance** (`AcmeCertProvider`): concurrent `ensureCert`/`renewIfNeeded` no longer race into duplicate ACME orders; added a concurrency test.
- **Request timeouts**: `FetchLike` gained `signal`; all DNS adapters (Cloudflare/deSEC/Vercel) now use `AbortSignal.timeout(30s)` so a stalled DNS API can't hang issuance/renewal.
- **Config fail-fast**: missing `tls.dnsProvider` now fails validation instead of silently defaulting to `cloudflare` (misspelled was already rejected by the zod enum).
- **Vercel pagination**: domain & record listing now follow `pagination.next` (>100 items) — avoids false "No Vercel domain found" / broken idempotency; added a pagination test.
- **Deterministic test expiry**: replaced `> Date.now()` fixture-expiry assertions (diskCertStore / importCert / parseCert) with parse-based / not-NaN assertions so they won't break when the fixture cert expires.
- **CLI startup line**: `relay start` now prints the `http(s)` scheme, not `ws(s)`.
- **Comments**: trimmed `DnsProvider.ts` to one-line WHY notes.

**Deferred (separate PR — pre-existing, out of scope here)**
- `relay-start.ts` not passing `trustedProxies`/`corsOrigins` to `RelayServer`: this divergence predates this PR (only the `tls` arg was added here) and is unrelated to LAN HTTPS. Tracked as a follow-up.

relay 336 + cli 208 green.

jo-duchan:
Second review round addressed in `e3cb95d`:

- **ACME account key persistence (Medium)** — now cached at `tls/account.pem` (0600) and reused across restarts/renewals via `loadOrCreateAccountKey` + `AcmeClientIssuer.accountKeyPath`, so we no longer register a fresh LE account per issuance (avoids the Accounts-per-IP limit).
- **Renewal scheduler `stop()` ignored** — captured and called from the SIGTERM/SIGINT handlers in `server.ts` and `cli relay start`.
- **Vercel `findRecords` pagination** — already handled in `d3bb9a1` (`findRecords`→`listAllRecords`, `zoneFor`→`listAllDomains` follow `pagination.next`).
- **deSEC removed entirely** — it was only a validation vehicle; free dynDNS isn't something we ship. This also removes the wizard/schema provider mismatch (both are now `cloudflare | vercel`). Shipped DNS providers: **Cloudflare + Vercel** (both verified against LE staging) + import-cert.

**Out of scope (follow-up):**
- `upsertAddressRecord` is implemented + tested on all adapters but not yet called — it's scaffolding for **wizard A-record automation** (currently the wizard gives manual A-record guidance). Will be wired in that follow-up.
- `relay start` proxy/CORS wiring divergence (pre-existing) — tracked separately.

relay 326 + cli 208 green.

PR #327, opened 2026-06-22
feat(setup): report per-step state (found / created / repaired)

### LINKED ISSUE #326: Setup: report per-step state (found / created / repaired), not just ok
## Context

This came out of a dev.to discussion on setup idempotency (design suggestion from Alex Shev).

`tapflow setup` is already idempotent: every step checks current state and no-ops if it's already satisfied, and the shell-rc edit is wrapped in a marker block so re-runs never duplicate PATH entries. The gap is **state reporting**, not idempotency.

## Problem

A step result only distinguishes `ok` vs not-ok (`SetupStepResult = DoctorCheck`, `ok: boolean` — `packages/cli/src/lib/setup.ts:11`). So two meaningfully different outcomes collapse into the same green check:

- something that was **already present** (no change this run)
- something this run **installed or repaired**

This also hides a partial-state case. The Android SDK readiness check `sdkSelfContained()` only verifies that `sdkmanager` and `adb` exist (`packages/cli/src/lib/setup.ts:89-91`), but the install pulls `cmdline-tools` + `platform-tools` + `emulator` + the system image in a single `sdkmanager` call (`:341-345`). If the system-image download dies halfway, the next run still reports the SDK as "ready" and then fails later during AVD creation. The user can't see the partial state before that failure.

## Proposal

Add an optional `state` to the setup step result (keep `ok` so `doctor` is unaffected):

```ts
state?: 'found' | 'created' | 'repaired'
```

- `found` — already present, no change
- `created` — newly installed this run
- `repaired` — reconfigured/fixed this run

Then `printResults()` can show which happened (e.g. `✓ found`, `✓ created`, `✓ repaired`), so the summary separates what was already there from what changed.

## Scope / files

- `packages/cli/src/lib/setup.ts` — set `state` in each `checkAndFix*` result
- `packages/cli/src/commands/setup.ts` — `printResults()` renders the new states
- Follow-up (can be a separate issue): strengthen `sdkSelfContained()` to verify the system image, so a partial SDK is caught as `repaired`/incomplete instead of silently "found"

## Out of scope

Rollback / transactional setup — the convergent idempotent-re-run model stays.

mudassiralladatkhan:
## What

Adds an optional `state` to the setup step result so the summary distinguishes what was **already present** from what this run **installed or repaired** — the gap described in the issue.

```ts
export type SetupStepState = 'found' | 'created' | 'repaired'
export type SetupStepResult = DoctorCheck & { state?: SetupStepState }
```

## Why this shape

The issue notes "keep `ok` so `doctor` is unaffected." Rather than add `state` to `DoctorCheck` (shared with `doctor`), I layered it onto `SetupStepResult` at the setup layer. `DoctorCheck` is untouched, so `doctor` is provably unaffected, and `ok` keeps its existing meaning.

## State mapping (per `checkAndFix*`)

- **found** — idempotent early-return, already satisfied (Homebrew/JDK/SDK/AVD/Xcode/Simulator all present)
- **created** — newly installed this run (Homebrew, JDK, Android SDK bootstrap, AVDs, simulator runtime, Xcode)
- **repaired** — reconfigured this run; currently the Xcode activation path when `sudo xcode-select -s` / license / first-launch runs

`printResults()` renders it next to the check, e.g. `✓ Android SDK ready (found)` / `✓ AVDs created: … (created)`.

## Scope

Matches the issue's listed files exactly:
- `packages/cli/src/lib/setup.ts` — type + `state` on each `ok` result
- `packages/cli/src/commands/setup.ts` — `printResults()` renders the state

The follow-up suggestion (strengthening `sdkSelfContained()` to detect a partial SDK as `repaired`) is intentionally left out of scope per the issue — happy to do it as a separate PR.

## Tests

Added to the existing suites:
- `__tests__/lib/setup.test.ts` — `found` on a fully-configured machine, `created` when SDK/AVD/sim-runtime are installed this run, `repaired` for the Xcode activation path (android + ios)
- `__tests__/commands/setup.test.ts` — `printResults()` renders `(found)`/`(created)`, plus a regression test that a state-less `ok` step prints unchanged

## Verification

- `pnpm --filter tapflow typecheck` → clean
- `eslint src` → clean
- `vitest run` on `setup.test.ts` + command `setup.test.ts` → **39/39 pass**
- The repo's own lefthook **pre-commit** ran full workspace `typecheck` + `lint` across all 9 packages → both ✔️

No `any`, no breaking changes (`state` is optional and additive).

Closes #326

## Summary by CodeRabbit

* **New Features**
  * The setup command now shows per-step status tags on successful lines to indicate whether each prerequisite was already found, created during the run, or repaired.

* **Tests**
  * Expanded coverage for Android and iOS setup to validate correct state tagging across all major steps.
  * Added regression tests to confirm output formatting stays the same when a step has no state information.

### DIFF
--- a/packages/cli/src/__tests__/commands/setup.test.ts
+++ b/packages/cli/src/__tests__/commands/setup.test.ts
@@ -130,6 +130,28 @@ describe('cmdSetup', () => {
     expect(logLines.join('\n')).toMatch(/new terminal/i)
   })
 
+  // issue #326: ok 단계의 state를 체크 옆에 함께 출력한다.
+  it('state가 있으면 체크 옆에 (found/created/repaired) 표기', async () => {
+    mockRunSetupAndroid.mockResolvedValue([
+      { label: 'Homebrew installed', ok: true, state: 'found' },
+      { label: 'Android SDK installed', ok: true, state: 'created' },
+    ])
+
+    await cmdSetup('android')
+    const out = logLines.join('\n')
+    expect(out).toMatch(/Homebrew installed.*\(found\)/)
+    expect(out).toMatch(/Android SDK installed.*\(created\)/)
+  })
+
+  it('state가 없는 ok 단계는 기존처럼 표기 없이 출력 (회귀)', async () => {
+    mockRunSetupIos.mockResolvedValue([{ label: 'Xcode installed', ok: true }])
+
+    await cmdSetup('ios')
+    const out = logLines.join('\n')
+    expect(out).toContain('Xcode installed')
+    expect(out).not.toMatch(/Xcode installed.*\((found|created|repaired)\)/)
+  })
+
   it('인자 없음 + 감지 0개 → 안내, exit 없음', async () => {
     vi.spyOn(process, 'platform', 'get').mockReturnValue('linux')
     mockResolveAdb.mockReturnValue(null)
--- a/packages/cli/src/__tests__/lib/setup.test.ts
+++ b/packages/cli/src/__tests__/lib/setup.test.ts
@@ -246,6 +246,60 @@ describe('runSetupAndroid', () => {
       expect.anything(),
     )
   })
+
+  // issue #326: state로 "이미 있었음(found)"과 "이번에 설치함(created)"을 구분한다.
+  it("state: 완전 구성 머신은 모든 단계가 'found'", async () => {
+    const results = await runSetupAndroid()
+    expect(results.every((r) => r.ok)).toBe(true)
+    expect(findStep(results, 'homebrew')?.state).toBe('found')
+    expect(findStep(results, 'java')?.state).toBe('found')
+    expect(findStep(results, 'android sdk')?.state).toBe('found')
+    expect(findStep(results, 'avd')?.state).toBe('found')
+  })
+
+  it("state: 이번 실행에 SDK를 부트스트랩하면 'created'", async () => {
+    setTTY(true)
+    let installed = false
+    mockExistsSync.mockImplementation((p) => {
+      if (p === SDK_SDKMANAGER || p === SDK_ADB) return installed
+      if (p === SDK_AVDMANAGER || p === SDK_EMULATOR) return true
+      return false
+    })
+    mockSpawnSync.mockImplementation((cmd, args) => {
+      const a = Array.isArray(args) ? args : []
+      if (typeof cmd === 'string' && cmd.includes('sdkmanager') && a.includes('cmdline-tools;latest')) {
+        installed = true
+        return okSpawn as never
+      }
+      if (cmd === SDK_EMULATOR && a.includes('-list-avds')) {
+        return { ...okSpawn, stdout: 'tapflow-phone\n' } as never
+      }
+      return okSpawn as never
+    })
+
+    const results = await runSetupAndroid()
+    expect(findStep(results, 'android sdk')?.state).toBe('created')
+  })
+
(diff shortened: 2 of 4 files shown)


mudassiralladatkhan (commented):
<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/jo-duchan/tapflow/pull/327)</sub>

jo-duchan (commented):
Layering `state` onto `SetupStepResult` instead of `DoctorCheck` is the right call — `DoctorCheck` stays untouched and `doctor` is provably unaffected. The found/created/repaired test coverage is thorough too. Two things before merge:

**1. `checkAndFixAndroidSdk` fast-path reports `found` even when this run modified the shell rc** (`packages/cli/src/lib/setup.ts`, the `sdkSelfContained()` early return)

```ts
const reg = registerAndroidEnv()
return {
  label: 'Android SDK ready',
  ok: true,
  state: 'found',
  detail: reg?.added ? newShellHint(reg.file) : undefined,
}
```

When `reg.added === true`, this run just wrote a new PATH marker block into the user's `.zshrc`/`.bashrc`, so `found` ("already present, no change") is inaccurate — and it contradicts the `detail` line, which simultaneously tells the user "Added ANDROID_HOME/PATH… open a new terminal". The SDK binaries are genuinely `found`, but the env registration is a change. Suggest `state: reg?.added ? 'repaired' : 'found'`. Note this will affect the `완전 구성 머신은 모든 단계가 'found'` test, so please check what `registerAndroidEnv`/`reg.added` returns under that mock and adjust if needed.

**2. Condense the type comment to one line** (`setup.ts:9-11`)

Our guideline is one-line comments. The three-line block can collapse to a single line, e.g.:

```ts
// SetupStepResult = DoctorCheck + optional state (found/created/repaired); ok is untouched so doctor is unaffected.
```

Feel free to write new comments in English here — no need to match the surrounding Korean.

The Docstring-coverage and `child_process` warnings from the bots are pre-existing patterns, not introduced by this PR — you can ignore them.

mudassiralladatkhan:
Thanks for the thorough review @jo-duchan — both addressed in `dff8aae`:

**1. `repaired` vs `found` on the SDK fast-path.** Changed to `state: reg?.added ? 'repaired' : 'found'`. You were right it contradicted the `detail` hint — the binaries are `found` but writing the PATH marker this run is a change.

As you predicted, this affected the `완전 구성 머신은 모든 단계가 'found'` test: the default mock has an empty rc (`readFileSync → ''`), so `registerAndroidEnv()` appends and returns `added: true` → now `repaired`. That mock wasn't truly "fully configured" (env not yet registered). I aligned it with the idempotent test — rc marker already present → `added: false` → `found` holds. Added a dedicated regression test for the new `repaired` path (SDK present, env registered this run).

**2. Comment condensed to one line**, using your suggested English wording.

All 27 setup tests pass, `tsc --noEmit` and eslint clean. Left the pre-existing docstring/`child_process` bot warnings untouched as noted.

jo-duchan (approved):
Both points were resolved cleanly. I especially appreciate that you traced the test failure yourself — the original "fully configured" mock wasn't actually registering the env, so switching to a pre-registered rc and adding a dedicated regression test for the fixed path was the right fix, not a workaround.

The comment is much more concise now, too.

Nicely done for a good-first-issue, and thanks for the careful, thorough work. 🙏

If Tapflow ends up being useful to you, I'd really appreciate a ⭐ on the repo — but no worries either way. 😊

PR #333, opened 2026-06-23
fix(cli): clarify Android PATH refresh guidance

### LINKED ISSUE #307: doctor/setup: PATH warning persists after `setup android` because rc changes don't reach the live shell
## Summary

`tapflow doctor` warns `adb (not in PATH)` and tells the user to `Run: tapflow setup android`. But when adb's PATH entry is already registered in the shell rc (`.zshrc`/`.bashrc`), running `setup android` reports `SETUP COMPLETE` **without any guidance**, and a re-run of `doctor` in the same terminal shows the **same warning** again. From the user's side it looks like "I did exactly what it told me, but nothing changed."

This is not a functional bug — it's a confusing UX loop.

## Why it happens (evidence)

1. `resolveAdb()` decides `inPath` via `which adb`, i.e. the **live shell's PATH** — not the rc file. (`packages/cli/src/lib/doctor.ts:161`)
2. If `which adb` fails it falls back to the standard SDK path and returns `inPath: false`, which renders as a **warning** (`warn: true`), so `doctor` still prints "All checks passed". (`packages/cli/src/lib/doctor.ts:84`)
3. `setup android` registers `ANDROID_HOME`/PATH into the shell rc idempotently. If the marker block already exists it returns `added: false` and **skips the "open a new terminal" hint**. (`packages/cli/src/lib/setup.ts:306`, `setup.ts:449`)
4. The rc export only takes effect in a **new shell**. The terminal that ran `doctor`/`setup` predates (or hasn't re-sourced) the rc change, so its live PATH still lacks `platform-tools` → the warning persists.

Note: tapflow itself is unaffected — the agent resolves adb by absolute path via `ANDROID_HOME` (`packages/android-agent/src/adb.ts:7`), never via PATH. The PATH entry only matters when a human types `adb` directly.

## Steps to reproduce

1. Have the tapflow Android PATH block already present in `~/.zshrc` (e.g. from a prior `setup android`).
2. Open a terminal **without** `platform-tools` in PATH (e.g. a session started before the rc was edited).
3. `tapflow doctor` → shows `⚠ adb (not in PATH)` + "Run: tapflow setup android".
4. `tapflow setup android` → `SETUP COMPLETE`, no hint.
5. `tapflow doctor` (same terminal) → the same `⚠ adb (not in PATH)` warning.

## Proposed fix (pick one or combine)

- In `setup`, when the rc block already exists **but** adb is not in the live PATH (`resolveAdb().inPath === false`), still print the "open a new terminal (or `exec $shell`)" hint instead of staying silent.
- In the `doctor` warning detail, point at the real remedy for the in-PATH case: "open a new terminal or run `exec $SHELL` / `source ~/.zshrc`" rather than only "Run: tapflow setup android" (which the user has likely already run).

## Acceptance

- After `setup android` on a shell where adb isn't yet in PATH, the user gets a clear, actionable next step (new terminal / re-source) regardless of whether the rc block was just added or already existed.
- The `doctor` warning text doesn't loop the user back into a no-op `setup` run.

sanmaxdev:
## Summary

Fixes the Android setup/doctor PATH guidance loop from #307.

When `adb` is already present at the self-contained SDK path but not loaded in the current shell's `PATH`, `doctor` now points users to refresh the shell instead of re-running setup. `setup android` also repeats the new-shell hint when the rc block already exists but the live shell has not picked it up yet.

## Checklist
- [x] No `any`

## Related `.work/` docs

None.

Fixes #307

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `tapflow setup android` and `tapflow doctor` guidance for cases where the Android SDK is present, but `adb` is missing from the current shell PATH—now prompts to open a new terminal or refresh the shell before re-running `tapflow doctor`.
* **Documentation**
  * Clarified Android setup/doctor instructions to match the new terminal-refresh behavior (instead of pointing back to setup).
* **Chores**
  * Updated release metadata to use a patch bump for `tapflow`.

### DIFF
--- a/.changeset/quiet-mangos-serve.md
+++ b/.changeset/quiet-mangos-serve.md
@@ -1,0 +1,5 @@
+---
+"tapflow": patch
+---
+
+Clarify the Android setup/doctor PATH refresh guidance when the SDK rc block already exists but the current shell has not loaded it.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- cli: `tapflow setup android` now reminds users to open a new shell when the Android SDK rc block already exists but `adb` is still missing from the live `PATH`; `tapflow doctor` now points to the shell-refresh step instead of looping back to setup.
+
 ## [0.10.0] - 2026-06-23
 
 ### Added
--- a/packages/cli/src/__tests__/commands/setup.test.ts
+++ b/packages/cli/src/__tests__/commands/setup.test.ts
@@ -123,7 +123,7 @@ describe('cmdSetup', () => {
 
   it('env를 방금 등록(detail에 new terminal)하면 새 터미널 안내 출력', async () => {
     mockRunSetupAndroid.mockResolvedValue([
-      { label: 'Android SDK installed', ok: true, detail: 'SDK at /x. Added ANDROID_HOME/PATH to ~/.zshrc — open a new terminal (or run: exec zsh) to use them.' },
+      { label: 'Android SDK installed', ok: true, detail: 'SDK at /x. ANDROID_HOME/PATH is configured in ~/.zshrc — open a new terminal (or run: exec zsh) to use them.' },
     ])
 
     await cmdSetup('android')
--- a/packages/cli/src/__tests__/lib/doctor.test.ts
+++ b/packages/cli/src/__tests__/lib/doctor.test.ts
@@ -206,7 +206,8 @@ describe('runDoctorChecks', () => {
     expect(result.android).not.toBeNull()
     const adbCheck = result.android?.find((c) => c.label.includes('not in PATH'))
     expect(adbCheck?.warn).toBe(true)
-    expect(adbCheck?.detail).toContain('setup android')
+    expect(adbCheck?.detail).toContain('new terminal')
+    expect(adbCheck?.detail).toContain('tapflow doctor')
     expect(adbCheck?.detail).toContain(sdkAdb)
   })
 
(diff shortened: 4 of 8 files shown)


jo-duchan (commented):
Thanks for tackling #307 — this is a clean fix that addresses both problems the original report called out: `doctor` no longer sends users through a no-op `setup` run, and `setup` now surfaces the new-shell hint even when the rc block already exists.

The tests, changeset, and CHANGELOG updates are all in place. Nice work.

One wording issue to address before merge:

**`newShellHint()` says "Added ..." even when nothing was added in the current run**

In the `needsLiveShellRefresh` path (`reg.added === false`), the rc block already existed. We didn't add anything during this run — we only detected that the current shell session hasn't loaded it yet. However, `newShellHint()` still prints:

> Added ANDROID_HOME/PATH to ~/.zshrc — open a new terminal ...

The action is correct, but "Added" isn't accurate in that case. I think we can make the message work for both scenarios without introducing a second helper:

```diff
- return `Added ANDROID_HOME/PATH to ${file} — open a new terminal (or run: exec ${shell}) to use them.`
+ return `ANDROID_HOME/PATH is set in ${file} — open a new terminal (or run: exec ${shell}) to load it.`
```

(Please keep the "new terminal" substring — `commands/setup.ts` relies on it to trigger the post-banner hint.)

There's a similar wording issue in the setup banner (`packages/cli/src/commands/setup.ts:78`):

> Open a new terminal (or run: exec $SHELL), then `tapflow doctor` to verify — ANDROID_HOME/PATH was just added to your shell config.

When the rc block already exists, "was just added" isn't true either. Since this PR now makes that path reachable, it'd be good to make that message neutral as well, for example:

> Open a new terminal (or run: exec $SHELL), then `tapflow doctor` to verify — ANDROID_HOME/PATH is configured in your shell, but the current session hasn't loaded it yet.

Other than that, the logic and the regression test look good to me. Thanks for the thorough fix.

PR #335, opened 2026-06-24
fix(cli): check relay port in doctor

### LINKED ISSUE #141: feat(cli): add port-availability check to tapflow doctor
## Summary

`tapflow doctor` checks Xcode, adb, and Node versions, but it doesn't check whether port 4000 (the default relay port) is already occupied. When port 4000 is taken, `tapflow start` crashes with a bare `EADDRINUSE` error that is hard to diagnose.

A new check should catch this proactively:

```
  ✓  Node v22.0.0
  ✓  Xcode 16.2
  ✗  Port 4000
       → Port 4000 is already in use. Run: lsof -ti:4000 | xargs kill
```

## Where to look

- **Add the check here:** `packages/cli/src/lib/doctor.ts`
  - New async helper `checkPort(port: number): Promise<DoctorCheck>`
  - Wire it into `runDoctorChecks().common`
- **Tests:** `packages/cli/src/__tests__/lib/doctor.test.ts`

## Implementation hint

Pure Node.js, no extra dependencies:

```ts
import { createServer } from 'node:net'

function isPortAvailable(port: number): Promise<boolean> {
  return new Promise((resolve) => {
    const server = createServer()
    server.once('error', () => resolve(false))
    server.once('listening', () => server.close(() => resolve(true)))
    server.listen(port, '127.0.0.1')
  })
}
```

## Steps

1. Write failing tests:
   - port available → `{ label: 'Port 4000', ok: true }`
   - port occupied → `{ ok: false, detail: 'Port 4000 is already in use. Run: lsof -ti:4000 | xargs kill' }`
2. Implement `checkPort()` and add it to `runDoctorChecks().common`
3. Mock `isPortAvailable` in tests (real port binding is not allowed in unit tests per [`CONTRIBUTING.md`](../../CONTRIBUTING.md))

## Checklist

- [ ] Failing tests written before the implementation
- [ ] Port probe is mocked in tests (no real ports opened)
- [ ] Detail message includes the `lsof` fix command
- [ ] `pnpm --filter tapflow test` passes

---

> **No device needed** — port probing is pure Node.js and is mocked in tests.
> **Estimated time:** 45–60 min

**Good first PR?** The existing `checkNodeVersion()` in [`packages/cli/src/lib/doctor.ts`](../../packages/cli/src/lib/doctor.ts) is almost identical in shape — it's a great template to follow.

sanmaxdev:
## Summary

Add a common `tapflow doctor` check for the default relay port so users get a clear recovery command before `tapflow start` runs into `EADDRINUSE`.

## Checklist

- [x] No `any`

## Related `.work/` docs

N/A

Closes #141

## Summary by CodeRabbit

* **New Features**
  * Added a `tapflow doctor` check to verify whether the default relay port (`4000`) is available.
  * When `4000` is in use, the doctor output now includes a recovery command hint before `tapflow start` fails.

* **Bug Fixes**
  * Improved port-conflict diagnostics to avoid confusing `address already in use` issues.

* **Tests**
  * Expanded CLI tests to validate `Port 4000` reporting for both available and occupied cases using mocked networking behavior.

### DIFF
--- a/.changeset/thick-falcons-smile.md
+++ b/.changeset/thick-falcons-smile.md
@@ -1,0 +1,5 @@
+---
+"tapflow": patch
+---
+
+Add a `tapflow doctor` check that warns when the default relay port 4000 is already occupied.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- cli: `tapflow doctor` now checks whether the default relay port 4000 is already in use and prints the `lsof -ti:4000 | xargs kill` recovery command before `tapflow start` hits `EADDRINUSE`.
 - cli: `tapflow setup android` now reminds users to open a new shell when the Android SDK rc block already exists but `adb` is still missing from the live `PATH`; `tapflow doctor` now points to the shell-refresh step instead of looping back to setup.
 
 ## [0.10.0] - 2026-06-23
--- a/packages/cli/src/__tests__/lib/doctor.test.ts
+++ b/packages/cli/src/__tests__/lib/doctor.test.ts
@@ -156,7 +156,7 @@ describe('runDoctorChecks', () => {
 
     const result = await runDoctorChecks()
     const portCheck = result.common.find((c) => c.label === 'Port 4000')
-    expect(portCheck).toEqual({ label: 'Port 4000', ok: true })
+    expect(portCheck).toStrictEqual({ label: 'Port 4000', ok: true, detail: undefined })
   })
 
   it('Port 4000이 점유되어 있으면 해결 명령을 포함해 실패로 표시', async () => {
--- a/packages/cli/src/__tests__/lib/doctor.test.ts
+++ b/packages/cli/src/__tests__/lib/doctor.test.ts
@@ -157,6 +157,7 @@ describe('runDoctorChecks', () => {
     const result = await runDoctorChecks()
     const portCheck = result.common.find((c) => c.label === 'Port 4000')
     expect(portCheck).toStrictEqual({ label: 'Port 4000', ok: true, detail: undefined })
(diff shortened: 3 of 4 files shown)


sanmaxdev:
Tightened the port-check test assertion after the review note.

Passed locally: `pnpm --filter tapflow test -- src/__tests__/lib/doctor.test.ts src/__tests__/commands/doctor.test.ts`, `pnpm --filter tapflow typecheck`, `pnpm --filter tapflow lint`.

jo-duchan (changes requested):
### `checkPort` binds IPv4 loopback only — the relay binds dual-stack

The probe binds to `127.0.0.1`, but the relay binds with `host: '::'`:

```ts
// doctor.ts (this PR)
server.listen(port, '127.0.0.1')

// RelayServer.ts:281
this.httpServer.listen({ port, host: '::', ipv6Only: false }, resolve)
```

So if something already holds `[::1]:4000` (IPv6 loopback, or an IPv6-only listener), the probe sees the IPv4 side as free and reports `✓ Port 4000`. Then `tapflow start` still dies with `EADDRINUSE` on the IPv6 bind — the exact case this check is meant to catch.

Binding the probe the same way the relay does reproduces the same conflict:

```ts
server.listen({ port, host: '::', ipv6Only: false })
```

For the record: the `127.0.0.1` came straight from the issue body, so this isn't on you. It's just that matching how the relay actually binds is the accurate call here.

One more, not blocking: the current mock doesn't assert the `listen` host, so the tests pass either way before and after this change and won't catch a regression. A small assertion on the bind args would cover it if you want.

sanmaxdev:
Updated this to probe the same dual-stack bind as the relay (`{ port, host: '::', ipv6Only: false }`) and added a test assertion for the listen args.

Re-ran:
- `pnpm --filter tapflow test -- src/__tests__/lib/doctor.test.ts src/__tests__/commands/doctor.test.ts`
- `pnpm --filter tapflow typecheck`
- `pnpm --filter tapflow lint`

Pre-commit also ran repo lint/typecheck cleanly.

jo-duchan (approved):
Both binds line up now, and the test asserts the `host: '::'` args so it won't silently drift back. Nice quick turnaround — thanks for picking up the issue!

Issue #334, opened 2026-06-24
feat: audio input (browser → device, mic injection) for simulator/emulator


jo-duchan:
## Goal

Inject audio **input** (browser → device, mic injection) into sim/emulator sessions over the existing `browser ↔ relay ↔ agent` path, so voice/mic features can be exercised during manual QA. **Hard constraint: must not degrade the current video/render (streaming) pipeline.**

Split out from #259, which now scopes the **output** direction only (device → browser, `phase-4`). Input is the lower-priority, harder direction.

## Why separate / why later

- **Smaller use case** — only matters for testing voice/mic-driven features, not general QA.
- **Harder** — needs a virtual input device on the device side plus a browser capture surface (`getUserMedia`), which adds a **privacy/permission surface** the output path doesn't have.
- **Prerequisite: output ships first** — the transport channel, the optional audio agent interface (ISP), and the per-platform capture plumbing land with #259. Input reuses that foundation.

## Open questions (investigate at planning time)

- **Android**: how to feed PCM into the emulator mic on the default **gRPC emulator host-encode path** (not scrcpy, which is MCP/CI only).
- **iOS**: simulator mic injection path — `simctl` has no audio API; likely needs a virtual input device.
- **Browser side**: `getUserMedia` consent UX and how mic capture rides the separate audio WS channel without touching video backpressure.
- **A/V sync**: loose sync acceptable (same as output) — do not over-engineer.

## Design notes

- Model as part of the **separate optional audio agent interface (ISP)** introduced with #259, not the core `DeviceAgent`.
- Roadmap: `phase-5-plus` (AX/later), revisit once output (#259) is in.

Related: #259 (output-first, the foundation this builds on).

PR #336, opened 2026-06-24
feat(android): audio output (device → browser) + per-emulator gRPC port fix

### LINKED ISSUE #259: feat: audio output (device → browser) for simulator/emulator
## Goal

Provide audio **output** (device → browser) for sim/emulator sessions over the existing `browser ↔ relay ↔ agent` path, so the team can **hear the app** during manual QA. **Hard constraint: must not degrade the current video/render (streaming) pipeline.**

> **Scope: output only.** The reverse direction — audio **input** (browser → device, mic injection) — is split out to #334 (`phase-5-plus`, later/harder). This issue ships the foundation (transport channel + optional audio agent interface) that #334 builds on.

## Feasibility — SPIKE DONE ✅

A per-platform feasibility spike confirmed **capture = YES on both platforms** (runtime-verified):

- **Android (emulator gRPC)** — difficulty **M**. `streamAudio` RPC exists + works (420 packets/5s, 44100/Stereo/S16). Needs: conditional removal of `-no-audio` at launch + host-side Opus encoding.
- **iOS (simulator)** — difficulty **M~L**. Simulator PCM captured from a virtual audio device (BlackHole) end-to-end. No dedicated PCM-tap API exists → needs a virtual audio device (system-extension approval = setup friction) + Swift-only private API `SimDeviceAudioClient.routeGuest` for per-simulator isolation.

**No-degradation risk = low** — audio rides a separate subsystem from video on both platforms; bandwidth <1% (Opus vs H.264). Only real risk is CPU saturation under the tier1 low-latency bar — measured as a regression gate at implementation time.

## Design notes

- Model audio as a **separate optional agent interface (ISP)**, not in the core `DeviceAgent` — capture is highly platform-asymmetric (same lesson as the H.264/JPEG split).
- **Transport**: a **separate WS channel** with a small jitter buffer — audio's drop policy is the opposite of video's (video may drop to keyframe; one dropped audio frame glitches), so it must not mix into the video backpressure path.
- **A/V sync**: loose sync (play as it arrives) is acceptable for manual QA — do not over-engineer lip-sync.

## Suggested order

**Android first** (zero user setup — just drop `-no-audio`, server-side handling), **iOS follow-up** (virtual-device system-extension approval needs an opt-in setup UX first).

jo-duchan:
## What

Adds **audio output (device → browser) for the Android emulator** (issue #259, output-first), and fixes an **Android stream-bleed bug** found while testing it.

### 1. Audio output — `feat(android)`

Stream Android emulator audio to the browser over the existing relay path. **Opt-in via `TAPFLOW_ANDROID_AUDIO=1`; default off keeps the video-only path byte-for-byte unchanged.**

- **agent-core**: optional `AudioStreamCapability` (ISP — `DeviceAgent` untouched); TFFE `CODEC_AUDIO` on an independent envelope bit (no JPEG/H.264 wire-format break); `sendAudioYieldingToVideo` — audio yields to video on the shared socket so it can never trip video backpressure.
- **android-agent**: `EmulatorGrpcClient.streamAudio` (S16LE/44100/Stereo, `MODE_REAL_TIME`); conditional `-no-audio` removal at launch; `pumpAudio` on the gRPC path, torn down with the device state.
- **relay**: route `CODEC_AUDIO` frames through the yielding sender, not the keyframe-aware video dropper.
- **dashboard**: Web Audio playback (jitter buffer + resync). Always-on — muting is delegated to the emulator's own volume keys; a sound on/off indicator sits in the device info card. Also shows the gray dot + "Idle" at 0 fps instead of blanking the state.

**No-degradation** is enforced on both hops (agent→relay and relay→browser) and pinned by tests.

### 2. Per-emulator gRPC port — `fix(android)`

Every emulator was launched with and connected to a fixed gRPC port (`8554`), so a second emulator on the same Mac collided on `8554` and **every session ended up streaming the first emulator** — two device cards showing the same screen. iOS is unaffected (per-simulator IOSurface capture, no shared port).

- Launch each emulator on a unique **free** gRPC port; reserve in-flight to avoid concurrent boots racing onto the same port.
- Connect to **this** emulator's port: the launched port, else the port it advertises in its discovery `.ini` (covers externally-booted emulators), else the legacy `8554`.
- New `emulator/discovery.ts` (parse `port.serial` → `grpc.port`) with unit tests.

## Verification

- **Audio**: heard end-to-end in the browser from a remote machine; verified the production `streamAudio` path captures non-silent PCM against a live emulator.
- **Port fix**: confirmed two concurrent emulators now show independent screens; the port picker correctly skips a held `8554` to `8556` and discovery resolves `emulator-5554 → 8554`.
- Tests + typecheck green across agent-core / android-agent / relay / dashboard (lint + typecheck also run by the pre-commit hook).

## Scope notes

- Output only. Audio **input** (browser → device) is tracked separately in #334.
- iOS audio output is a planned follow-up; this PR is Android-first.

Closes #259

## Summary by CodeRabbit

* **New Features**
  * Added optional Android emulator audio streaming to the browser (opt-in via `TAPFLOW_ANDROID_AUDIO=1`), including a sound on/off indicator in the device info card.
  * Browser playback now supports receiving and playing PCM audio frames alongside video.
* **Bug Fixes**
  * Fixed multi-emulator behavior by ensuring each emulator uses its own gRPC port, preventing cross-session screen mix-ups.
  * Improved audio/video load sharing so audio uses a dedicated backpressure-aware path and won’t disrupt video streaming.

### DIFF
--- a/.changeset/android-audio-output.md
+++ b/.changeset/android-audio-output.md
@@ -1,0 +1,5 @@
+---
+"@tapflowio/android-agent": minor
+---
+
+Add audio output (device → browser) for the Android emulator. Opt-in via `TAPFLOW_ANDROID_AUDIO=1` (default off keeps the video path unchanged). Emulator audio is captured over the gRPC `streamAudio` path and carried on a separate envelope codec that yields to video, so it never affects the existing stream; the browser plays it via Web Audio with a sound on/off indicator in the device info card.
--- a/.changeset/android-grpc-port-per-emulator.md
+++ b/.changeset/android-grpc-port-per-emulator.md
@@ -1,0 +1,5 @@
+---
+"@tapflowio/android-agent": patch
+---
+
+Fix concurrent Android emulators sharing one video stream. Each emulator now launches on, and connects to, its own gRPC port (discovered from the running emulator's `.ini`) instead of a fixed `8554`, which collided when more than one emulator ran on the same Mac and made every session show the first emulator's screen.
--- a/packages/agent-core/src/AudioStreamCapability.ts
+++ b/packages/agent-core/src/AudioStreamCapability.ts
@@ -1,0 +1,32 @@
+import type { DeviceAgent } from './DeviceAgent.js'
+
+// Optional audio-output capability (device → browser), kept OUT of the core
+// DeviceAgent interface (ISP): audio capture is platform-asymmetric and opt-in,
+// so only agents that can capture audio implement it. Consumers must feature-detect
+// with hasAudioCapability() before using it — the video path never depends on this.
+
+export type AudioSampleFormat = 's16' | 'u8'
+export type AudioChannels = 'mono' | 'stereo'
+
+export interface AudioFormat {
+  sampleRate: number          // e.g. 44100
+  channels: AudioChannels
+  sampleFormat: AudioSampleFormat
+}
+
+export interface AudioFrame {
+  payload: Buffer             // raw PCM samples in the stream's AudioFormat (no codec)
+  timestamp: number           // capture timestamp, epoch microseconds — for loose A/V sync
+}
+
+export interface AudioStreamCapability {
+  // The constant PCM format this agent produces (known up front, no await needed).
+  audioFormat(): AudioFormat
+  // Raw PCM frames. Mirrors DeviceAgent.stream(): a ReadableStream the caller drains.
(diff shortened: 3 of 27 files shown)


jo-duchan:
@coderabbitai review

PR #343, opened 2026-06-28
fix(cli): repair partial Android SDK setup

### LINKED ISSUE #330: Setup: sdkSelfContained() reports a partial SDK as "ready" — verify the system image
## Context

Follow-up split out of #326, which intentionally left this out of scope. With per-step `state` reporting (found/created/repaired) now in place (#327), the next gap is the readiness check itself.

## Problem

`sdkSelfContained()` only checks that `sdkmanager` and `adb` exist (`packages/cli/src/lib/setup.ts:89-91`):

```ts
function sdkSelfContained(): boolean {
  return existsSync(join(SDK_CMDLINE_BIN, 'sdkmanager')) && existsSync(join(ANDROID_SDK_DIR, 'platform-tools', 'adb'))
}
```

But the install pulls `cmdline-tools` + `platform-tools` + `emulator` + the system image in a single `sdkmanager` call (`:341-345`). If the system-image (or emulator) download dies halfway, the next `tapflow setup` run still reports the SDK as `Android SDK ready (found)` and then fails later during AVD creation. The user can't see the partial state before that failure.

## Proposal

Strengthen `sdkSelfContained()` (or add a companion check) to also verify the emulator binary and the expected system image are present. When the SDK is partially installed, the run should re-run the install and report `created`/`repaired` instead of silently passing as `found`.

## Scope / files

- `packages/cli/src/lib/setup.ts` — readiness check + the `checkAndFixAndroidSdk` fast-path
- `packages/cli/src/__tests__/lib/setup.test.ts` — partial-SDK case

## Out of scope

Rollback / transactional setup — the convergent idempotent-re-run model stays.

sanmaxdev:
## Summary

- Tighten Android SDK readiness so `tapflow setup android` only reports the SDK as ready when cmdline-tools, platform-tools, the emulator binary, and the expected system image are all present.
- Let partial SDK installs fall back through the existing sdkmanager repair path instead of being reported as `found`.
- Add a regression test for the partial-SDK case and update the changelog.

Closes #330

## Checklist

- [x] No `any`

## Related `.work/` docs

N/A — small scoped CLI setup fix.

## Verification

- `pnpm --filter tapflow exec vitest run src/__tests__/lib/setup.test.ts`
- `pnpm --filter tapflow exec vitest run src/__tests__/lib/setup.test.ts src/__tests__/commands/setup.test.ts src/__tests__/lib/doctor.test.ts`
- `pnpm --filter tapflow lint`
- `pnpm --filter tapflow typecheck`
- pre-commit hook: root `pnpm -r --if-present lint` and `pnpm -r --if-present typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android SDK setup detection: missing emulator binary or the configured system image is now recognized as a partial SDK and automatically repaired.
  * Android setup now labels and classifies the “Android SDK” step more accurately when components are incomplete, preventing “ready” status from appearing too early.
* **Documentation**
  * Added a release note describing that `tapflow setup android` will repair a partial Android SDK instead of treating it as fully configured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- cli: `tapflow setup android` now treats a missing emulator binary or Android system image as a partial SDK and repairs it instead of reporting the SDK as ready.
 - cli: `tapflow doctor` now checks whether the default relay port 4000 is already in use and prints the `lsof -ti:4000 | xargs kill` recovery command before `tapflow start` hits `EADDRINUSE`.
 - cli: `tapflow setup android` now reminds users to open a new shell when the Android SDK rc block already exists but `adb` is still missing from the live `PATH`; `tapflow doctor` now points to the shell-refresh step instead of looping back to setup.
 
--- a/packages/cli/src/__tests__/lib/setup.test.ts
+++ b/packages/cli/src/__tests__/lib/setup.test.ts
@@ -186,9 +186,19 @@ describe('runSetupAndroid', () => {
       if (p === SDK_EMULATOR || p === SDK_SYSTEM_IMAGE) return installed
       return false
     })
+    const expectedSystemImage =
+      process.arch === 'arm64'
+        ? 'system-images;android-35;google_apis;arm64-v8a'
+        : 'system-images;android-35;google_apis;x86_64'
+
     mockSpawnSync.mockImplementation((cmd, args) => {
       const a = Array.isArray(args) ? args : []
-      if (typeof cmd === 'string' && cmd.includes('sdkmanager') && a.includes('cmdline-tools;latest')) {
+      if (
+        typeof cmd === 'string' &&
+        cmd.includes('sdkmanager') &&
+        a.includes('cmdline-tools;latest') &&
+        a.includes(expectedSystemImage)
+      ) {
         installed = true
         return okSpawn as never
       }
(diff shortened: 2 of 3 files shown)


jo-duchan (approved):
Verified the fix against the setup code and it holds up:

- `androidSystemImageDir()` maps `system-images;android-35;google_apis;<abi>` to the right on-disk path (the `;`→`/` split matches the sdkmanager layout), and the emulator path lines up with the one `doctor.ts` already uses.
- A partial SDK now fails `sdkSelfContained()` and falls through to the existing install call, which re-pulls `emulator` + the system image idempotently and reports `created`. The post-install re-check at `setup.ts:379` uses the strengthened check too, so a second failure is reported honestly instead of passing as ready.
- Ran a quick mutation check: reverting `sdkSelfContained()` to its old two-path form makes the new test fail, so the test is actually pinning the new behavior rather than passing vacuously.

Approving. One optional thought, not blocking:

`existsSync(androidSystemImageDir())` only checks that the directory exists. If sdkmanager dies mid-download and leaves an empty or partial image dir behind, this would still read as present. If sdkmanager stages into a temp dir and moves atomically on completion, this never happens and the directory check is fine — do you know which it is? If partial dirs can linger, checking for a marker file like `source.properties` inside would close that gap. Either way it's no worse than the existing adb/sdkmanager file checks, so this is a follow-up at most.

jo-duchan:
Confirmed CodeRabbit's catch here. I ran a quick mutation check: dropping `androidSystemImage()` from the install call in `setup.ts` and this test still passes, so the system-image part isn't actually pinned down. Since repair (not just detect) is the whole point of this PR, it's worth closing that gap.

Two small tweaks do it: hold `installed` from flipping until the system image is requested too, and add `androidSystemImage()` to the `arrayContaining` list.

No real bug today — the install command and the `setup.ts:379` re-check both handle it — this is purely about the test guarding against a future regression. Nice find by the bot.

sanmaxdev:
Thanks, agreed. I tightened the regression test so it only flips the mock SDK to installed when the system image package is requested too, and the sdkmanager assertion now includes that expected system image.

Verified:
- `pnpm --filter tapflow exec vitest run src/__tests__/lib/setup.test.ts`
- `pnpm --filter tapflow lint`
- `pnpm --filter tapflow typecheck`
- pre-commit root lint/typecheck

PR #355, opened 2026-07-02
docs: add systemd relay deployment guide

MFA-G:
## Summary by CodeRabbit

* **Documentation**
  * Added a new “systemd (Linux relay server)” section to the self-hosting guides.
  * Includes steps to install the package globally, create a dedicated user and data directory, and configure a protected environment file with required relay settings.
  * Provides a systemd service unit setup plus commands to reload systemd, enable/start the service, check status, view logs, and restart after updates.
  * Updated the Korean self-hosting guide with the same systemd instructions.

### DIFF
--- a/docs/guide/self-hosting.md
+++ b/docs/guide/self-hosting.md
@@ -362,3 +362,75 @@ pm2 restart relay
 ::: tip Next step
 Once the relay is running, open `http://localhost:4000` in a browser — the dashboard redirects to the setup page automatically. For headless servers, use `tapflow admin init` instead. For team invitations and your first build upload, see [First-time Setup](/dashboard/setup).
 :::
+
+## systemd (Linux relay server)
+
+Use systemd when the relay runs on a Linux host and you want it to start at boot, restart after crashes, and write logs to journald. Install tapflow globally first:
+
+```sh
+npm install -g tapflow
+```
+
+Create a dedicated user and data directory:
+
+```sh
+sudo useradd --system --create-home --home-dir /var/lib/tapflow --shell /usr/sbin/nologin tapflow
+sudo mkdir -p /etc/tapflow /var/lib/tapflow/.tapflow-data
+sudo chown -R tapflow:tapflow /var/lib/tapflow
+```
+
+Put relay secrets in `/etc/tapflow/relay.env`:
+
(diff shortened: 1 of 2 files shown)


jo-duchan:
Thanks for the thorough guide — I verified the CLI commands, \`TAPFLOW_DATA_DIR\`, and \`JWT_SECRET\` usage against the existing docs and they all line up. Two small, non-blocking suggestions before merge:

1. **Environment file location.** The rest of self-hosting.md keeps relay secrets in \`.tapflow-data/.env\` (the file the relay reads on start), whereas this section uses systemd's \`EnvironmentFile=/etc/tapflow/relay.env\`. The systemd-native approach is perfectly valid and arguably cleaner, so this is fine to keep — just flagging that we now document two conventions. A one-line note pointing readers to the existing [JWT_SECRET](#jwt-secret) section (so they know both paths exist) would smooth this over.

2. **Working directory has no config.** The relay reads \`tapflow.config.json\` from its working directory, but \`WorkingDirectory=/var/lib/tapflow\` has no config file, so it will fall back to defaults (port 4000, etc.). That is fine for a basic setup, but a short sentence like "place \`tapflow.config.json\` in \`/var/lib/tapflow\` if you need to customize the port or other settings" would help anyone who wants non-default config.

Neither is a blocker — the guide is accurate and the EN/KO pages are in sync. Nice work!

jo-duchan:
Thanks @MFA-G — both pieces of feedback are addressed cleanly. ✅

- The `/etc/tapflow/relay.env` section now points readers to the existing [JWT_SECRET](#jwt-secret) / `.tapflow-data/.env` convention, so the two approaches no longer read as conflicting.
- The `tapflow.config.json` / `WorkingDirectory` note is exactly what was needed for anyone customizing the port or other settings.

I double-checked the `#jwt-secret` anchor (the `### JWT_SECRET` heading exists in both the EN and KO pages), and the EN/KO docs are in sync. 

I'm going to merge this now. Really appreciate the well-structured, accurate guide and the quick follow-up on the review. 🙌

If tapflow has been useful to you, a ⭐️ on the repo would mean a lot and helps others discover the project: https://github.com/jo-duchan/tapflow

PR #366, opened 2026-07-03
docs: harden relay systemd guide

### LINKED ISSUE #365: Harden the relay systemd unit and document a first-run smoke test
## Context

The relay **systemd** deployment guide landed in #355 (closes the original scope of #353): unit file example, dedicated user + data dir, `EnvironmentFile`, `Restart=on-failure`, journald logs, and the KO mirror.

A follow-up review ([comment on #353](https://github.com/jo-duchan/tapflow/issues/353#issuecomment)) suggested two **additive** improvements on top of the merged guide. This issue tracks them separately since they are a security/operability enhancement rather than the base guide. Still **docs-only**.

## Scope

- [ ] Add a minimal **hardening baseline** to the unit example in `guide/self-hosting.md`. The unit already isolates with a dedicated `User=tapflow`; layer on the standard sandboxing directives, e.g.:
  ```ini
  NoNewPrivileges=true
  ProtectSystem=strict
  ProtectHome=true
  PrivateTmp=true
  ReadWritePaths=/var/lib/tapflow
  ```
  Verify these do not block the data dir (`/var/lib/tapflow/.tapflow-data`) or the relay's listen socket before documenting — `ProtectSystem=strict` makes the FS read-only except `ReadWritePaths`.
- [ ] Document a **first-run smoke test** to run *before* `systemctl enable --now`. There is no dedicated offline config-validate command in the CLI, so use the real flow:
  1. Run `tapflow relay start` in the foreground and confirm it binds.
  2. In another shell, check `tapflow status`.
  3. Stop it (Ctrl-C), then enable the service.
- [ ] Mirror both additions in `ko/guide/self-hosting.md` (KO).

## Note

Keep to the platform-neutral, `---`-free, and sh-codeblock conventions in `docs/AGENTS.md`. This is a good first contribution.

vku2018:
Closes #365

Summary:
- Add a minimal systemd sandboxing baseline to the relay unit example.
- Document why ReadWritePaths=/var/lib/tapflow preserves the default data directory and what to change if TAPFLOW_DATA_DIR moves.
- Add a foreground first-run smoke test before systemctl enable --now.
- Mirror the same guidance in the Korean self-hosting page.

Validation:
- git diff --check
- npm install --ignore-scripts --no-package-lock (from docs)
- npm run build (from docs)

Note:
- 
pm --prefix docs install --no-package-lock initially failed on Windows because the repo lifecycle script uses lefthook install || true; retrying inside docs with --ignore-scripts installed the docs dependencies for the build check.

### DIFF
--- a/docs/guide/self-hosting.md
+++ b/docs/guide/self-hosting.md
@@ -409,12 +409,39 @@ EnvironmentFile=/etc/tapflow/relay.env
 ExecStart=/usr/bin/env tapflow relay start
 Restart=on-failure
 RestartSec=5
+NoNewPrivileges=true
+ProtectSystem=strict
+ProtectHome=true
+PrivateTmp=true
+ReadWritePaths=/var/lib/tapflow
 
 [Install]
 WantedBy=multi-user.target
 ```
 
+The hardening directives keep the filesystem read-only except for
+`/var/lib/tapflow`, which includes the configured
+`/var/lib/tapflow/.tapflow-data` directory. If you move
+`TAPFLOW_DATA_DIR` outside `/var/lib/tapflow`, add that path to
+`ReadWritePaths` too.
+
 Place `tapflow.config.json` in `/var/lib/tapflow` if you need to customize the port or other settings, because that is the service `WorkingDirectory`.
+
+Run a foreground smoke test before enabling the service. This confirms the
+relay can bind and write to the data directory with the same dedicated user:
+
+```sh
+cd /var/lib/tapflow
+sudo -u tapflow env TAPFLOW_DATA_DIR=/var/lib/tapflow/.tapflow-data tapflow relay start
+```
+
+In another shell, confirm the relay responds:
+
+```sh
+tapflow status
+```
+
+Stop the foreground relay with Ctrl-C after the status check passes.
 
 Enable and start the service:
 
--- a/docs/ko/guide/self-hosting.md
+++ b/docs/ko/guide/self-hosting.md
@@ -409,12 +409,40 @@ EnvironmentFile=/etc/tapflow/relay.env
 ExecStart=/usr/bin/env tapflow relay start
 Restart=on-failure
 RestartSec=5
+NoNewPrivileges=true
+ProtectSystem=strict
+ProtectHome=true
+PrivateTmp=true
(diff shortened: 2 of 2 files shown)


jo-duchan:
Thanks for this, @vku2018 — the doc changes look solid and match the issue scope well. One thing left from #365 is the runtime check: since the hardening directives were suggested but not yet run on an actual Linux systemd host, it'd be good to confirm the relay still binds and writes under `ProtectSystem=strict` + `ReadWritePaths`. Are you still planning to pick this up? If it's easier, we're happy to do that verification on our side and take it from your draft — just let us know. No rush either way, and thanks again for the contribution! 🙏

jo-duchan:
Hi @vku2018 — thanks for this, and sorry it sat so long. We reached out by email as well and didn't hear back, so rather than let it go stale we've carried it forward in #461.

Your commit is cherry-picked there unchanged and first in the history, so it stays authored by you. The sandboxing baseline and the idea of a foreground smoke test before `systemctl enable --now` are both yours and both kept.

Two things came up when we checked the text against the code, in case you're curious:

- The smoke test named `/var/lib/tapflow/.tapflow-data`, but the unit's EnvironmentFile sets `TAPFLOW_DATA_DIR=/var/lib/tapflow/.tapflow/data`. `.tapflow-data` is the older layout the same page tells you to migrate off, so following the steps would have exercised a different directory than the service uses. It now loads the same env file systemd does.
- `tapflow status` resolves the relay from the calling shell's own config, so in a second terminal it looks for the default port — not one set in `/var/lib/tapflow/tapflow.config.json`, which the page suggests a few paragraphs earlier. It now passes `--relay` explicitly.

Neither is obvious from reading the page; they only show up by cross-checking against the CLI. Happy to have you review #461 if you have time, and very welcome back for the next one.

jo-duchan:
Merged as part of #461, so closing this one.

Your commit `7798bdf` is on `main` authored by you — it shows up under your account in the contributors graph and in `git log`, which is the outcome we wanted. #365 is closed.

Thanks again, @vku2018. The hardening baseline and the smoke-test idea were both right; everything after that was detail. If you want another, the `good first issue` label has a few, and I'll watch them more closely than I did this one.

PR #373, opened 2026-07-05
release: graduate mcp-server and flow-runner to the standard npm channel

jo-duchan:
## Summary

Closes the release-pipeline gap found while adding flow-runner: the experimental dist-tag scheme turned out to only work for **leaf** packages. Two structural facts, both surfaced by adversarial review and reproduced locally:

1. Once the stable CLI depends on flow-runner, changesets **hard-fails config validation** ("tapflow depends on the skipped package @tapflowio/flow-runner") — the next tag push would have died before publishing anything.
2. The existing `pnpm exec npm publish` path **never rewrote `workspace:*`** — published mcp-server tarballs were installable only because mcp-server had zero workspace deps until M2 added one. The next release would have shipped uninstallable tarballs.

Instead of patching the special-casing, both packages graduate to the standard channel.

## Changes

- **Both packages**: `0.13.0-experimental.1` → `0.13.0`, `publishConfig.tag` removed; joined the changesets **fixed group**, removed from `ignore`.
- **release.yml**: private-toggle + two per-package experimental publish steps deleted. A single `pnpm changeset publish` now covers every public package and publishes via pnpm, which rewrites `workspace:*` to exact versions — verified with `pnpm pack` on both tarballs.
- **Changeset added** for the M1/M2 feature set (missed at merge time). `changeset version` was simulated with the real CLI: the fixed group bumps cleanly to 0.14.0, no double-bump.
- **Docs**: experimental banners removed (README, cli README, `docs/guide/mcp-server.md` EN/KO, INDEX.md, both AGENTS.md). README + LICENSE added to both packages — their npm pages were blank.
- **`.claude/commands/release.md`** rewritten: the manual experimental-bump procedure is retired, and the one-time pre-tag requirements are documented.

## Before tagging the next release (one-time, documented in release.md §7)

- [ ] **Seed-publish flow-runner locally via the pnpm path**: `pnpm --filter @tapflowio/flow-runner publish --access public --no-git-checks`. npm OIDC trusted publishing cannot publish a brand-new package (a trusted publisher can only be registered on an existing one), and changesets publishes concurrently with no topological order — a partial failure would put a `tapflow` on `latest` that exact-pins a nonexistent flow-runner version.
- [ ] Register the trusted publisher for flow-runner on npmjs.com (repo `jo-duchan/tapflow`, workflow `release.yml`).
- [ ] MCP stabilization track (press_key/press_button fixes + full tool verification) merged first — graduating a package with a known no-op tool would be a bad first impression on `latest`.
[...]

PR #374, opened 2026-07-05
process: adversarial review gate before every code-change PR

jo-duchan:
## Summary

Encodes the adversarial-review step into the workflow and enforces it mechanically. The motivation is fresh: this process just caught two release-pipeline blockers on the graduation PR (#373) that the authoring session had missed — the authoring context inherits its own assumptions, so refutation must come from an independent context.

## What this adds

- **AGENTS.md — Adversarial Review section**: every code-change PR gets refuted by a fresh subagent that has NOT seen the working conversation (diff + repo access + refute-first prompt only). Protocol / public-interface / release-infra changes escalate to a second independent channel (different-lens subagent, or Codex for cross-model independence). Findings and dispositions (fixed / skipped+reason) go to `.work/reviews/<branch>.md` with the full 40-char reviewed HEAD hash. Docs-only PRs may skip the review but still write the record.
- **`.claude/hooks/adversarial-review-gate.sh`** (registered in `.claude/settings.json`): a PreToolUse gate that blocks PR creation unless that record exists and references the current HEAD — any commit after the review invalidates the record until it is refreshed. Matches the gh invocation only in command position (line start / `;` `&&` `|` / `$(` / `then` `do`) so commit messages and docs that merely mention the command don't false-block. Fail-open by design: the gate guards a cooperative-but-forgetful agent, not an adversary.

## Dogfooding

This PR went through its own gate, twice over:

- The gate **blocked this branch's first commit attempt** — the commit message mentioned the gh command and the initial substring matcher false-positived. That produced the command-position matcher.
- The change was then adversarially reviewed by an independent shell/hooks-lens subagent, which found 4 majors (docs-only contradiction, `$(...)`/`then` accidental bypasses, short-hash trap with a misleading error, quoted-mention false positives) — all fixed or documented in the second commit, with a 7-scenario regression test.
- This PR's creation itself passed only after the review record for the final HEAD was written.

Verified live on PR #373 as well: blocked without a record, blocked on stale hash, passed with a valid record.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the release workflow to require an adversarial review record before creating release PRs.
  * Added guidance that version-bump PRs can skip review only when a skip reason and the full HEAD commit hash are recorded.
  * Documented the required diff refutation step and review record format enforced before code-change PRs.
* **Chores**
  * Added an automated preflight gate that blocks PR creation unless a local review record exists for the branch and matches the current HEAD commit hash.

### DIFF
--- a/.claude/commands/release.md
+++ b/.claude/commands/release.md
@@ -66,6 +66,7 @@ model: claude-opus-4-8
 
 - `git push -u origin release/vX.Y.Z`
 - `gh pr create --base main` — 본문에 버전 표 + 릴리즈노트 + 호환성 근거.
+- adversarial review gate가 릴리즈 PR에도 적용된다 — 버전 범프 PR은 docs-only 취급으로 리뷰 생략 가능하되 `.work/reviews/<브랜치>.md`에 스킵 사유 + full HEAD 해시(git rev-parse HEAD) 기록은 필수.
 - **PR을 머지하지 않는다.** 머지는 사용자 몫.
 
 ## 11. 머지 후 — 태그 push로 발행 트리거 (놓치면 발행이 안 됨)
--- a/.claude/hooks/adversarial-review-gate.sh
+++ b/.claude/hooks/adversarial-review-gate.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# PreToolUse(Bash) gate: blocks `gh pr create` unless an adversarial review
+# record exists for the current branch AND references the current HEAD commit.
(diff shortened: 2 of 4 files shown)


jo-duchan:
Addressed the CodeRabbit nitpick: `then`/`do` were matched as bare substrings, so a word ending in "do" directly before the gh invocation text could spuriously count as command position and over-block. They now require line start or whitespace before them — verified with an 8-scenario regression run (the new case included).

PR #401, opened 2026-07-21
docs(contributing): record workshop-lab fork observations (capacity data + extension-seam signal)

jo-duchan:
## What

Adds `contributing/workshop-lab-fork-observations.md`, a third-party field-observation record drawn from [Raghu Betina's `tapflow-workshop-lab` fork](https://github.com/raghubetina/tapflow-workshop-lab), plus its two index registrations (`contributing/README.md`, root `INDEX.md`).

Two takeaways worth keeping:

1. **Independent capacity data point** — on a 32 GB M1 Max, ~4 prewarmed iOS simulators is comfortable, 6 borderline, 8 unacceptable. The first non-self-run measurement of concurrent-simulator headroom; the 4→8 cliff corroborates the memory/swap assumption behind the relay's boot-time resource gate.
2. **Extension-seam signal** — the fork's seat-pool layer is not a new *platform*, so `AgentRegistry.register()` doesn't cover it; the snapshot commit modifies core (`RelayServer` +1161/−40, `IOSAgent`, `SimctlWrapper`, `SessionManager`) rather than registering an add-on. Records *where* session-lifecycle features fall outside our one clean seam — as an open question, not a plan.

The doc also explicitly records what we do **not** take: the seat-pool feature (out of manual-first / trusted-team scope) and the report's simulator-network-stack security finding (a containment concern that only bites with untrusted actors; our trusted-team + dual LAN/id-pw gate model does not have that exposure until a public tunnel removes the LAN gate).

## Evidence

- Capacity numbers paraphrased from the fork's `FIRST_DRAFT_WORKSHOP_LAB_EXPERIMENT.md` (facts, not copied prose).
- Core-modification claim verified against the GitHub commit API for `3945d74` (83 files: 32 added / 51 modified vs upstream `10aa531`); cited line counts are the API's own diffstat.
- License checked: the fork retains tapflow's MIT and `Copyright (c) 2026 tapflow contributors`, so citation needs no permission. Attribution + source link included as a courtesy.

## Note on merge

Merge is intentionally gated on the fork author's reply to a courtesy outreach message. Opening the PR now; **@jo-duchan will merge after the author acks.**

## Review

Docs-only — adversarial review itself skipped per policy; record kept for the gate at `.work/reviews/docs__contributing-workshop-lab-fork-observations.md` (HEAD `cbeb89dfb3ee65a30c12217d3a64306d2cf58929`). All external claims traced to primary sources (commit API, report, LICENSE) before writing.

## Summary by CodeRabbit

* **Documentation**
  * Added a new reference document with third-party workshop observations on iOS simulator capacity and extension boundary signals.
  * Updated the contributing documentation indexes to reference the new record and its topics.
  * Included key measurement caveats, explicitly excluded certain reported takeaways, and added links to related internal materials.


jo-duchan:
Merged `main` in and took the finding.

**The over-claim was real.** The passage named "host-local / private-LAN services" and then answered only the LAN half before concluding there was nothing new to contain. The LAN half holds up — a logged-in teammate already reaches those services from their own machine. The host-local half does not: a simulator shares the host's network stack, so it reaches services bound to the agent Mac's loopback, which a teammate on the LAN cannot. That is reach the participant did not previously have, trusted or otherwise.

Rewritten to split the two and to say what we actually do about the second one — accept it under the trusted-team model rather than contain it. Accepted is not the same as absent, which was the whole of your point.

**Conflict resolved** in `INDEX.md`. `main` removed the `## Hierarchy` block in #415 (context-engineering slimming), so that deletion is kept and only this PR's table row survives.

On the `changeset` check that was stuck at "Expected — waiting for status to be reported": nothing was wrong with the PR. The branch predates the job, so the required check had no run to report and would have waited forever. Merging `main` in gave it one, and it passes — this diff touches `INDEX.md` and `contributing/` only, so no changeset is required or wanted.

PR #410, opened 2026-07-25
fix(ios-agent): lazily set up TouchHelper so boot-less session input isn't dropped

jo-duchan:
## Summary

Fixes **followups H-E** — iOS agent input was silently dropped on a session that never ran `handleDeviceBoot`.

`TouchHelper` was created only in `handleDeviceBoot -> sendChromeData`. When a session becomes active without a fresh `device:boot` — an **agent reconnect re-issues the session** (`deviceStates` recreated with `touchHelper=null`) while the simulator stays booted, and the client sends input — every touch/pinch/key/button was silently dropped. Symptom: **reads work, writes vanish** (UI-tree reads work because the tree runner is udid-keyed and session-independent; input writes are lost).

## Fix

Add synchronous `ensureTouchHelper(state)` and call it at the input entrypoints (`input:touch:start`, `input:pinch:start`, `input:key`, `input:button`, `input:type`). Input self-heals instead of dropping.

**Kept synchronous on purpose**: a tap is `input:touch:start` + `input:touch:end` (two messages). An async setup would let the `end` run before the helper exists, dropping `touchEnd` (a stuck finger). Sync setup means `start` creates the helper before `end` is handled.

## Scope / follow-ups

- Restores the touch channel: **tap / swipe / key / pinch / home button**.
- **Chrome-named buttons** (volume, etc.) additionally need `loadedChrome` (also boot-only) — narrower follow-up, not in this PR.
- **Truthful failure** when input genuinely can't be injected (today's false `{"tapped":true}`, and `input:type` now emitting `input:type-done` on a shut-down sim) is **H-F** — a separate additive input-ack PR, tracked in the AI Automation maturity plan.

## Verification

- Two new reproducing tests (lazy creation on boot-less attach; start/end pairing / no stuck finger) — **fail without the fix, pass with it** (verified by reverting the source).
- Full ios-agent suite **227/227**, `pnpm typecheck` green, lint clean.
- **Adversarial review** by an independent context: sound, no BLOCKER/MAJOR; two MINORs (both inherent tradeoffs of the sync design, deferred to H-F). Record: `.work/reviews/fix__ios-input-lazy-touchhelper.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved touch, pinch, typing/paste, keyboard key, and button input handling when sessions are reissued after reconnects or when touch support isn’t initialized yet.
  * Prevented lost or stuck touch interactions by lazily initializing missing touch support before dispatching input events.

* **Tests**
  * Added coverage for scenarios where sessions skip initial device setup, verifying touch support is created on first touch input and that touch start/end both dispatch correctly.

### DIFF
--- a/packages/ios-agent/src/IOSAgent.ts
+++ b/packages/ios-agent/src/IOSAgent.ts
@@ -577,6 +577,19 @@ export class IOSAgent implements DeviceAgent {
     }
   }
 
+  // Ensure the session's touch channel exists. handleDeviceBoot is the normal setup
+  // path, but a session can become active without it — an agent reconnect re-issues
+  // the session (deviceStates recreated with touchHelper=null) while the simulator
+  // stays booted, and the client may send input without a fresh device:boot. Create
+  // the helper lazily so input self-heals instead of being silently dropped. Kept
+  // synchronous so a tap's start+end stay paired: an async setup would let the end
+  // run before the helper exists and drop touchEnd (a stuck finger).
+  private ensureTouchHelper(state: DeviceState): void {
+    if (state.touchHelper) return
+    state.touchHelper = new TouchHelper(state.deviceId)
+    state.touchHelper.start()
+  }
+
   private handleRelayMessage(msg: { type: string; sessionId?: string; payload?: unknown }): void {
     switch (msg.type) {
       case 'device:boot': {
@@ -624,9 +637,10 @@ export class IOSAgent implements DeviceAgent {
       }
       case 'input:touch:start': {
         const state = this.deviceStates.get(msg.sessionId!)
-        if (!state?.touchHelper) { logger.error('touch:start — touchHelper not ready'); break }
+        if (!state) break
+        this.ensureTouchHelper(state)
         const { x, y } = msg.payload as { x: number; y: number }
-        state.touchHelper.touchStart(x, y)
+        state.touchHelper?.touchStart(x, y)
         break
       }
       case 'input:touch:move': {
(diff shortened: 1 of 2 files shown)


jo-duchan:
Thanks @coderabbitai — dispositions on the two findings (pushed in a7022f7):

**#2 (comments, English + one line) — Fixed.** Condensed the `ensureTouchHelper` rationale and the test comments to single English lines and dropped the Korean phrase. Lint clean.

**#1 (model the reconnect lifecycle after an initial boot) — Skipped, with reasoning.** I implemented the boot → reconnect → rejoin-without-boot → input integration test, but it times out in-harness (the reconnect handshake / old-socket teardown is flaky) — the "heavy lift" fragility flagged in the finding. Rationale for keeping the direct tests:

- They correctly exercise the code under test: a `DeviceState` with `touchHelper=null` on a booted sim + input → `ensureTouchHelper` lazily creates it. The *mechanism* that produces the null state doesn't change that code path.
- They are non-Potemkin — both fail when the source fix is reverted (verified).
- The reconnect→null-state mechanism is code-verified: `_scheduleReconnect` clears `deviceStates` → `initDeviceStates` recreates them with `touchHelper=null`, sims stay booted across reconnect, and `SessionManager.create` mints a fresh `randomUUID` session id.
- The cited guideline ("TouchHelper tests go through `device:boot` first") has a premise this change intentionally breaks: the lazy path exists to create `TouchHelper` *without* boot, so a boot-first test would create the helper via boot and never exercise the lazy branch. The intentional boot-skip is now documented in the test comment.

PR #411, opened 2026-07-25
feat(mcp): truthful input acks for tap/swipe/press_key/press_button (H-F)

jo-duchan:
## Summary

Fixes **followups H-F** — MCP input tools were fire-and-forget. The client sent `input:touch:*` / `input:key` / `input:button` and the tool fabricated `{tapped:true}` regardless of what the agent did. On a session whose device isn't booted, the input was silently dropped but reported as success — a false positive that also makes parallel test results untrustworthy (the trustworthiness prerequisite in the AI Automation maturity plan).

## Change (4 layers, reuses the existing request/response convention)

Adds a terminal-input ack with the same shape as `input:type-done` / `screenshot:done`:

- **agents (ios + android)** — on a gesture's terminal message (`input:touch:end`, `input:pinch:end`, `input:key`, `input:button`) send `input:done` if the input was dispatched to a booted device, else `input:error`. `input:done` means *dispatched* (agent handled it), **not a landing guarantee** — same semantics as the sibling `:done` acks (HID injection is fire-and-forget). Truthfulness uses a cached `DeviceState.booted` flag (set on `device:ready`, cleared on shutdown) and, only when the flag is false (post-reconnect / dead sim), one `simctl`/`adb` verify. Injection stays **synchronous**, so a tap's start/end pairing (H-E) is unaffected — the ack runs off that path.
- **relay** — forward `input:done` / `input:error` to the browser socket (allowlist + `MessageType` + forward switch), mirroring `input:type-done`.
- **mcp-server** — `tap` / `swipe` / `press_key` / `press_button` await the ack and surface `input:error` as a tool failure.

## Backward compatibility

**Additive + graceful degradation — not a breaking change for real (matched-version) users.** An older agent/relay that never acks lapses a short 2s window and the client falls back to optimistic success. All four packages release together (fixed changeset group), so skew is transient.

- The `run_flow` driver shares this client, so its input becomes truthful too; the flow-runner **CLI** (separate `RelayClient`) is a follow-up.
- Also resolves the H-E review MINOR#1 (input:type false-success on a shut-down sim is now covered by the ack family).

## Verification

- New tests: client ack round-trip (**done → success, error → throw, no-ack → optimistic**), ios-agent `input:done`/`input:error` acks, android-agent `input:done` ack. Relay forwarding is covered end-to-end (agent tests drive a real `RelayServer` and receive the ack on the browser socket).
- Full suites green — relay 480, mcp-server 38, ios-agent 229, android-agent 173. `pnpm typecheck` + lint clean.
- **Adversarial review** by an independent context: no BLOCKER/MAJOR, ship-able; 3 non-blocking notes (stale-ack NIT, 2s×N skew penalty by-design, iOS unknown-key now surfaced). Record: `.work/reviews/feat__mcp-input-ack.md`.

## Summary by CodeRabbit

* **New Features**
  * Input actions now acknowledge `done` vs `error` based on per-session device readiness.
  * Terminal input acknowledgements are forwarded so the client can reliably confirm completion.

* **Bug Fixes**
  * Improved acknowledgements for touch, pinch, key, and button inputs during boot/shutdown and after reconnects.
  * Clear errors now returned for agent-offline sessions and unknown key codes.

* **Tests**
  * Added/updated tests covering booted vs not-booted acknowledgements, client ack-wait behavior, and agent-offline error responses.

* **Documentation**
  * No user-facing documentation changes.

### DIFF
--- a/packages/android-agent/src/AndroidAgent.ts
+++ b/packages/android-agent/src/AndroidAgent.ts
@@ -126,10 +126,7 @@ interface DeviceState {
   sessionId: string
   deviceId: string
   touchHelper: AndroidTouchHelper | null
-  // Whether this session's device is booted, for truthful input acks. Set when the
-  // device is ready, cleared on shutdown. Starts false after a reconnect (state is
-  // recreated) even if the emulator is still up — the input-ack path verifies once
-  // via adb and caches the result here.
+  // Device-booted flag for truthful input acks — set on device:ready, cleared on shutdown; false after a reconnect until the ack path re-verifies once via adb.
   booted: boolean
   streamWs: WebSocket | null
   scrcpySession: ScrcpySession | null
@@ -851,10 +848,7 @@ export class AndroidAgent implements DeviceAgent {
     }))
   }
 
-  // Ack a terminal input (tap/swipe/key/button) — mirrors ios-agent.ackInput.
-  // input:done means the input was dispatched to a booted device (not a guarantee it
-  // physically landed); input:error means there was no live pointer channel or the
-  // device wasn't booted, so nothing was dispatched.
+  // Ack a terminal input (mirrors ios-agent.ackInput): input:done = dispatched to a booted device (not a landing guarantee); input:error = no live pointer channel / not booted.
   private async ackInput(state: DeviceState, dispatched: boolean): Promise<void> {
     const booted = dispatched && (state.booted || (await this.isBooted(state.deviceId)))
     if (booted) state.booted = true // cache the post-reconnect verify so later inputs skip adb
--- a/packages/android-agent/src/AndroidAgent.ts
+++ b/packages/android-agent/src/AndroidAgent.ts
@@ -126,6 +126,11 @@ interface DeviceState {
   sessionId: string
   deviceId: string
   touchHelper: AndroidTouchHelper | null
+  // Whether this session's device is booted, for truthful input acks. Set when the
+  // device is ready, cleared on shutdown. Starts false after a reconnect (state is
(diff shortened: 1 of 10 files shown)


jo-duchan:
Thanks @coderabbitai — all four addressed (pushed in a059468):

**#3 (Major) `awaitInputAck` masks disconnects — Fixed.** The bare `catch` swallowed a `WebSocket closed` rejection as optimistic success. Now only a `'Request timed out'` falls back; close/other errors rethrow so a dropped relay connection fails the tool truthfully. Test added (drop the connection mid-input → `tap` rejects).

**#4 (Major) relay drops terminal input when the agent is offline — Fixed, with a broader scope.** Your instinct was right; I replied to the **sender socket** (`ws`) rather than `session?.browserSocket`, because the agent's `close` handler *evicts the session* (`evictAgentSocket` → `sessions.remove`), so the common case is `session === undefined`, not `session` present with a closed agent socket. Replying to `ws` covers both the not-`OPEN` and session-gone cases. Test added (close the agent, then a terminal input → the browser gets `input:error "agent offline"`).

**#2 (Minor) ios `input:error` test skips the boot flow — Fixed.** Rewrote it to boot → `device:ready` → shut the device down (flip the simctl mock to `shutdown`) → then the terminal input acks `input:error "device not booted"`. Now goes through `device:boot` first per the guideline, and exercises the truthful post-shutdown path.

**#1 (Minor) multi-line comments — Fixed.** Condensed all seven to single English lines.

Full suites green (relay 481, mcp-server 39, ios-agent 229, android-agent 173); typecheck + lint clean.

Issue #341, opened 2026-06-28
Symmetric host-mute on Android via a shared process-tap helper


jo-duchan:
## Background

With audio default-on, iOS mutes the host (agent Mac) via the process tap's muteBehavior=.muted — only the browser gets sound. Android can't do this natively: emulator's `-audio <backend>` couples play+record, and `-audio none` disables both (gRPC capture included), so there is no "host-output off, gRPC capture on". Measured: a headless emulator (-no-window, audio on) DOES output to the host Mac speakers.

So Android is asymmetric today: on = capture + host also audible (silence it with the Mac's volume).

## Idea

A Core Audio process tap works on any macOS process by PID, and the Android emulator (qemu) is a macOS process. So audiotap-helper with muteBehavior=.muted pointed at the emulator PID could silence its host output while gRPC streamAudio keeps doing the browser capture. That restores symmetry (browser-only on both platforms).

## Scope

- Share audiotap-helper so android-agent can use it (currently in packages/ios-agent/src) — shared location or cross-package reuse
- Add a mute-only mode to the helper (create the .muted tap, do not capture/stream — Android captures via gRPC)
- android-agent: enumerate the emulator (qemu) PID, launch the mute-only tap when audio is on
- Verify once: does a .muted tap on the emulator PID actually silence host output?
- Gate on macOS 14.2+ (same as iOS); fall back to "use Mac volume" below that

Context: PR #339 (audio default-on, asymmetric host-mute), contributing/simulator-audio.md.

jo-duchan:
Already implemented and released in **v0.11.0** — closing as done. The commit referenced the number but never closed it, so this stayed open for two months against shipped code.

Evidence:

- `6bd8ebe feat(android): symmetric host-mute via shared audiotap-helper package (#341)`, released in `d401460 chore: release v0.11.0`
- The helper is shared: `packages/audiotap-helper` is its own package, and `packages/android-agent/package.json` depends on `@tapflowio/audiotap-helper` (`workspace:*`) — same as `ios-agent`
- `--mute-only` mode exists in `audiotap-helper.swift` and its TS wrapper, has tests, and is invoked from `AndroidAgent.ts` / `EmulatorLauncher.ts`
- Documented in the present tense in `contributing/simulator-audio.md` — "Host output: both muted (iOS by capture, Android by a mute-only tap)" — including the macOS 14.2+ gate and the fall back to the Mac's volume below it

[...]

PR #412, opened 2026-07-25
fix(android): restore emulator-encoder exec bit so gRPC audio + host-mute work (M3)

jo-duchan:
## Summary

Fixes **followups M3** — with the dashboard and Android agent on **separate Macs**, the emulator's audio came out of the *agent* Mac and never reached the dashboard.

Root cause is a **packaging bug, not the mute logic**. On a fresh install the android-agent's `bin/emulator-encoder` (a Swift VideoToolbox H.264 encoder) came without an executable bit → the encoder spawn failed with `EACCES` → the gRPC video backend fell back to scrcpy. Because gRPC `streamAudio` capture **and** the #341 host-mute tap are only set up on the gRPC path, that fallback silently killed **both**:

- dashboard got no audio (no gRPC `streamAudio` capture)
- the agent Mac played the emulator's audio (host-mute never ran)

Diagnosed from the agent log on a 2-Mac repro: `emulator-encoder … EACCES … gRPC backend failed … falling back to scrcpy` (with no `host-mute:` line, because that path never runs `startHostMute`).

Why it hid until now: on **one Mac** the local emulator playback masked it, and dev runs use the repo's already-executable binary. **iOS was unaffected** — its `.muted` Core Audio tap captures + mutes in a single step, so the two can't diverge.

## Fix

- **`ensureExecutable()`** restores the exec bit right before spawning the encoder — the load-bearing runtime safety net that self-heals **however** the bit was lost (including `--ignore-scripts` installs that skip the postinstall).
- **`postinstall`** chmods `bin/emulator-encoder` to `0o755` at install time, matching `ios-agent` / `audiotap-helper`.

> Note: `npm pack` preserves the `0o755` bit in the published tarball, so the bit was dropped somewhere in the install path (global-install layer / extraction tooling / package-manager store), not by the tarball itself. `ensureExecutable` is the piece that stands on its own.

## Follow-up (not this PR)

Audio + host-mute die silently on **any** gRPC→scrcpy fallback, not just this EACCES — worth a user-visible degraded-audio signal, or wiring host-mute/audio independent of the video backend.

## Verification

- New tests: `ensureExecutable` restores the bit / no-ops when already set / doesn't throw on a missing binary (exercises the real exported function).
- android-agent suite **176** green; `pnpm typecheck` + lint clean.
- **Adversarial review** by an independent context: no BLOCKER/MAJOR, safe to merge; one MINOR (narrative precision) fixed by amending the commit. Record: `.work/reviews/fix__android-encoder-exec-bit.md`.

> This is a published-package bug, so the fix reaches users on the next release.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android emulator video startup by best-effort restoring executable permissions for the encoder to prevent permission errors.
  * Added a macOS installation safeguard to automatically set the required permissions.
  * Updated behavior to handle cases where the encoder is already executable or the target is missing without breaking existing fallback handling.
* **Tests**
  * Added coverage for executable-permission restoration, including no-op and missing-binary scenarios.

### DIFF
--- a/packages/android-agent/package.json
+++ b/packages/android-agent/package.json
@@ -46,6 +46,7 @@
     "access": "public"
   },
   "scripts": {
+    "postinstall": "node -e \"if (process.platform === 'darwin') { try { require('fs').chmodSync('bin/emulator-encoder', 0o755) } catch {} }\"",
     "build": "tsc -b",
     "typecheck": "tsc -b",
     "test": "vitest run",
--- a/packages/android-agent/src/__tests__/EmulatorVideo.test.ts
+++ b/packages/android-agent/src/__tests__/EmulatorVideo.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect, vi } from 'vitest'
-import { EmulatorVideo, detectCornerRadius, type EncoderProcess } from '../emulator/EmulatorVideo'
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+import { EmulatorVideo, detectCornerRadius, ensureExecutable, type EncoderProcess } from '../emulator/EmulatorVideo'
 import { EmulatorGrpcClient, type RawEmulatorController } from '../emulator/EmulatorGrpcClient'
 import type { ScrcpyFrame } from '../scrcpy/ScrcpyVideo'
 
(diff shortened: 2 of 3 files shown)


jo-duchan:
Thanks @coderabbitai — addressed in 20007f4.

**Comments one line (Minor) — Fixed.** Condensed the `ensureExecutable` rationale (`EmulatorVideo.ts`) and the test scenario comment (`EmulatorVideo.test.ts`) to single English lines. I also dropped the `--ignore-scripts`/"leaves it without the exec bit" framing from the code comment to match the corrected root-cause narrative (npm preserves the `0o755` bit in the tarball; the bit was lost elsewhere in the install path, and `ensureExecutable` is the runtime safety net that self-heals regardless).

The ast-grep `child_process` note is informational — the spawn already uses an argument array (`spawn(ENCODER_BIN, [String(f)])`) with no shell and no user-derived command string, so there's no injection surface.

android-agent suite 176 green; typecheck + lint clean.

PR #420, opened 2026-07-26
fix(clipboard): restore the runner injection point, align the deadlines, and answer before restoring

jo-duchan:
## Summary

Follow-ups the clipboard bridge review (#418) deferred rather than folded in. No user-facing behaviour changes on the happy path; what changes is what happens when a copy fails.

- `SimctlWrapper.getPasteboard`/`setPasteboard` went around the `SimctlRunner` injection point, so clipboard calls were the only simctl calls without CoreSimulatorService version-mismatch auto-recovery — and could not be unit tested without spawning a real process. The runner contract now carries stdin and bounds (`execWithOpts`), and both are covered.
- The browser gave up at 3000ms while the agent's slow path also ended at 3000ms, so the agent's specific message ("did not copy anything — is something selected?") arrived after the browser had already shown a generic one. The timings now live in `agent-core` and the browser derives its budget from them.
- On a failed copy the agent used to restore *before* answering, putting the restore window inside the caller's round trip. It answers first and restores after, still holding the per-device queue.

Plan: `.work/2026-07-27-clipboard-followups-plan.md`.

## `sentinelParked`

While a copy is in flight the agent parks a marker on the device clipboard. If the copy fails, the viewer used to press the plain chord as a fallback so the copy at least landed on the device. That is only safe when no marker is parked — otherwise the agent's restore lands a moment later and overwrites what the chord just copied, handing the user a stale value.

`clipboard:error` now carries `sentinelParked`, and the viewer presses the chord only when the agent says nothing is parked (or that the backend has no clipboard channel at all, which cannot park anything). **Absent means assume parked** — an agent that predates the field cannot answer, and a silent stale paste is the worse half of that trade.

The flag counts markers **per device, not per operation**. Several sessions and MCP can address one simulator, and the fallback chord travels as `input:key`, outside the queue that keeps clipboard operations apart — so an operation that never touched the clipboard still has to report a marker another one holds.

## Also fixed here

iOS treated every failed pasteboard read during the copy wait as "the app copied" and skipped the restore. Only the size ceiling means that. A timed-out `pbpaste` left the sentinel parked permanently: the user's clipboard destroyed, and an invisible zero-width marker pasted into the app under test. The ceiling is now its own error type. This predates the branch, but the comment claiming Android behaved identically was false, and this is the branch about making the two read paths match.

## Verification

`pnpm test` **1687 passed | 1 skipped** across 12 packages; `pnpm typecheck` and `pnpm lint` clean. Both agent suites run 3× to check for flake.

**Adversarial review: 4 rounds**, the last escalated to two independent channels (wire contract; runtime correctness) because the error payload is a protocol change. Every round found real defects and every fix exposed the next — recorded round by round with dispositions in `.work/reviews/fix__clipboard-followups.md`.

What the last two rounds caught is worth stating plainly, because it is the same shape three times: a guard written to close a hole had the same hole. The budget-drift test read only the first line of a declaration that already wraps, so a term added on a continuation line was invisible — and the value it missed was 5000ms, equal to the browser budget, which is the exact defect that test exists to prevent. The budget tests themselves borrowed their environment from a sibling `describe` and passed at any budget when run alone. And the `ClipboardErrorPayload` interface added to document the wire shape was never exported, so `{ sentinelPARKEDtypo: true }` compiled.

Every changed behaviour is mutation-verified: breaking it deliberately fails the test that claims to protect it, and no other.

Reviewers' remaining items are filed rather than folded in: the fallback chord still bypasses the queue for the ~60ms it is in flight (closing that means the agent pressing it itself, which removes the field entirely); no test file in the repo is type-checked or linted, which is why a duplicate object key passed both gates; and `packages/dashboard` lints `src` but not `hooks`.

## Summary by CodeRabbit

* **New Features**
  * Clipboard operations now coordinate overlapping requests safely per device, using a keyed per-device queue.
  * Clipboard bridge errors now include additional status (e.g., whether a sentinel marker is still parked).
  * Clipboard timing/timeout behavior is now budgeted from a shared worst-case constant for more consistent slow-device handling.

* **Bug Fixes**
  * Prevented stale/temporary sentinel markers from being returned or restored as user clipboard content.
  * Improved iOS simulator clipboard reliability with better bounded execution and a dedicated “too large” clipboard error classification.


jo-duchan:
CodeRabbit 지적 두 건 모두 반영했습니다 — `4a87044`.

**재시도가 실제 원인을 삼키는 문제(MAJOR)는 이 PR이 만든 것입니다.** 두 커밋 전에 `OutputTooLargeError` 분류를 도입하면서, 재시도 경로에서 그 타입이 지워지는 걸 놓쳤습니다. CoreSimulatorService 재시작 후의 재시도가 maxBuffer로 실패하면 `coreSimServiceError()`로 뭉개지고, `getPasteboard`의 `instanceof`가 빗나가고, 에이전트는 "복사 안 됨"으로 판단해 **사용자가 방금 복사한 값 위에 이전 값을 복원**합니다. 분류를 도입한 목적과 정반대입니다.

제안된 대로 고쳤고, `exec`/`execBinary`에도 같이 적용했습니다. 그 둘은 `instanceof` 소비자가 없어 메시지 품질 문제에 그치지만, 세 메서드가 같은 파일 몇 줄 간격으로 붙어 있어 한둘만 고치는 게 이 브랜치가 반복해서 데인 패턴입니다.

**stderr 상한(MINOR)은 판단을 바꿨습니다.** 직전 독립 리뷰에서도 같은 지적이 나왔는데 그때는 "의도와 일치"로 스킵했습니다. 다시 보니 아닙니다 — 상수가 이름으로 선언한 경계를 실제로 지키지 않았습니다. 제안대로 남은 예산만큼 잘라 담습니다.

둘 다 회귀 테스트를 붙이고 mutation으로 확인했습니다. 되돌리면 각각 정확히 해당 테스트만 실패합니다.

`pnpm test` **1689 passed | 1 skipped**, typecheck·lint 클린. 리뷰 기록도 현재 HEAD로 갱신했습니다.

PR #424, opened 2026-07-27
docs(changeset): backfill release notes for four merged fixes

jo-duchan:
## Summary

Four merged PRs changed published code without a changeset. As it stands, 0.17.0's changelog would describe only the clipboard bridge and say nothing about the fixes underneath it.

| PR | | changeset |
|---|---|---|
| #418 | clipboard bridge | present |
| #413 | Android Cmd/Ctrl+C/V/X typed the letter instead of copying | **missing** |
| #412 | `emulator-encoder` exec bit — no dashboard audio, emulator sound on the agent Mac | **missing** |
| #411 | MCP input tools always reported success | **missing** |
| #410 | iOS input silently dropped after an agent reconnect | **missing** |

The other six merged since v0.16.0 (#417, #415, #414, #409, #408, #407) are docs, chore or test-only and correctly have none.

#413 is the one that stings: it is what made Android clipboard shortcuts work at all, and the bridge in #418 builds directly on it. A user reading the release notes would see the bridge announced and no mention that the shortcuts themselves were broken until now.

## No version change

`npx changeset status` still resolves to **minor → 0.17.0**. The packages are one `fixed` group and the clipboard changeset already requires minor, so these four only add changelog entries.

## Verification

Each entry was written from the originating PR's commit body rather than from the diff summary. One draft claimed #412 restored the executable bit "at startup"; it actually restores it just before spawning the encoder, and the wording was corrected.

Adversarial review skipped — docs-only, no executable code in the diff. Recorded with the reason in `.work/reviews/chore__backfill-changesets.md`, per AGENTS.md.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Android clipboard shortcuts so Cmd/Ctrl+C/V/X trigger copy/paste/cut correctly without breaking other modifier-key commands.
  - Restored Android emulator audio capture and correct routing after fresh installs.
  - Fixed iOS sessions where tap, swipe, keystroke, and button input could be dropped after an agent reconnect.
- **New Behavior**
  - MCP input tools now report the actual terminal result (acknowledged dispatch success or error) instead of always indicating success.
- **Documentation**
  - Added/updated release notes for these fixes and behavior changes.

PR #427, opened 2026-07-27
chore: release v0.17.0 — copy and paste cross between the dashboard and the device

jo-duchan:
## v0.17.0 — copy and paste cross between the dashboard and the device

| package | version |
|---|---|
| `tapflow` (CLI) · `@tapflowio/agent-core` · `ios-agent` · `android-agent` · `relay` · `mcp-server` · `flow-runner` | **0.17.0** |
| `@tapflowio/audiotap-helper` | 0.2.7 |

`@tapflowio/dashboard` and `playground` are in the changesets `ignore` list and are untouched.

## What users get

**Copy and paste now cross the boundary.** Cmd/Ctrl+V sends your clipboard to the simulator or emulator and pastes it there; Cmd/Ctrl+C brings what you copied on the device to your own clipboard in one press. Neither direction existed before, so accounts, tokens and deep links had to be retyped by hand.

Paste works everywhere, including plain-HTTP LAN. Copy needs the dashboard over HTTPS or localhost: proving the copy actually landed takes a round trip, and no clipboard API available on plain HTTP accepts a value that arrives that late. On plain HTTP the copy still reaches the device and the dashboard says why it stopped there.

The agent presses the device-side chord itself and confirms the clipboard actually changed before answering, so a slow or busy device cannot hand back its previous contents as if freshly copied.

**MCP input tools stopped lying.** `tap`, `swipe`, `press_key` and `press_button` were fire-and-forget — against a session whose device was not booted the input was dropped and still answered `{tapped: true}`, which also made parallel test results untrustworthy.

**Three fixes underneath.** Cmd/Ctrl+C/V/X on Android typed the letter instead of copying. Android audio never reached the dashboard while the emulator's sound played out of the agent Mac's speakers — only visible with the dashboard and agent on separate Macs. iOS silently dropped every tap and keystroke after an agent reconnect, while still looking responsive because screenshots and UI-tree reads travel a different path.

## Compatibility — verified, minor is enough

Two protocol additions. Checked against existing tests rather than a fresh e2e.

**Clipboard messages** are gated on an advertised capability:

| test | what it pins |
|---|---|
| `useClipboardBridge.test.tsx:199` | the viewer stays inert against an agent without the capability |
| `IOSAgent.test.ts:1265`, `AndroidAgent.test.ts:1042` | the agent advertises it end to end, agent → relay → viewer |

An agent that does not advertise `clipboard` is never sent these messages, and the viewer keeps forwarding the shortcuts as plain key input. **v0.16.0 contains no clipboard code at all**, so every released agent takes that path.

**Input acks** treat a timeout as an older peer — `mcp-server/src/client.ts:188`: *"Only a timeout means an older agent/relay that never acks → assume dispatched."* A dropped connection is distinguished and fails loudly (`client.test.ts:194`).

## Verification

All five changesets consumed; the seven fixed-group packages moved together. `pnpm typecheck` and `pnpm lint` clean. **No lockfile change** — internal dependencies are `workspace:*`, so a change there would have been a red flag.

The code in this release was reviewed as it landed: `feat__clipboard-bridge` (7 rounds), `fix__clipboard-followups` (5), `chore__dev-and-changeset-guards` (4). Review skipped for this PR as a version bump, recorded with the reason in `.work/reviews/release__v0.17.0.md`.

Worth noting that the new changeset gate passed this PR through its release path — `Release branch: 5 changeset(s) consumed into a changelog.` That exemption was added days ago after discovering the gate would otherwise block the release PR it exists to serve.

## After merging

Merging does **not** publish. `release.yml` fires on a `v0.17.0` tag only.

## Summary by CodeRabbit

* **New Features**
  * Added bidirectional clipboard sharing between the dashboard and devices.
  * MCP input tools now report actual completion or failure results.
  * Added MCP tools for installing builds before replay and shutting down devices.
  * Agents now advertise supported capabilities.

* **Bug Fixes**
  * Fixed Android keyboard shortcuts, dashboard audio, and host audio leakage.
  * Prevented iOS input from being lost after agent reconnects.

* **Release**
  * Released version 0.17.0 across the platform packages.


jo-duchan:
Skipping this one — it is changesets' own output format, not a defect in this release.

changesets emits **one `Updated dependencies` line per changeset**, then attaches the resulting version list once at the end. v0.16.0 has the same shape with a single line, because that cycle had a single changeset:

```
### Patch Changes

- Updated dependencies
  - @tapflowio/flow-runner@0.16.0
  - @tapflowio/agent-core@0.16.0
  …
```

There are five lines this time because there are five changesets. Regenerating produces the same thing — the input has not changed — and hand-editing generated output would drift from the tool and be overwritten next release.

The repeated `eaa78ac` is explained by the same mechanism: changesets records the commit that **added** each changeset, and four of the five were backfilled together in #424 (release notes for #410–#413, which had merged without one). Only the clipboard changeset carries its original commit, `661356e`.

The concern about the notes reading oddly is fair, but this file is the per-package machine-generated log. What a user actually reads is the root `CHANGELOG.md` and the GitHub Release body, both hand-written for this release — Added / Fixed / Changed, no dependency bookkeeping.

PR #429, opened 2026-07-27
fix(scripts): let a backfilled changeset clear the merge it covers

jo-duchan:
## Summary

`pnpm changeset:audit` judges each merge on its own, so a changeset written by a **later** PR to cover an earlier one never cleared it. Through the whole v0.17.0 cycle it reported #410–#413 as gaps after #424 had already written their release notes — and telling a real gap from a filled one meant matching changesets to merges by hand on every run, which is the manual step the audit exists to remove.

A backfill now says what it answers for, on its own line:

```
Backfills: #413
```

Strict on purpose. The keyword must open the line, outside any code fence and outside the frontmatter, and the claim only counts while the changeset that made it is still alive. Every loosening of those rules lets a real gap go unreported, and a silently cleared gap means a shipped change with no release note.

## What the review changed

The first version had a blocker and two holes, all clearing gaps silently.

**Deletions stopped counting.** I narrowed the merge's file list to added/changed, which dropped removals from the check — a merge that *deletes* a shipped file, the highest-consequence change there is, audited clean. Nothing about this feature needed that narrowing; it belongs on the changeset list, not the file list. All 99 tests passed with the regression in.

**A quoted example counted as a claim.** `extractReason`, forty lines up in the same file, carries a fence-and-indent guard with a comment saying it was learned the hard way: a body that merely *quotes* the marker must not trip it. `parseBackfills` reproduced the hole exactly — and the audit's own failure message prints `Backfills: #413` indented, while CONTRIBUTING, AGENTS and `/release` all carry it verbatim. The guard is shared now rather than written twice, which is what invited the drift.

**A claim was never scoped or re-checked.** It could pre-clear a merge that landed after it, and it survived deleting the changeset that made it — claim coverage, drop the changeset in a follow-up, and the gap stays quiet for the rest of the cycle.

Also: a renamed changeset counts; a bad `--audit` revision exits 2 rather than 1, so a typo cannot look like a finding; the gap report goes to stderr like the PR gate's.

## Verification

`pnpm test` **1793 passed | 1 skipped** — 104 tooling tests plus every package. `pnpm typecheck` and `eslint scripts` clean, tooling suite run 3× with no flake.

The integration tests build a throwaway git repository and drive the real CLI. That is what caught the two wiring bugs the parsers could not: reading changesets at a merge that had already deleted them, and counting only *added* ones so a PR that amended an existing entry read as a gap. Five mutations, five tests, one each.

Review record with findings and dispositions — including two knowingly skipped (back-merge commits carry no PR number; squash merges are invisible to `--merges`, both pre-existing and inert for this repo's history) — in `.work/reviews/fix__changeset-audit-backfills.md`.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how to backfill changesets for earlier pull requests, including adding a `Backfills: #<PR_NUMBER>` line so release auditing can attribute the coverage correctly.

* **Bug Fixes**
  * Improved audit accuracy for missing changesets, backfilled/renamed/amended changesets, and release merges.
  * Strengthened `Backfills:` detection (case/indentation/prose-safe) and updated audit reporting output behavior.

* **Tests**
  * Added unit and integration coverage for backfill parsing and end-to-end audit gaps/backfill resolution scenarios.


jo-duchan:
Fixed — `b97f6f6`. You are right, and the split I claimed in the last commit did not actually hold.

Everything around it had moved to `console.error`; that one block had not, because my replacement pattern did not match the way the template literal wraps and I never checked that it applied. So `2>/dev/null` still leaked the guidance text, and a caller reading stdout for a clean signal got output on a failing run — exactly the two cases the change was for.

The convention is now stated in the code and pinned by two tests rather than left implicit:

```
gaps   → exit 1, stdout empty, whole report on stderr
clean  → exit 0, stdout = "Every merge that changed published source carries a changeset."
```

Reverting the line fails the first of them.

Verified end to end rather than by reading:

```
$ node scripts/check-changeset.mjs --audit v0.16.0 >/tmp/o 2>/tmp/e
  exit=1  stdout=[]  stderr=25 lines
$ node scripts/check-changeset.mjs --audit >/tmp/o 2>/tmp/e
  exit=0  stdout=[Every merge that changed published source carries a changeset.]
```

106 tooling tests pass, typecheck and lint clean.

Issue #440, opened 2026-07-30
ios-agent: app install intermittently fails with "No devices are booted" — simctl calls target the `booted` alias, not the session's udid


jo-duchan:
Observed during manual E2E, and reported as something that has happened intermittently for a while — not a one-off.

## Symptom

Starting a QA session on a freshly booted simulator, the build install fails:

```
Install failed: Command failed: xcrun simctl install booted /var/folders/.../tapflow-install-<uuid>/Food Truck.app
No devices are booted.
```

Re-entering the session from the dashboard installs successfully against the same simulator. So the install path works; what fails is *when* it runs and *what it targets*.

### 1. The install can run before the device is actually `Booted` (the race)

`simctl boot` returns before the device reaches the `Booted` state, and the relay replays readiness rather than re-deriving it. `packages/ios-agent/AGENTS.md` already records the latter:

> **`device:ready` is not a sync point.** With `booted=true` the relay registers the session as already booted and **replays a `device:ready` on `session:start`** (browser-reconnect support)

`DeviceViewer` installs as soon as it sees `device:ready` (`if (buildId) { setInstalling(true); send({ type: 'app:install', ... }) }`), so a replayed or early ready sends the install at a moment when no device is booted yet. That matches the observed failure exactly, and matches why a second entry succeeds — by then the device really is up.

Worth checking whether `simctl bootstatus` (or polling `listDevices()` for `status === 'booted'`, which `IOSAgent.isBooted` already does) should gate `device:ready`, so readiness means the device is usable rather than that a command was issued.

### 2. Every simctl app command targets `booted`, not the session's udid (the structural half)

`packages/ios-agent/src/SimctlWrapper.ts` uses the `booted` alias in six places:

| line | call |
|---|---|
| 140 | `uninstall booted <bundleId>` |
| 147 | `terminate booted <bundleId>` |
| 148 | `get_app_container booted <bundleId> data` |
| 161 | `install booted <appPath>` |
| 167 | `launch booted <bundleId>` |
| 211 | `io booted screenshot` |

The agent knows the session's udid — `DeviceState` tracks it, `ScreenCaptureStreamer` and `touch-helper` are both given it explicitly — but these calls throw it away and ask simctl for "whatever is booted".

[...]

Issue #439, opened 2026-07-30
Full reset stays armed after use, and erase fails on an already-booted device


jo-duchan:
> **Rewritten twice.** Filed from code reading as "a stale `resetMode` after reconnect", then rewritten after a reproduction as "`resetMode` is re-sent on every boot". Both were wrong in the same direction — they blamed the delivery path. Reproducing it properly showed two separate defects, neither of which is about staleness. Earlier versions are at the bottom.

Two defects that are only dangerous together.

## 1. The toggle stays armed after it is used

`resetMode` is `useState` in `useDeviceSelector`, and `QASession` swaps screens by conditional render rather than routing — so leaving a session with `← All Macs` does not unmount anything, and `handleBackToMacs` does not touch it:

```ts
setActiveSessionId(null); setSelectedAgent(null); setBooting(false); setStatus('')
// resetMode lives in a different hook and stays as it was
```

So:

1. Turn **Full reset** on, pick **device A** → A is erased and boots. Correct.
2. Go back to the device list (**not** a browser refresh).
3. Pick **device B** → the toggle is still on → **B is erased too.**

Nobody asked for B to be wiped. The toggle is visible on that screen, so this is not invisible — but nothing distinguishes "apply to this launch" from "a setting that stays on", and after using it once the natural reading is that it is done.

A browser **refresh** is not affected: it drops the component state and the toggle returns to `app-only`. Only in-app navigation carries it.

## 2. `erase` is called without checking device state

```ts
// IOSAgent.ts handleDeviceBoot
if (fullErase) {
  await this.simctl.erase(deviceId)        // throws if the device is Booted
  await this.simctl.boot(deviceId)
} else if (target.status !== 'booted') {
  await this.bootWithZombieRecovery(deviceId)
}
```

`simctl erase` requires a shut-down device. Ask for Full reset on a device that is already running and the boot fails outright:

```
Boot failed: Command failed: xcrun simctl erase <udid>
```

Reproduced: with Full reset on and the device already booted, the viewer shows that error and never streams. `launchd_sim` kept its PID (the erase threw before `boot` was reached).

A device is often already running — it stays up across agent restarts (#426), and across sessions that ended without a clean shutdown.

## Why they have to be fixed together

**Defect 2 is currently containing defect 1.** The unwanted erase from a stale toggle usually targets a device that is already booted, so it throws and nothing is destroyed — a loud failure standing in front of a silent one.

[...]

Issue #422, opened 2026-07-26
test: no test file is linted, and only the dashboard's are type-checked


jo-duchan:
Found while reviewing #420: a duplicate object key sat in `packages/ios-agent/src/__tests__/SimctlWrapper.test.ts` and passed both gates.

Two independent exclusions overlap:

- `packages/*/tsconfig.json` has `"exclude": ["src/__tests__", …]`
- `eslint.config.mjs` globally ignores `**/__tests__/**`

So a test can reference a field that does not exist, silently shadow an object key, or drift from the interface it exercises, and nothing reports it. Test doubles are exactly where that matters: `#418` already shipped a `as RawEmulatorController` cast whose double was missing two members, and the same class of gap hid the missing `execWithOpts` on seven iOS doubles.

## Scope note

Turning either gate on will surface a backlog. Worth measuring first (`tsc --noEmit` with the exclusions lifted, `eslint` with the ignore removed) and deciding whether to fix or to ratchet — typecheck first is probably the higher value of the two, since the drift it catches is the kind that makes a test pass for the wrong reason.

jo-duchan:
Half of this title was wrong. Measured during #456:

- **Lint — accurate.** `eslint.config.mjs:17` globally ignores `**/__tests__/**`, so no test file in any package is linted. Confirmed by planting an unused variable in a dashboard test and a relay test; both report "File ignored because of a matching ignore pattern."
- **Type-check — not accurate.** `packages/dashboard/tsconfig.json` has no exclude for `src/__tests__`, so its 28 test files *are* checked. The other eight packages with tests (119 files) exclude them.

Retitled to match. The measurement and what the gap costs are written up in #454, which covers the type-check half in detail — worth deciding whether these two stay separate or #454 folds in here.

jo-duchan:
Folding in #454 (closed as covered by this one), which measured the type-checking half of this and demonstrated it.

## Measured scope

**119 of 147 test files run with their types stripped.** 8 of the 10 packages exclude `src/__tests__` from `tsc`; the root `typecheck` is `tsc -b` plus a few `--filter` runs that cover `src/` only. Test files are transformed by vite at run time and their types are never checked.

Demonstrated on #453: planting `const bogus: number = reply.registeredSessions` in a test file and running `pnpm typecheck` reports **zero errors**.

Only the dashboard's tests are type-checked, which is why its `tsc --noEmit` is a separate `--filter` entry in the root script rather than part of `tsc -b`.

[...]

PR #441, opened 2026-07-30
chore(dashboard,ios-agent): lint the whole dashboard source, drop the dead Simulator hide

### LINKED ISSUE #423: packages/dashboard lints src but not hooks
Found while reviewing #420.

`packages/dashboard`'s lint script is `eslint src`, and `hooks/` is a sibling of `src/`. The entire clipboard bridge, the relay client, the decoder stream and the recording hook live there and are never linted.

Running `npx eslint hooks` today reports 4 errors and 2 warnings across `useClientRecording.ts`, `useDecoderStream.ts`, `useRelay.ts`, `useBreadcrumb.tsx` and `useDeviceSelector.ts` — mostly `react-hooks` rules, which is the category that actually bites in this directory. `useClipboardBridge.ts` itself is clean.

Fix is `eslint src hooks` plus whatever those six turn out to need.

### LINKED ISSUE #431: ios-agent: remove the dead Simulator.app hide, or make it work
Carried out of the v0.14.0 follow-ups plan, which is now closed. Confirmed still present.

`SimctlWrapper.boot` ends with:

```ts
// packages/ios-agent/src/SimctlWrapper.ts:131
execFile('osascript', ['-e', 'tell application "Simulator" to set visible to false'], () => {})
```

The callback swallows every error, and on this environment (macOS Tahoe / Xcode 26) the call fails with `-10006` every time — `simctl boot` no longer auto-opens Simulator.app, so there is no application to address. The comment above it still says "Xcode 14+ auto-opens Simulator.app on `simctl boot`", which is why the line is there and why it is now wrong.

So the line does nothing, and its silence means nobody would notice if the assumption changed back.

## Decide which

- **Remove it.** Correct here, and the occlusion throttle that made hiding matter only applies to a window that exists. Risk: an Xcode/macOS combination that *does* auto-open would leave a window on screen.
- **Make it work** via System Events (`set visible of process "Simulator" to false`), which addresses the process rather than the app and works whether or not the app scripting interface is up.

Check the supported-environment matrix before choosing. If any supported Xcode still auto-opens, the second option is the one; otherwise the first.

Either way the stale comment at ~line 129 goes with it.

## Verify

Boot an iOS agent and stream after the change: no window management regression (no occlusion throttle), tree queries still fine. That is the whole surface — the call has no other caller.

jo-duchan:
Closes #423
Closes #431

## Why

`packages/dashboard`'s lint script was `eslint src`, and `hooks/`, `components/` and `lib/` are siblings of `src/`. So **83 of its 123 source files were never linted**, while tsconfig type-checked all four directories. #423 reported the `hooks/` half; measuring first showed the hole was three times wider and hid two errors in `components/` nobody had seen.

## What changed

**Gate** — `lint` now covers `src hooks components lib *.config.ts`, matching what tsconfig already checks. `eslint.config.mjs` needed no rule changes; only the CLI argument was narrow.

**hooks (6 findings)** — render-time ref writes moved into effects (`useRelay`, `useDecoderStream`); the recursive `rafLoop` became a named function expression; an unused import; one Fast Refresh suppression with its reason.

**components (18 findings)** — 6 missing `recordCanvasRef` deps (a stable `useRef` return, so the effect cadence is unchanged); `load` wrapped in `useCallback` so its effect dependency is honest — listing it as-is would have refetched on every render; an unused import; and two false positives suppressed per line with the evidence recorded rather than restructured.

**`lib/` extraction** — `avatarColors` and the perf-notice policy moved out of their component files. The Fast Refresh warning was pointing at a real structure problem: a test was importing a component file to reach one of them.

**shadcn** — `react-refresh/only-export-components` is waived for `components/ui/**` only. Every correctness rule still applies there, which is what caught a `Math.random()` during render in `sidebar.tsx`.

**ios-agent** — `SimctlWrapper.boot` ended with an `osascript` call to hide Simulator.app whose callback swallowed every error. On the only supported Xcode (26.x, per `docs/guide/requirements.md`) `simctl boot` does not open Simulator.app. Also added the mirror guard to `shutdown`, which had none: tearing down an already-stopped device logged `shutdown failed` and was indistinguishable from a device that genuinely refused to stop.

## Deliberately not fixed here

`DeviceViewer` can send a stale `resetMode` after a reconnect. Adding the dep is a behaviour fix, not a lint fix, and it is unreproduced — tracked in **#439**, suppressed on that one line with a pointer.

## Verification

- dashboard lint clean across all 123 files; root lint / typecheck / test green (dashboard 236, ios-agent 275, cli 246)
- **Gate mutation-tested per directory**: a planted violation in `hooks/`, `components/`, `lib/` and `vite.config.ts` is caught each time; the three per-line suppressions still catch unrelated errors, so none of them over-waives
- `shutdown` guard mutation-tested: removing it fails exactly the new test and leaves the other 33 passing
- **Manual E2E**: iOS and Android streaming, tap/swipe, 10s recording → upload → download, Android rotation. #431 verified the decisive way — quitting Simulator.app entirely and booting from the dashboard produces no Simulator process

## Review

Three passes, recorded in `.work/reviews/chore__lint-hooks-scope-and-dead-simulator-hide.md`: a design review of the plan before implementing (this spans two packages), a diff review of the initial hooks change, and a pre-PR diff review of the full branch across two parallel lenses. No blockers or majors; three minors, all fixed — the last two in `34634c8`.

The design pass changed the plan: two `components/` errors were going to be code edits, and it established that one was a false positive on the Android rotation path and the other sits in an export that never renders. Both became suppressions with recorded reasons.

## Also found while testing

**#440** — build install intermittently fails with `No devices are booted`; re-entering the session succeeds. Two causes: a race against `device:ready` replay, and six `simctl` calls targeting the `booted` alias instead of the session's udid.

## Summary by CodeRabbit

* **Bug Fixes**
  * iOS simulator shutdowns now complete cleanly when a device is already stopped.
  * Removed an unnecessary simulator-window hiding step that could obscure boot failures.
  * Improved dashboard reliability for comment loading, recording/canvas syncing, and real-time relay updates.

* **Refactor**
  * Centralized avatar color generation and performance-notice auto-show/dismiss behavior for consistent dashboard use.

* **Tests**
  * Expanded coverage for simulator shutdown error handling and performance-notice behavior.

* **Chores**
  * Broadened dashboard lint coverage and clarified code-quality configuration.

### DIFF
--- a/.changeset/dead-simulator-hide.md
+++ b/.changeset/dead-simulator-hide.md
@@ -2,4 +2,6 @@
 "@tapflowio/ios-agent": patch
 ---
 
-Remove the dead `Simulator.app` hide from `SimctlWrapper.boot`. On the supported Xcode (26.x) `simctl boot` does not open Simulator.app, so the `osascript` call failed with `-10006` on every boot while its callback swallowed the error — it hid nothing, and its silence meant nobody would notice if the assumption changed back. The occlusion throttle it was guarding against only applies to a window that is on screen.
+Remove the dead `Simulator.app` hide from `SimctlWrapper.boot`. On the supported Xcode (26.x) `simctl boot` does not open Simulator.app, so the `osascript` call failed with `-10006` on every boot while its callback swallowed the error — it hid nothing, and its silence meant nobody would notice if the assumption changed back. The occlusion throttle it was guarding against only applies to a window that is on screen. Verified by quitting Simulator.app entirely and booting from the dashboard: no Simulator process appears.
+
+Also make `SimctlWrapper.shutdown` tolerate an already-stopped device, mirroring the guard `boot` has had. Tearing down a session whose device is already `Shutdown` raised `code=405 / Unable to shutdown device in current state: Shutdown`, so a routine teardown logged `shutdown failed` and became indistinguishable from a device that genuinely refused to stop.
--- a/.changeset/dead-simulator-hide.md
+++ b/.changeset/dead-simulator-hide.md
@@ -1,0 +1,5 @@
+---
+"@tapflowio/ios-agent": patch
+---
+
+Remove the dead `Simulator.app` hide from `SimctlWrapper.boot`. On the supported Xcode (26.x) `simctl boot` does not open Simulator.app, so the `osascript` call failed with `-10006` on every boot while its callback swallowed the error — it hid nothing, and its silence meant nobody would notice if the assumption changed back. The occlusion throttle it was guarding against only applies to a window that is on screen.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -60,6 +60,12 @@ export default tseslint.config(
   // structure we do not own and would have to re-fix on every upstream sync. Only that hint is
   // waived — every correctness rule still applies here, which is what caught the `Math.random()`
   // during render in sidebar.tsx.
+  //
+  // The scope is the directory, not the file's origin, and `ui/` also holds a few of our own
+  // components (`search-input`, `tech-label`). They export a single component today so the rule
+  // would not fire on them anyway — but the cost of this shortcut is that adding a hand-written
+  // file here that exports a component *and* a helper loses the hint silently. Put such a file in
+  // `components/` rather than `components/ui/`.
   {
     files: ['packages/dashboard/components/ui/**/*.{ts,tsx}'],
     rules: {
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -54,4 +54,16 @@ export default tseslint.config(
       'react-refresh/only-export-components': ['warn', { allowConstantExport: true }],
     },
   },
+
+  // shadcn primitives are generated and re-synced from upstream, and their convention is to export
+  // a `*Variants` helper next to the component. That trips the Fast Refresh hint in five files for a
+  // structure we do not own and would have to re-fix on every upstream sync. Only that hint is
+  // waived — every correctness rule still applies here, which is what caught the `Math.random()`
+  // during render in sidebar.tsx.
+  {
+    files: ['packages/dashboard/components/ui/**/*.{ts,tsx}'],
+    rules: {
+      'react-refresh/only-export-components': 'off',
+    },
+  },
 )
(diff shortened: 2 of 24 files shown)


jo-duchan:
Thanks — two of the three were valid and are fixed in `f962050`.

### 1. `resetMode` in the `handleMessage` deps — **skipped, deliberately**

This is the one finding the PR already carries a decision on. Adding the dep is a **behaviour change**, not a lint fix: it changes what `device:boot` sends. The failure is also unreproduced — reaching it needs `resetMode` to change while `DeviceViewer` stays mounted *and* a `session:joined` to follow, which is plausible on the reconnect path but has not been demonstrated.

Folding an unreproduced behaviour fix into a PR whose subject is "turn a lint gate on" would put it behind a chore review. It is tracked in #439 with the reproduction path to confirm first, and the suppression comment names the issue and says to remove the suppression when #439 lands.

### 2. Layout effect for the callback refs — **fixed**

Correct, and it is a regression this PR would have introduced. Writing the refs during render had no gap; moving them to `useEffect` created one, because passive effects run after paint and a socket message arriving in that window still reaches the previous `onMessage`/`onBinaryFrame`.

`useLayoutEffect` runs synchronously on commit, so no task can interleave. Applied to `useDecoderStream` as well — same shape, same window.

### 3. `execBinary` missing from the `SimctlRunner` doubles — **fixed, and it was four places**

`execBinary` is a required member (`packages/ios-agent/src/simctl.ts:43`). The two doubles flagged are the new `shutdown` tests, but they copied their shape from the pre-existing `boot` tests, which had the same hole — so four inline doubles were incomplete, not two. All four now carry it.

Worth noting *why* no gate caught this: `packages/*/tsconfig.json` excludes `src/__tests__`, so a test double can drift from the interface it stands in for and typecheck stays green. That is exactly the gap #422 describes, and this is now a concrete instance of it. Turning that gate on is a separate track (it surfaces ~117 eslint errors across 132 test files and needs a ratchet), but the case for it is stronger than it was this morning.

Verified after the fixes: dashboard lint clean across 124 files, root typecheck green, dashboard 236 / ios-agent 275 tests passing, and the test file type-checks clean when checked directly with the exclusion lifted.

Issue #426, opened 2026-07-27
Restarting an agent orphans the open dashboard tab — it spins on "Waiting for first frame..." forever


jo-duchan:
Restart a device agent while a dashboard tab is open on one of its devices, and that tab never recovers. It shows `Waiting for first frame...` indefinitely, input does nothing, and no message says why. Only a manual page refresh fixes it.

Pre-existing — `evictAgentSocket` behaves this way in v0.16.0 too. Not a regression from the clipboard work.

## Reproduction

1. Start the relay, then an agent.
2. In the dashboard, boot a simulator. Confirm tapping works.
3. Restart the agent process. Leave the relay running, leave the browser tab open, leave the simulator booted.
4. Wait. → `Waiting for first frame...` forever, device does not respond.
5. Refresh the page → back at the Mac list → pick the still-booted device → everything works.

## Why

When the agent socket closes, `evictAgentSocket` (`packages/relay/src/RelayServer.ts:737`) deletes every session that socket owned:

```ts
for (const s of agentSessions) this.sessions.remove(s.id)
```

`SessionManager.remove` (`packages/relay/src/SessionManager.ts:115`) drops the session from its indexes and deletes it. It does not notify the browser and does not close the browser socket. So:

- the browser's WebSocket stays open, still addressing a `sessionId` that no longer exists
- the restarted agent registers and creates **new** sessions with **new** ids
- nothing links the two, so everything the browser sends is dropped as unknown, and nothing streams back
- `IOSViewer.tsx:658` keeps rendering its placeholder because no first frame ever arrives

The refresh works because the dashboard then starts a fresh session against the new registration.

## Not the same as #410

#410 fixed the case where the session survives but the agent's `touchHelper` is null — screen streams, input is silently dropped. Here the session itself is gone, so the failure happens one step earlier and the screen is dead too. A repro that restarts the agent therefore cannot exercise #410: it hits this wall first.

## Direction — rebind rather than notify

The minimal fix is to tell the viewer its session ended so the dashboard can say "the agent disconnected" and return to the device list. That turns a silent failure into a stated one, which is worth having, but it still makes the user redo their navigation for something that should be invisible.

Rebinding is achievable because the pieces already exist:

- `handleAgentRegister` (`RelayServer.ts:760`) already identifies a returning agent by `msg.agentId ?? msg.agentName` plus platform, and already looks up that identity's previous sockets to evict them. The same lookup can instead **re-point the existing sessions at the new socket** when the device set still matches.
[...]

Issue #345, opened 2026-06-28
exports.types points at unpublished src/ across all packages


jo-duchan:
## Background

CodeRabbit flagged (PR #344) that `@tapflowio/audiotap-helper`'s `exports["."].types` resolves to `./src/index.ts`, but `src/` isn't in the published tarball (`files` ships only `dist`/`bin`). So an npm consumer can't reach the type entrypoint (node16 resolution prefers `exports.types` over the top-level `types`).

## Scope — this is monorepo-wide, not audio-specific

**All five library packages share the exact same pattern**, so fixing only the audio packages would create an inconsistency:

| package | exports.types | files |
|---|---|---|
| agent-core | `./src/index.ts` | dist |
| android-agent | `./src/index.ts` | dist, bin, proto |
| audiotap-helper | `./src/index.ts` | dist, bin |
| ios-agent | `./src/index.ts` | dist, bin |
| relay | `./src/index.ts` | dist, public |

This is the monorepo's **source-first** convention: pointing at `src/` lets workspace typecheck resolve types without a dist build first. The trade-off is that the published package's type entrypoint is unreachable.

## Why it hasn't surfaced

Inside the workspace, consumers (cli, etc.) resolve via the `source`/`tsx` conditions against `src/`, so types work. The breakage only appears for an npm-installed consumer.

## Options (pick one, apply to all five)

1. `exports.types` → `./dist/index.d.ts` (matches the top-level `types`), keep `source`/`tsx` → `src` for workspace dev. Needs verification that workspace typecheck still resolves without forcing a dist build.
2. `publishConfig.exports` override so the published manifest points `types` at `dist`.
3. Add `src` to `files` so `src/index.ts` ships (exposes source).

Option 1 or 2 is preferred. Deferred out of PR #344 (audio host-mute) because it touches every package and the build/typecheck ordering, not just audio.

PR #443, opened 2026-07-30
feat(protocol): one wire contract, and type every send against it

### LINKED ISSUE #419: dashboard: narrow relay send() from object to the RelayMessage union
## Problem

Every WebSocket message the dashboard sends goes through a `send` typed as `(msg: object) => void`:

```ts
// packages/dashboard/hooks/useRelay.ts
const send = useCallback((msg: object) => {
  if (ws.current?.readyState === WebSocket.OPEN) ws.current.send(JSON.stringify(msg))
}, [])
```

The same shape is threaded into `DeviceViewer`, both viewers, and `useClipboardBridge`. `object` accepts anything, so nothing checks that an outbound message matches the protocol: a typo in a `type` string, a missing `sessionId`, or a payload with the wrong field name all compile and fail silently at runtime — the relay simply drops what it does not recognise.

`RelayMessage` in `packages/dashboard/lib/types.ts` already describes the protocol, but it is currently only a readable record. Nothing enforces it on the sending side.

## Why now

Raised while reviewing the clipboard bridge (#418). CodeRabbit flagged that the union was missing the two browser→agent clipboard messages, on the grounds that it would give "full type coverage of `send()` calls". The messages were added for consistency, but the stated benefit does not exist yet — `send` takes `object`, so widening the union changes nothing about type safety.

Worth fixing properly rather than leaving the union looking like it does more than it does.

## Suggested direction

Narrow `send` to the union:

```ts
send: (msg: RelayMessage) => void
```

Expect this to surface real mismatches. Some outbound messages are constructed inline and may not match their union member exactly — `device:boot` carries `resetMode`, `acceptH264` and `secureContext`, and the union entry may be behind. Each mismatch is either a union that needs updating or a caller that is genuinely sending the wrong shape; both are worth knowing.

Points to decide while doing it:

- Whether `RelayMessage` should be split into inbound and outbound unions. A viewer can only *send* browser→agent messages, and typing `send` with a union that also contains `clipboard:data` would let a caller send a message only an agent may send (the relay closes such a socket with 1008).
- Whether the relay's `MessageType` and the dashboard's `RelayMessage` should share a source. They are maintained separately today and can drift; the dashboard has no `@tapflowio/*` dependency, so sharing means either adding one or generating.

## Scope note

Not urgent and not a defect on its own — no known bug traces to it. It is a missing guardrail, and the clipboard work showed the cost of missing guardrails: several defects in that feature were only found because someone deliberately broke the code to see whether a test noticed.

jo-duchan:
Closes #419

## Why

Nothing checked the WebSocket messages. The relay built them as inline literals handed to `JSON.stringify`, the dashboard's `send()` took `object`, mcp-server's took `Record<string, unknown>`. Each package kept its own definition of the protocol, and those definitions were **descriptions, not contracts** — so they drifted, and nothing said a word.

Four drifts were found. Two were known going in; two more surfaced while converting the call sites.

| Message | What the definition said | What the wire carries |
|---|---|---|
| `stream:request-idr` | absent from the relay's `MessageType` | sent by the relay from two places |
| `input:key` | `payload: { key: string }` | `{ code, modifiers }`, and `modifiers` is a **HID bitmap** — the field name *and* the type were wrong |
| `app:launch` | no `buildId` | carries `buildId` |
| `input:touch:end`, `app:clear-state` | no payload | mcp-server sends one on both |

The `input:key` one is the instructive case. The authority turned out to be the **consumer**: `IOSAgent.ts` reads `{ code: string; modifiers?: number }` and passes the number to `touchHelper.sendKey(usage, modifiers ?? 0)`. A union that nothing enforces will eventually describe something that does not exist.

## What changed

**`@tapflowio/protocol`** — types only, no runtime code. `dist/index.js` is `export {}`, consumers use `import type`, so nothing reaches the browser bundle. Wired through the root solution tsconfig, relay's `references` (relay is composite — without this its `tsc -b` fails and takes `build`, `typecheck` and `release` with it), the build chain, and the changesets `fixed` group.

**relay** — all 25 originating sends go through `sendTo(socket, msg: RelayOutbound)`. The `readyState` check folds into the helper; it was repeated at most call sites and missing at two. Forwarded messages keep `send(JSON.stringify(msg))` — nothing new is created, so there is nothing new to check.

**dashboard / mcp-server** — `send` and the four `send` props and `sendRef` take `BrowserToRelay`. The dashboard union drops its outbound entries: they were kept "as the readable record of the protocol", which is exactly how `input:key` went wrong.

**`Session.chromeData`** is `ChromePayload` instead of `unknown`. The relay still only stores and forwards it, but the contract states the shape because the viewer parses it — and a new platform now extends the union rather than relay code, which is what the OCP rule asks for.

**No message changed shape on the wire.** Types only.

## Guarding the guard

`packages/protocol/src/typeAssertions.ts` holds lines that **must not compile**, plus valid ones that must. If a union is later widened — a payload back to `unknown`, an index signature added — the `@ts-expect-error` becomes unused, which is itself a compile error. Verified by loosening `modifiers` to `unknown` and watching it fail.

The fixture had a bug of its own first: `@ts-expect-error` applies to the next **line**, so a multi-line literal whose error lands three lines down leaves the directive unused. Now one line each, with that reason recorded.

## Verification

- typecheck / lint / build / test green across all packages (relay 490, dashboard 236, ios-agent 275, android-agent 207, cli 246, mcp-server 39)
- Guard mutation-tested in both directions: a mistyped `type`, a missing field and an extra field are all compile errors (TypeScript even suggests the correct spelling); loosening a payload makes the assertion fixture fail
- `dist/index.js` confirmed to be `export {}` — `main`, `exports.import` and `exports.types` all resolve to something that exists, which is the `#345` trap for a types-only package

## Review

Two rounds, four lenses total, recorded in `.work/reviews/feat__protocol-type-guard.md`.

The **design review found a blocker before any code was written**: the plan wired the build chain but not TS project references, which would have broken `build`, `typecheck` and `release` together. It also caught that `emitDeclarationOnly` would leave `main` pointing at a file that does not exist, that an `enum` would put JavaScript in the browser bundle, and that `DeviceInfo` named three different shapes across the two packages.

The **pre-PR review found no blockers or majors**. Its most useful output was a negative: it established that `sendTo`'s new `readyState` guard cannot drop anything, because `ws.on('message')` and every `route()` case are fully synchronous, so no task can interleave between receiving and sending. Three minors were fixed in `4eec9cc`.

## Follow-on

#426 stage 1 (`session:terminated`) was the work that prompted this. It now lands on top of a checked contract instead of adding a fifth unchecked message.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of device controls, clipboard actions, app operations, and streaming requests through more consistent message handling.
  * Fixed protocol inconsistencies affecting key input, input completion, state clearing, and stream ID requests.
  * Key input now works when modifier information is omitted (defaults appropriately).
  * Prevented outbound messages from being sent when the WebSocket connection is closed.

* **Refactor**
  * Introduced a shared WebSocket wire contract to keep browser, relay, and agent communication aligned, with centralized, typed message sending.


jo-duchan:
Four of the six were valid and are fixed in `7d998e6`. Two need a correction.

### Fixed

**`playground/Dockerfile` missing the protocol `COPY`** — real, and the more dangerous of the two majors. That Dockerfile copies packages individually rather than the whole directory, so the build step I added had nothing to build. CI did not catch it because only the root image is built there. Both the `package.json` and the source are copied now.

**`typeAssertions.ts` emitting into `dist`** — correct, and worse than a style issue: the file declares `export const` values, so a package whose own AGENTS.md says "no runtime values of any kind" was publishing some. It is excluded from the build tsconfig now and checked by a dedicated `tsconfig.assertions.json`, so it is still compiled — an unused `@ts-expect-error` is only an error if something checks it — but not shipped. Re-ran the mutation (loosen a payload to `unknown`) to confirm the assertions still fail.

**`request()` making the write payload optional** — correct. The argument tuple is now derived from the request type, so `clipboard:write` requires its payload and `clipboard:read` may omit one. `request('clipboard:write')` is now `Expected 2 arguments, but got 1`.

**AGENTS.md contradiction (validators here / no runtime values)** — fair reading. It now states the main entry is runtime-free and that a validator requires a second entry point, which is a scope change rather than an addition. Tracked in #444.

### Corrections

**`playground/pre-release` build order** — changed it to match the documented order, but the stated reason does not hold: `relay`'s build is `tsc -b` plus a migrations copy (`packages/relay/package.json:48`) and never reads `public/`. The dashboard writes there and the relay reads it at runtime, after `relay start`. So this was a consistency fix, not a correctness one.

**android-agent as a protocol consumer** — it is not one. `grep -c "@tapflowio/protocol" packages/android-agent/src/AndroidAgent.ts` returns 0. What this PR changed there is a local cast (`modifiers: number` → `modifiers?: number`) so the agent reads `input:key` the way the contract declares it and the way iOS already did. The consumer list of three (relay, dashboard, mcp-server) is accurate.

### Not fixed here

**Correlated errors on the install/launch failure paths** — the finding is right: those errors omit `sessionId`, and an offline agent produces no reply at all, so an MCP client waits out its timeout. But that is a behaviour change in the relay's error handling, not a typing change, and it predates this PR. It belongs with #440, which already covers the install path's other failure mode; I have added it there rather than widening a types-only PR.

Issue #445, opened 2026-07-31
relay: app install/launch failures are uncorrelated or silent — MCP callers wait out the timeout


jo-duchan:
Found by CodeRabbit while reviewing #443. Separate from #440, which covers the same two commands from the **agent** side (device targeting and a boot race); this is the **relay's** error reporting.

## Three paths, none of which reach an MCP caller

`handleBrowserAppInstall` / `handleBrowserAppLaunch` in `packages/relay/src/RelayServer.ts`:

```ts
if (!build) {
  this.sendTo(ws, { type: 'app:install-error', message: 'Build not found' })   // no sessionId
  return
}
if (session.agentSocket.readyState === WebSocket.OPEN) {
  this.sendTo(session.agentSocket, { type: 'app:install', sessionId: msg.sessionId!, ... })
}
// agent offline → nothing is sent to anyone
```

1. **`Session not found`** → `{ type: 'error', message }`, no `sessionId`
2. **`Build not found` / `Bundle ID not available for this build`** → `app:install-error` / `app:launch-error`, no `sessionId`
3. **Agent socket not open** → the `if` has no `else`. **Nothing is sent at all.**

`mcp-server` waits for `app:*-(done|error)` whose `sessionId` matches its own (`packages/mcp-server/src/client.ts:294-317`). All three cases therefore look identical from the client: silence until the deadline expires. The caller is told "timed out" when the truth was "that build has no bundle ID" or "the agent is offline".

There is a comment directly above `handleBrowserAppLaunch` reading *"Relay looks up bundle_id from DB; includes sessionId for response routing"* — which is true of the success path and false of every error path below it.

## Why it stayed hidden

A dashboard viewer holds one session per socket, so an uncorrelated error still lands somewhere sensible and the operator sees a message. MCP is the case that needs correlation, and its failure mode is a wait rather than an error — the least visible kind.

## Fix direction

- Carry the request's `sessionId` on every app-specific error.
- Add the missing `else`: answer immediately when the agent socket is not open, rather than letting the caller wait. `clipboard:read` already does exactly this (`agent offline` reply) — that is the pattern to copy.
- Decide whether `Session not found` should stay a generic `error` or become the app-specific variant. The generic one cannot be correlated by construction.

## Note on the contract

[...]

Issue #444, opened 2026-07-30
relay: inbound messages are never validated — the contract only guards the sending side


jo-duchan:
Follow-on from #419, which typed every message the relay, dashboard and mcp-server **send**. Nothing checks what they **receive**.

The two solve different problems. The outbound guard catches *our own* mistakes — a typo, a missing field, a shape that drifted from the definition — and compile-time is the right place for that. Inbound validation catches *the other side's* malice or bugs, and only a runtime check can do it.

## Where the gap is

| Direction | Trust | Today |
|---|---|---|
| relay ← **browser** | **Attacker-controllable.** A viewer can send arbitrary frames from devtools | `AGENT_MSG_TYPES` closes the socket (1008) if a browser sends an agent-only *type*. The **payload is not checked at all** |
| relay ← agent | PAT with `agent` scope, inside the trust boundary | A buggy agent can still send a wrong shape. `msg.payload as ChromePayload` takes its word |
| dashboard ← relay | Relay is trusted | A wrong shape becomes a runtime error in the viewer |

So the type gate exists and the payload gate does not. Concretely, `RelayServer` reaches for `msg.sessionId!` and `msg.payload as X` throughout — every one of those is an assumption about a message that arrived over the network.

## One part of this is actionable today, and needs no validator

Found while landing #503 (L3 of the wire-contract program). `sessionId` is **absent**, not merely unvalidated, on seven relay replies — and the fix does not depend on the validator design below.

`route()` (`RelayServer.ts:551`) assigns a socket role and blocks a browser from sending agent-only types. It does **not** require a `sessionId`. So a client that sends a bare

```json
{"type":"input:touch:end"}
```

reaches `this.sessions.get(msg.sessionId!)` -> `Map.get(undefined)` -> miss. `input:touch:end` is in `TERMINAL_INPUT_TYPES`, so the relay answers `'Session not found'` through `sessionId: msg.sessionId!` — and `JSON.stringify` **drops a key whose value is `undefined`**. The message that leaves the socket has no `sessionId` at all:

```json
{"type":"input:error","message":"Session not found","reason":"channel-unavailable"}
```

[...]

PR #449, opened 2026-08-01
fix(relay,dashboard): make app install/launch failures reach the caller that asked

jo-duchan:
Refs #445. First of three PRs — see "Sequencing" below.

Three paths through `handleBrowserAppInstall` / `handleBrowserAppLaunch` ended without a usable answer:

1. unknown session → generic `error`, no `sessionId`
2. missing build / bundle id → app-specific error, also no `sessionId`
3. agent socket not open → **nothing sent at all**, because the `if` had no `else`

A dashboard viewer holds one session per socket, so an unattributed error still lands somewhere sensible and a human sees it. An MCP caller waits for the reply carrying its own `sessionId`, so all three looked identical from there: silence until the deadline. The caller was told "timed out" when the truth was "that build has no bundle ID".

## What changed

- **relay** — every exit from both handlers carries the request's `sessionId`, including `Session not found`. A generic `error` cannot be correlated by construction, so it can never end a caller's wait. An unreachable agent is answered now rather than left to time out, the same shape `open-url` and `clipboard:read` already use.
- **relay** — `device:boot` too. A boot the agent never receives left the viewer on "Waiting for first frame…" with nothing said. It also distinguishes a missing session from a live session with a dead socket: `bootDevice` is the first call an MCP caller makes, so reporting a stale id as a dead Mac sends the reader after the wrong problem. `device:shutdown` stays fire-and-forget — nothing waits on it, and a reply type for it would grow the contract for a message no one reads.
- **protocol** — `sessionId` is required on `app:install-error` / `app:launch-error`, and `device:boot-error` joins `RelayToBrowser`. `sendTo` is typed against that union, so an omission at those call sites is a compile error; verified by removing one and watching tsc fail.
- **dashboard** — the viewer ignores messages addressed to another session. Its union is separate from protocol, so adding the field there alone would have compiled clean and changed nothing. Also drops a `msg as unknown as { message: string }` cast that was hiding a field the union already declared.

## Sequencing — why this is PR1

Two parallel design reviews ran **before any code**, and both independently found that the boot-race half of #440 lives in the relay, not the agent: `RelayServer.ts:863` replays `device:ready` off a memory flag set from the agent's register-time `simctl list`, which nothing ever re-derives from the device.

That reordered the work. The plan had this last; it has to be first, because **the failure the race work needs to observe is the one this PR stops swallowing**.

| PR | Scope |
|---|---|
| **1 (this)** | relay error correlation + dashboard consumer + `device:boot` silence |
| 2 | `booted` alias → session udid (ios-agent) |
| 3 | make `device:ready` mean the device is actually up |

## Review

Three independent passes, none of which saw the authoring conversation. Record: `.work/reviews/fix__app-error-correlation.md`.

The pre-PR pass found the dashboard filter is much wider than its comment implied — the relay forwards agent messages verbatim and both agents stamp `sessionId` on nearly all of them, while the local union declares it on only eight. The gate is a runtime `in` check, so it is live on all of them. It is safe because the relay only routes a session-scoped message to that session's own socket, an invariant the types do not express. The comment says that now, and there are tests for the message whose loss would be worst: a dropped `session:terminated` strands the tab, which is the whole of #426.

Four mutations, all caught. The one worth naming: removing the `device:boot` `else` makes its test take **1011ms** — it waits out the timeout, which is precisely the defect.

## Note on scope

`agent offline` turns out to be a narrower branch than #445 describes. Losing an agent drops every session it owned, so that path surfaces as `Session not found`. The real silence was path 1 — a generic `error` no caller could attribute.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error reporting for app installation, app launch, and device boot failures.
  * Errors are now associated with the correct session, preventing unrelated device views from updating.
  * Clear feedback is provided when sessions, builds, bundles, or agents are unavailable.
  * Device boot failures are reported promptly, while shutdown requests remain silent.

* **Tests**
  * Added coverage for session isolation, error handling, device boot behavior, and successful app installation forwarding.


jo-duchan:
> Validate that `msg.buildId` is an integer before calling the database query in both handlers.

Measured rather than assumed — and it turns out the property this PR is about already holds here.

`better-sqlite3` treats both a missing and a non-numeric parameter as "no row" rather than throwing, so the query returns `undefined` and the existing `fail('Build not found')` runs:

```
no buildId  → {"type":"app:install-error","sessionId":"a5c188cf…","message":"Build not found"}
non-numeric → {"type":"app:install-error","sessionId":"d6d9fc0e…","message":"Build not found"}
```

No exception kills the handler, and nothing goes silent. Adding a guard would change the message, not the presence of one.

The message *is* imprecise — "Build not found" when the truth is "you sent no buildId" — but that precision belongs to inbound validation, which is #444 and deliberately outside this PR. Fixing it here would mean hand-validating one field in two handlers while every other inbound field stays unchecked.

Kept the measurement as a test (`answers %s without going silent`) so the behaviour is pinned rather than relied on.

jo-duchan:
Correction to my previous comment — CodeRabbit is right and I was not thorough enough.

I measured `undefined` and `'abc'`, saw better-sqlite3 bind both harmlessly, and generalised. The case it actually named is the one I skipped: **an object or an array makes the driver throw**, and that exception landed in the message loop's catch alongside genuine parse failures.

```
PROBE {}      → NOTHING (silence)
PROBE []      → NOTHING (silence)
PROBE {"a":1} → NOTHING (silence)
```

That is the exact defect this branch exists to remove, reached through the type system's blind spot — `JSON.parse` does not honour `RelayMessage`, so `buildId` is whatever the sender put there.

Fixed in 148844e:

- Both handlers check `Number.isInteger(msg.buildId)` before the query and fail through the same correlated path as every other exit.
- The message loop no longer catches a parse failure and a routing failure in the same block. A throw inside `route()` is a bug rather than a bad message, and it is logged now instead of vanishing — the caller is stuck either way, so the operator should at least be able to see why.

Four new tests cover it. Removing either guard makes all four hang for their full budget, which is what the silence looks like from outside.

General inbound validation stays #444; this is the one field on the path this PR owns.

Unrelated but found while verifying: the 1s reply budget in these tests went flaky under a full-monorepo run, where the relay competes with nine other packages' workers. Raised to 3s — replies are emitted in the same tick, so a passing run never touches it.

jo-duchan:
Right again, and this one was worse than the last.

Splitting the parse and route catches — the change that stops a handler throw from vanishing — introduced a crash path. `JSON.parse` returns bare `null`, numbers, strings and booleans without throwing, so those reach `route`, which reads `.type` off them. The routing catch then did the same in its log line, and that TypeError had nowhere left to go:

```
TypeError: Cannot read properties of null (reading 'type')
```

Reproduced by sending a bare `null` frame. The old single catch swallowed it, which is why it never showed before — so the previous commit traded a silence for a crash, which is not the trade this branch is making.

Fixed in 81b8d84: non-objects are rejected before routing, which is also what lets the catch name the message safely. Four tests send a bare scalar and then assert the relay still answers a normal request; removing the guard makes the `null` case raise an unhandled error and hang for its full budget.

Two rounds, two real defects in code I wrote — both in the same blind spot: I reasoned about the inputs I expected the change to see, not the ones the type system was never enforcing. Thanks.

PR #450, opened 2026-08-01
fix(ios-agent): target the session's simulator instead of simctl's booted alias

jo-duchan:
Refs #440. Second of three PRs; PR1 was #449.

Every app command in `SimctlWrapper` passed `booted` — simctl's alias for "whichever device is running" — instead of the udid of the device the session is about: `install`, `launch`, `uninstall`, `terminate`, `get_app_container`, `io screenshot`.

With one simulator up that is accidentally correct. With two it lands wherever simctl decides, and the wrong device accepts it without complaint. What makes the defect visible today is the loud case — `No devices are booted` — which only happens when nothing is running at all. The quiet case leaves no trace.

`AndroidAgent` already passes an explicit serial; this brings iOS in line.

## What changed

- **The udid is a required leading parameter with no default.** That is the whole guard. With a default, every call site keeps compiling and every test stays green while the alias comes back — `ScreenCaptureStreamer` already carried `udid: string = 'booted'` as a live demonstration. Removed.
- Session call sites pass `DeviceState.deviceId`; `MjpegStreamer` takes the device through its constructor.
- The three `DeviceAgent` entry points have no device parameter (the interface is shared with Android and predates multi-session agents). They resolve the one **booted** session and throw when there is none — or when there is more than one, because this interface has no way to say which is meant.

## Two things the compiler could not do

**It did not catch every call site.** Making the parameter required surfaced nine, but not `screenshot:request`: the old call was `screenshot(format ?? 'png')`, and adding a leading `udid: string` left it type-correct — the format string simply became the device id, so it would have quietly addressed a device named `"png"`. The tests assert arguments rather than trusting the signature.

**Filtering matters more than resolving.** The first version of the `DeviceAgent` lookup took `deviceStates.values().next()`, which reads as "the one live session" and is not: the relay opens a session per *registered* simulator, and this Mac reports 73. It would have addressed whichever simctl listed first, almost always shut down — worse than the alias, which at least found the running device. `AndroidAgent` looks like it does the same thing but gets a liveness check for free, since `getSerial` only returns for a launched device. Caught in review; it filters on `booted` now.

## Review

Record: `.work/reviews/fix__simctl-device-targeting.md`. One independent pass, lens "which path still reaches the wrong device". It found the regression above, and that the tests could not have caught it — `mockSimctl` returns a single device, which makes "the session's device", "the first registered device" and a hardcoded constant indistinguishable. There is a two-device fixture now.

Three mutations, all caught. One of them was not, at first: the regression test asserted on `simctl.screenshot`, which the MJPEG stream calls on a timer with the same udid, so it passed before the line under test ran. It asserts on `installApp` now.

## Still to verify by hand

The quiet case cannot be proven with mocks. Boot two simulators, open a session on one, and confirm the other is untouched (`xcrun simctl get_app_container`). Noted in the review record.

## Summary by CodeRabbit

* **New Features**
  * iOS simulator actions now target the specific device associated with each session.
  * App installation, launching, data clearing, screenshots, and video streaming support reliable multi-device workflows.
* **Bug Fixes**
  * Prevented commands from being sent to the wrong simulator when multiple devices are available.
  * Requests now fail clearly when no single booted simulator can be identified.
  * Full-device erase operations better preserve previously booted devices if an error occurs.
* **Tests**
  * Expanded coverage for multi-device targeting and session-specific simulator operations.


jo-duchan:
Manual verification done — the quiet case is covered now.

First, the premise. With two simulators booted at once, `xcrun simctl get_app_container booted …` **does not error**; it silently picks one. That is why this defect has no symptom of its own.

Then the test: both devices booted, app uninstalled from both, session opened on one.

| | |
|---|---|
| iPhone 15 Pro (the session's device) | installed ✓ |
| iPhone 15 (not the target) | **untouched** ✓ |
| Both devices | still Booted throughout — the condition under which the alias could have picked either |
| udid in the agent log | `0535B2B2…`, one value |

This is the case mocks cannot reach: a fake runner does not resolve `booted`, so the second device is dead state in a unit test and the only observable difference is the literal argument. Recorded in `.work/reviews/fix__simctl-device-targeting.md`.

jo-duchan:
Fixed in 026ba9a — and the nitpick turned out to be load-bearing.

The reorder itself is straightforward: dropping `udid`'s `= 'booted'` default left it required between two defaulted parameters, so `fps = 30` was unreachable. It is device-first now, matching `SimctlWrapper`.

What it surfaced is the part worth reading. Updating the constructor's callers showed that `stream()` still used `deviceStates.values().next()` — and so did `queryUITree` and `openUrl`. That is the same "first registered simulator, usually shut down" hazard the earlier review found in `installApp` / `launchApp` / `screenshot`. Fixing only the three that were named would have left it in four more places on the same interface.

They all go through `soleDeviceState()` now. The input methods (`touchStart` / `touchMove` / `touchEnd`) use a softer variant that returns undefined instead of throwing: a tap with no unambiguous device is better dropped than turned into an exception, and `touchHelper` being null already made them no-ops in the common case.

`sessionId` keeps its first-entry lookup deliberately — it answers "which session does this agent report as its own", which is not a question about boot state.

PR #451, opened 2026-08-02
fix(relay): replay device:ready for a streaming session, not a running device

### LINKED ISSUE #440: ios-agent: app install intermittently fails with "No devices are booted" — simctl calls target the `booted` alias, not the session's udid

jo-duchan:
Closes #440. Last of three PRs; #449 and #450 are merged.

`Session.deviceStatus` was answering two questions with one field.

| Question | Read by | Snapshot answers it |
|---|---|---|
| Is this device up? | device list, REST guards | correctly |
| Did this session announce a stream? | the `device:ready` replay | **no** |

The relay opens a session for **every** device an agent reports, and `deviceStatus` is initialised from the agent's `simctl list` snapshot at registration. So a simulator somebody left running had a session marked `booted` from the moment it registered, and a browser joining it was told the device was ready before the agent had done anything with it.

Observed while writing tests for #450: `session:joined` → **`device:ready`** → (boot sent) → `device:booting` → `device:ready`. The first one meant nothing — no stream, no work done.

## What changed

`readySent` answers the second question. Set when the agent sends `device:ready`; cleared on `device:shutdown-done`, on `device:booting`, and when the stream socket closes.

That last one matters because the agent does not always get to report the end. `handleDeviceShutdown` cancels the stream reader and closes the stream socket *before* `await simctl.shutdown(deviceId)`, and a throw there is only logged — no `device:shutdown-done` is sent. Without the socket signal the relay keeps claiming a stream, and the dashboard, on a replayed `device:ready`, fires an install into nothing.

`deviceStatus` is untouched. There is a test that the device list still reports it: moving the existing readers onto the new flag would have made every device read as shut down.

## Two documents this made false

`packages/ios-agent/AGENTS.md` and a test comment both teach that the relay replays `device:ready` for any already-booted session, and that this is what made the codec-negotiation test flake at ~2/10 runs. That mechanism is what this PR removes. The `vi.waitFor` rule they exist to justify still holds — `device:ready` is sent as the stream is handed off, before the helpers a test is about to read are observable — so they now say that, with the old cause noted as fixed. Caught in review; the PR would otherwise have left contributors defending against a hazard that no longer exists.

## Review

Record: `.work/reviews/fix__device-ready-liveness.md`. One independent pass, lens "what stopped working, or started lying".

Six mutations, all caught. The one worth naming is removing the replay entirely: with only the "does not arrive" assertions it would have been green, and that is the regression where a reconnecting tab never gets a picture back. The reconnect case is what makes the negative ones mean something.

Also from the review: the IDR request rides the replay branch and had lost its independent assertion, and three negative assertions were waiting out 800 ms timers. The replay is written synchronously with `session:joined`, so a round-trip barrier proves the same thing — the file went from 5.5 s to 363 ms.

## Follow-ups, deliberately not here

- The agent's own `device:ready` timing. `simctl boot` returns before the device reaches Booted, and the agent sends ready right after the stream handoff. Whether that is early was impossible to measure while this replay was also firing; it is measurable now.
- Both agents log a failed `simctl shutdown` / `adb emu kill` and tell nobody. This PR stops the relay from being misled by it, but the silence itself remains.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved device readiness notifications during session reconnects.
  - Readiness is now replayed only when an active stream is available, preventing stale notifications for devices that are merely registered or previously connected.
  - Readiness is cleared when a device reboots, shuts down, or its stream disconnects.
  - Device availability status remains accurate independently of stream readiness.

- **Tests**
  - Added coverage for reconnects, shutdowns, reboots, stream disconnections, and related IDR requests.

### DIFF
--- a/.changeset/device-ready-liveness.md
+++ b/.changeset/device-ready-liveness.md
@@ -8,4 +8,4 @@ The relay replays `device:ready` when a browser joins, so a tab that lost its so
 
 The replay now keys off whether this session announced a stream and has not since taken it back, tracked separately from the device's own state. `deviceStatus` is unchanged and still answers "is this device up" for the device list and the REST guards — the two questions were sharing one field.
 
-Also released when a reboot starts: `device:booting` already cleared the cached chrome for the same reason, and a browser joining mid-boot should not be promised a stream that is being torn down.
+Also released when a reboot starts and when the stream socket goes away. `device:booting` already cleared the cached chrome for the same reason, and a browser joining mid-boot should not be promised a stream that is being torn down. The stream socket matters because the agent does not always get to report the end: `handleDeviceShutdown` tears the streamer down before running `simctl shutdown`, and if that throws, no `device:shutdown-done` is ever sent.
--- a/.changeset/device-ready-liveness.md
+++ b/.changeset/device-ready-liveness.md
@@ -1,0 +1,11 @@
+---
+"@tapflowio/relay": patch
+---
+
+Stop telling a viewer a device is ready when nothing is streaming.
+
+The relay replays `device:ready` when a browser joins, so a tab that lost its socket mid-session gets a picture back without waiting for another boot. The condition for that replay was `deviceStatus === 'booted'` — and `deviceStatus` starts life as the agent's `simctl list` snapshot at registration. Since the relay opens a session for every device an agent reports, a simulator somebody left running had a session marked booted before the agent had done anything with it. Joining that session produced a `device:ready` with no stream behind it.
+
+The replay now keys off whether this session announced a stream and has not since taken it back, tracked separately from the device's own state. `deviceStatus` is unchanged and still answers "is this device up" for the device list and the REST guards — the two questions were sharing one field.
+
+Also released when a reboot starts: `device:booting` already cleared the cached chrome for the same reason, and a browser joining mid-boot should not be promised a stream that is being torn down.
--- a/.changeset/device-ready-liveness.md
+++ b/.changeset/device-ready-liveness.md
@@ -8,4 +8,6 @@ The relay replays `device:ready` when a browser joins, so a tab that lost its so
 
 The replay now keys off whether this session announced a stream and has not since taken it back, tracked separately from the device's own state. `deviceStatus` is unchanged and still answers "is this device up" for the device list and the REST guards — the two questions were sharing one field.
 
-Also released when a reboot starts and when the stream socket goes away. `device:booting` already cleared the cached chrome for the same reason, and a browser joining mid-boot should not be promised a stream that is being torn down. The stream socket matters because the agent does not always get to report the end: `handleDeviceShutdown` tears the streamer down before running `simctl shutdown`, and if that throws, no `device:shutdown-done` is ever sent.
+The flag is cleared on three events: `device:shutdown-done`, `device:booting`, and the stream socket closing.
+
+`device:booting` already cleared the cached chrome for the same reason — a browser joining mid-boot should not be promised a stream that is being torn down. The stream socket matters because the agent does not always get to report the end: `handleDeviceShutdown` tears the streamer down before running `simctl shutdown`, and if that throws, no `device:shutdown-done` is ever sent.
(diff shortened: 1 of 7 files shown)


jo-duchan:
Both fixed — and the second one was a race I introduced in this PR.

Replacing the 800 ms waits with a round-trip barrier made the negative assertions deterministic, but I put the barrier on the browser socket only. The lifecycle messages these tests depend on (`device:ready`, `device:shutdown-done`, `device:booting`) go out on the *agent* socket, and two connections have no ordering between them — nothing guaranteed the relay had processed the agent's message before the browser joined. The tests passed on timing, not on anything they established.

There is a `barrier(ws)` helper now, applied on the agent socket after each lifecycle sequence and on the browser socket after the join. WebSocket preserves order within a connection, so the reply proves everything sent before it has been handled. The 50 ms timer in the IDR test is gone the same way — both files are sleep-free, and the whole replay file runs in ~360 ms.

The helper registers its listener before sending, which is the mistake this branch had already made twice.

The changeset now names the three events that release the flag instead of saying "also released".

PR #453, opened 2026-08-02
test: make socket helpers order-proof, in one place instead of nine

### LINKED ISSUE #452: test: waitForType silently misses messages that already arrived — make the API order-proof, not the caller
## The failure mode

`waitForType(ws, type)` attaches its listener when it is called, so a reply that lands before the call is lost and the test times out with no indication of why. The correct order is to build the promise *before* triggering:

```ts
const ready = waitForType(browser, 'device:ready')   // must come first
browser.send(JSON.stringify({ type: 'device:boot', ... }))
await ready
```

Nothing enforces that, and getting it wrong costs a 5s timeout pointing at the assertion rather than the cause.

I hit this three times in one branch (#451), twice with a correct example a few lines above in the same file. It is not a knowledge problem — the shape of the helper allows a mistake that has no signal.

## Scale

```
9  files define their own copy of waitForType
201 call sites
10 files define their own waitForOpen
175 socket creations
```

The copies are near-identical (one differs only by a type assertion), so a fix in one is a fix in none.

## What to change

**Record from the socket's first moment, and answer from the recording.** Then "did I ask before or after it arrived" stops being a question the caller has to get right.

```ts
const ws = new WebSocket(url)
await waitForOpen(ws)        // starts recording here
ws.send(...)
await waitForType(ws, 'device:ready')   // finds it whether it already came or not
```

Sketch: one message listener per socket, installed by `waitForOpen`. Arriving messages hand off to a matching waiter if there is one, otherwise queue. `waitForType` takes from the queue first, then waits. Matched messages are removed so two waits for the same type behave as they do today.

**The call sites do not need to change.** Every socket already goes through `waitForOpen`, and the signature stays the same — the work is deleting nine copies and importing one.

Binary frames need care: the current copies `JSON.parse` unconditionally, which throws on stream data. Skip on `isBinary`.

## Why not just a comment

A JSDoc note is worth having and will be part of this, but it is the last line of defence rather than the first. The two mistakes in #451 happened next to working examples of the right pattern — reading material was not what was missing. An API that cannot be called wrongly does not depend on anyone being careful.

## Scope

- `packages/relay/src/__tests__/` — six copies, the biggest win
- `packages/ios-agent`, `packages/android-agent` — one copy each, same hazard

Test-only change; no production code.

jo-duchan:
Closes #452.

`waitForType(ws, type)` attached its listener when called, so a reply that landed first was lost and the test died on a timeout pointing at the assertion rather than the ordering. I made that mistake three times in one branch, twice with a correct example a few lines above — reading material was not what was missing.

## What changed

The helpers **record from the moment a socket opens** and answer from the recording, so asking after the fact behaves exactly like asking before:

```ts
const ws = new WebSocket(url)
await waitForOpen(ws)          // recording starts here
ws.send(JSON.stringify({ type: 'device:boot', … }))
await waitForType(ws, 'device:ready')   // fine whether it has landed yet or not
```

**No call site changed.** Every socket already goes through `waitForOpen`, and the signatures are the same — 201 call sites, none touched. The work was deleting nine copies and importing one.

They live in `@tapflowio/test-utils`: `private: true`, no build step. `main` points at TypeScript source because vitest transforms it anyway, and a `dist/` would drag the package into the production build order for nothing.

An earlier attempt kept one copy per package, which would have preserved this issue's own defect at a smaller scale. The workspace package is the structural answer, and `@tapflowio/protocol` was the precedent.

## Two tests were relying on the old imprecision

- `returns open-url:error…` waited for "the next message", and a dead agent also produces a `session:terminated`. Recording surfaced the message that was previously being skipped. It waits by type now.
- `answers a boot the agent will never receive` — see below.

## Review found the interesting one

Record: `.work/reviews/test__order-proof-socket-helpers.md`.

The reviewer's first target — "recording will match a stale message" — was **disproved by instrumentation**: tracing the queue-hit path across three full suites showed exactly one queue hit, in the test that intends it. Everything else still resolves through a waiter, observing what it observed before.

But it found something better. I had removed a `Session not found` assertion, calling it timing-dependent. The reviewer said it was unrelated to this refactor and passed five times running. Half right — at **ten** runs it fails about once. So the assertion was neither safe to keep as it stood nor safe to drop: it is #445's guard that a stale session id is not reported as a dead Mac. The race predates this branch; closing an agent socket starts the relay's teardown and nothing orders it against the browser's next request. There is a wait for the eviction to be observable now, and twenty consecutive runs pass.

Weakening the assertion was the easy path, and it is exactly how a guard gets lost.

It also found that **nothing type-checked `packages/test-utils/src`** — not in the root `tsc -b` references, no `typecheck` script, and every consumer excludes `src/__tests__`. Proved by planting a type error and watching the gate pass. Wired in now; turning the check on immediately surfaced a missing `@types/node`.

Plus: a tenth copy of the order-dependent helper still living in one test file, undocumented queue-vs-broadcast semantics (a message now goes to exactly one waiter), and two dead imports that neither eslint nor tsc can see.

## Summary by CodeRabbit

* **New Features**
  * Added shared WebSocket testing utilities for reliable message handling, ordering, and timeout assertions.
  * Added integration coverage for socket ordering and relay error scenarios, including unknown-session boot responses.

* **Bug Fixes**
  * Improved relay test synchronization and targeting of specific error messages.

* **Documentation**
  * Documented the private test utilities package and usage guidelines.

* **Chores**
  * Unified test helpers across Android, iOS, and relay test suites.
  * Added package type checking to the root validation workflow.

### DIFF
--- a/INDEX.md
+++ b/INDEX.md
@@ -25,6 +25,7 @@ Common rules are in the root [AGENTS.md](./AGENTS.md).
 | cli | [packages/cli/AGENTS.md](./packages/cli/AGENTS.md) | CLI UX rules |
 | mcp-server | [packages/mcp-server/AGENTS.md](./packages/mcp-server/AGENTS.md) | MCP server bridging tapflow to LLM agents |
 | flow-runner | [packages/flow-runner/AGENTS.md](./packages/flow-runner/AGENTS.md) | Deterministic YAML flow engine (automated QA axis) |
+| test-utils | [packages/test-utils/AGENTS.md](./packages/test-utils/AGENTS.md) | Test-only helpers shared across packages (private, never published) |
 
 ## Local Only
 
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "pnpm --filter @tapflowio/protocol build && pnpm --filter @tapflowio/agent-core build && pnpm --filter @tapflowio/dashboard build && pnpm --filter @tapflowio/flow-runner build && pnpm --filter @tapflowio/relay build && pnpm --filter @tapflowio/ios-agent build && pnpm --filter @tapflowio/android-agent build && pnpm --filter tapflow build && pnpm --filter @tapflowio/mcp-server build",
     "test": "vitest run --config scripts/vitest.config.mjs && pnpm -r test --if-present",
     "lint": "eslint scripts && pnpm -r --if-present lint",
-    "typecheck": "tsc -b && pnpm --filter @tapflowio/dashboard --filter @tapflowio/mcp-server typecheck",
+    "typecheck": "tsc -b && pnpm --filter @tapflowio/dashboard --filter @tapflowio/mcp-server --filter @tapflowio/test-utils typecheck",
     "prepare": "lefthook install || true",
     "dev": "node scripts/dev-preflight.mjs relay dashboard && concurrently -k -n relay,dashboard,ios,android -c cyan,magenta,green,yellow \"pnpm dev:relay\" \"pnpm --filter @tapflowio/dashboard dev\" \"pnpm dev:ios\" \"pnpm dev:android\"",
     "dev:pool": "node scripts/dev-preflight.mjs relay && concurrently -k -n relay,ios,mocks -c cyan,green,yellow \"pnpm dev:relay\" \"pnpm dev:ios\" \"pnpm --filter @tapflowio/playground mock-agents\"",
--- a/packages/android-agent/package.json
+++ b/packages/android-agent/package.json
@@ -62,6 +62,7 @@
   },
   "devDependencies": {
     "@tapflowio/relay": "workspace:*",
+    "@tapflowio/test-utils": "workspace:*",
     "@types/node": "^20.19.43",
     "@types/ws": "^8.0.0",
     "typescript": "^5.0.0",
--- a/packages/android-agent/src/__tests__/AndroidAgent.test.ts
+++ b/packages/android-agent/src/__tests__/AndroidAgent.test.ts
@@ -128,6 +128,7 @@ import { isAudioSupported, launchMuteOnlyTap } from '@tapflowio/audiotap-helper'
 import type { ScrcpyControl } from '../scrcpy/ScrcpyControl'
 import type { ScrcpyFrame } from '../scrcpy/ScrcpyVideo'
 import type { AdbRunner } from '../adb'
+import { waitForOpen, waitForType } from '@tapflowio/test-utils'
 
 // Test-only view of a per-device state entry (the real DeviceState is not exported).
 interface TestState {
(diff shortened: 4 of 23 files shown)


jo-duchan:
All six addressed. The first one was the real find.

**`barrier` was eating replies it did not ask for.** It went through `waitForType`, which reads the recording first, so it could return on an `agents:listed` a test had queued earlier — no round-trip performed, nothing proven, and the test's own message consumed on the way past. It registers its waiter before sending and never reads the queue now: arriving messages go to a waiter ahead of the queue, so it gets the reply to its own request.

**The helper tests could not have caught it.** They used `barrier` to establish "has already arrived", which lets a broken barrier prove itself — reverting the fix left all five green. They use an invalid `session:start` as an ordering sentinel now (it is answered with `error`, and order within a connection does the rest), so nothing under test is used to test itself.

Telling the two barriers apart also needed a discriminating assertion, which took a second attempt. Counting messages cannot do it: a barrier that eats the queued reply still leaves its own behind, so the count matches either way. Registering an agent between the two requests makes the replies differ — one lists no sessions, the other lists one.

The rest:

- `waitForOpen` now rejects on `error` instead of hanging until the suite timeout. Test connects to a closed port.
- Dropped the wall-clock assertion. Measuring elapsed milliseconds is a flake waiting for a loaded machine; the sentinel is what makes a 0 ms budget meaningful, and the test says that instead.
- `waitForType` and friends take a type parameter for narrowing. Almost nothing uses it, deliberately: every package excludes `src/__tests__` from `tsc`, so narrowing is decorative today. It is there so bringing tests under `tsc` becomes a config change rather than a rewrite — noted in the package's AGENTS.md.

jo-duchan:
@coderabbitai review

jo-duchan:
Both inline comments handled — thanks for flagging them, I had only worked from the summary and missed these two.

**uiTree** — real, though latent. `setupAgent` read `agent:registered` with a raw `once('message')`, and the recorder queues that frame regardless, so it sat unclaimed. Nothing in that file waits on the agent socket today, so no test observes it — but leaving a stale message behind in the PR whose subject is stale messages is the wrong shape. It goes through `waitForType` now, which consumes it.

**Type narrowing** — done at the five places that read `registeredSessions`. Being plain about what it buys: **nothing at the gate**. Planting `const bogus: number = reply.registeredSessions` in a test file and running `pnpm typecheck` still reports zero errors, because every package excludes `src/__tests__`. So 1854 tests run with their types stripped. The narrowing helps an editor and prepares for that changing; it is not a correctness fix today.

That last part turned out to be worth its own issue — #454. This PR is where it became measurable: `@tapflowio/test-utils` was unchecked until it was wired in, and switching the check on immediately surfaced a missing `@types/node`.

On the re-review: CodeRabbit is incremental and does not re-review commits it has already seen, so the `pass` on the checks reflects the earlier pass rather than a verdict on the fixes. Happy to have another look triggered if that is useful.

PR #456, opened 2026-08-02
feat(dashboard,protocol): let a viewer recover from an agent restart — the receiving half

jo-duchan:
Stage 2 of #426, split in two. This is PR1: the message and the viewer that acts on it. **Nothing sends `session:rebound` yet**, so no behaviour changes on merge.

## Why this half first

The reverse order leaves the tab worse than today. The viewer would drop the unknown message, and `device:boot` is only re-sent from the `session:joined` branch — so there would be no recovery path at all: a frozen frame that looks live until someone refreshes.

## What a rebind is

Restart a device agent today and the tab is told `session:terminated` and sent back to the Mac list. Better than the silence it replaced (#446), but it makes the tester redo their navigation for something that should be invisible.

The alternative is three steps: the relay keeps the session and re-points it at the new agent socket → `session:rebound` → the browser asks for its device back. The relay cannot restart the stream itself, because the codec negotiation and the downscale tier ride in the browser's own `device:boot` payload and exist nowhere the relay can see. `capabilities` rides along because `session:joined` is sent once and a restarted agent may advertise a different set — an upgrade is the usual reason to restart one.

## What the viewer does on receipt

Tears down first, then re-boots:

- Clears what a restart invalidates, including three flags `device:booting` never touches — `launching`, `swKeyboardPending`, `swKeyboardVisible`. Their acknowledgements died with the old agent. This was unreachable before: a dead agent unmounted the viewer, so nothing could outlive it.
- Re-sends `device:boot` carrying `app-only`, never a reset. A restart is not a request to erase the device (#439).
- Skips the reinstall **when the build was already on the device** — the simulator stayed up, and reinstalling would kill the app state the recovery exists to preserve.

That last one is where the pre-PR review did its work. It broke the first cut three ways, all from the same root: a boolean cannot say how many rebinds are in flight, or whether the app was actually there when one started.

| | |
|---|---|
| Rebind lands **mid-install** | Skipping and setting `installed` anyway leaves a Launch button for an app that is not on the device |
| New agent **never answers** | The flag survives; the next socket blip's `device:ready` is absorbed as if it were the rebind's, suppressing installs for the rest of the mount |
| Agent **crash-loops** | The first `device:ready` spends the flag and the second reinstalls — the exact thing the skip prevents |

Now a count plus the install state captured when the rebind starts, reset by `session:joined` and `device:boot-error`.

## Tests

13 cases, and **mutation kills all eleven guards** in the block. Two of the original assertions could not fail and were rewritten: `swKeyboardPending` was never true in the test that claimed to clear it (`keyboard:toggled` sets it *false*; only a click sets it true), and `launching` was covered by reading the code rather than by a test.

The commit gate found a third. The negative assertion `getByRole('button', {name: /launch/i})` matched nothing, because the Launch button is icon-only — so it passed whatever the viewer rendered. Both viewers now name that button, which fixes the accessibility defect and makes the assertion die when the control appears early. That is the only place this diff widens its scope, and that is why.

## Review

`.work/reviews/feat__session-rebind.md` — three design rounds before any code (two packages), one pre-PR round. Two minor findings are carried to PR2 with reasons: an in-progress recording is discarded by the teardown, and `IOSViewer`'s unmount `input:rotate` lands on a booting agent. Both are unreachable here and both already exist on the `device:booting` path; the difference is that a rebind does not actually reboot the device, and what that means depends on how the agent handles a boot for an already-booted device — PR2's territory.

## Next

PR2 is the relay half: detect the restart, re-point the session, send the message.

## Summary by CodeRabbit

* **New Features**
  * Improved device viewer recovery after agent restarts, including automatic reinitialization and preservation of app state where possible.
  * Automatically reinstalls apps when installation was interrupted during recovery.
  * Supports repeated recovery events and updated device capabilities.

* **Bug Fixes**
  * Cleans up stale viewer state and pending actions after reconnects or boot errors.
  * Ignores recovery events from unrelated sessions.

* **Accessibility**
  * Added accessible “Launch app” labels to Android and iOS launch buttons.


jo-duchan:
All three taken. Pushed in 0fbeba0.

**Button count → named assertion.** Correct, and it was mine to catch: I wrote that test before the Launch control had a name, then added the name in the same PR and never went back to upgrade the assertion. Counting passes if Launch vanishes and anything else appears in the same render.

**Hardcode `app-only`.** Also correct, and the reasoning is the part worth keeping: the derived expression agrees with the invariant only because a rebind cannot precede a join on the same mount. The comment above it already asserted the invariant outright, so the code was quietly weaker than its own documentation.

**Dropping the casts — my earlier disposition was wrong.** I dismissed this same suggestion on #453 with "test files are outside `tsc`, so narrowing is decoration." That measurement was taken on `relay`, and it does not hold here: `packages/dashboard/tsconfig.json` is `include: ['src', 'components', 'hooks', 'lib']` with no exclude for `__tests__`. Planting a deliberate type error confirms the dashboard catches it.

So the casts were suppressing a live check, exactly as described. With them gone, a `capabilities: 'not-an-array'` fixture is now `TS2322`; before, it compiled. Only the six `session:chrome` deliveries keep a cast, since `CHROME` is a partial shape.

Both changed guards are mutation-checked: `resetMode: 'app-only'` → `resetMode` kills *does not erase the device on the way back*, and removing `setInstalled(true)` on the rebind path kills *keeps the installed app…* — which now fails on the name rather than on a count.

Issue #313, opened 2026-06-19
Add WebSocket heartbeat (ping/pong) to detect dead agent sockets fast


jo-duchan:
## Problem

The relay has no application-level WebSocket heartbeat (ping/pong + liveness check). When an agent drops uncleanly (Wi-Fi loss, sleep, cable pull), the old socket's `close` event fires only when the OS TCP stack times out — tens of seconds to minutes. Until then the relay still holds the dead socket and its sessions.

This is the slow-decay half of the duplicate "Stale" Mac card bug: even after the register-time dedup fix, a dead socket lingers until TCP timeout if the agent never reconnects.

## Proposal

Add a heartbeat to the relay's WebSocket server:

- Periodic `ws.ping()` to each connection, track `isAlive` via the `pong` handler.
- On a missed pong window, `ws.terminate()` the dead socket (which triggers the existing `close` cleanup in `RelayServer.ts`).
- Tune interval/timeout to balance responsiveness vs overhead; ensure browser and stream sockets are handled too (not just agents).

## Context

- Register-time dedup (closing a same-`(agentName, platform)` agent's stale socket on re-register) is handled separately as the primary fix for the duplicate-card symptom.
- `agentName` is `os.hostname()` — not a unique id; see the dedup fix for the keying caveat.

## References

- `packages/relay/src/RelayServer.ts` — `WebSocketServer` setup, `ws.on('close')` cleanup
- Related: duplicate "Stale" Mac card fix (register-time dedup)

PR #459, opened 2026-08-02
fix(relay): hold a session while its agent restarts — close #426

### LINKED ISSUE #426: Restarting an agent orphans the open dashboard tab — it spins on "Waiting for first frame..." forever

jo-duchan:
Third and last stage of #426. Stage 1 stopped the silence (#446). Stage 2 built the machinery to re-point a session at a restarted agent (#456, #458) — and it almost never ran. This makes it run.

## Why stage 2 did not close the issue

The rebind required the old socket to still be registered when the new one arrived. On a real restart it never is: the close is processed in under 400 ms while a new agent takes about a second to register. Measured twice on an iPhone 15 Pro with an app open — the tab got the same red toast and trip back to the Mac list it got before any of this existed.

Six design review rounds, two pre-PR review channels and 1889 tests all checked *how* the rebind behaves. None checked *whether its trigger occurs*, because the integration tests opened both sockets at once and handed the precondition over for free.

## What this does

A closed agent socket no longer ends its sessions. They are held for 15 seconds (`TAPFLOW_AGENT_GRACE_MS`), the returning agent reclaims them through the existing `rebind()`, and the tab keeps its place. If nobody comes back the window closes and the session ends exactly as before.

The sessions stay where they are rather than moving somewhere of their own. A returning agent is found by walking the sessions and reading the socket each one points at, so a session parked in a holding structure could never be reclaimed — the design review flagged that as the one way to write this that silently does nothing.

**`session:agent-away`** tells an attached viewer what is happening. Without it the tab sits on a frame that stopped updating for the length of the window, which is the complaint #426 opened with. A genuinely dead agent now reads better than before, not worse: the wait is explained, and only the news that it is over arrives later.

While an agent is away, its devices are not offered, joining is answered rather than refused, nothing the previous process reported is replayed, and `device:boot` says `agent offline` instead of `Session not found`.

## Verified on hardware

iPhone 15 Pro / iOS 17.2, an app open on an inner screen.

```
[relay] agent socket lost — holding 73 session(s) for 15000ms
[relay] agent restarted — 73 session(s) kept across the restart
```

`kill -9` and respawn: the stream comes back on its own and the app is still on the same screen. Killing without a respawn shows the waiting line, then ends the session after 15 s — judged better than today's instant bounce, which is what set the window.

## What the reviews changed

Both the design review and the two pre-PR channels found things worth more than the code they touched.

**Before any code**: the draft refused joins during the window. That reintroduces #426 through a different door — the viewer sends `session:start` once per reconnect and ignores a plain error, so one browser blip would strand the tab past any recovery. The draft also left the Mac list showing a dead agent with its last resource reading, and would have drawn two cards for one Mac when the restart came from the upgrade that starts sending a machine id.

**At PR time**: the flagship test could pass for the exact reason stage 2 shipped broken. `closeAndSettle` waits for the *client's* close event; the relay's close handling is a separate event, and the hold was not armed in 20 of 20 runs when it resolved. What ordered them was that opening a second socket takes longer than the server's close handling — luck. At a zero window a rebind still landed 1 run in 30 through the old overlap. It now waits for `session:agent-away`, whose only sender is the hold itself.

Two more permanent-wait paths, a `stop()` that cleared its holds and then re-armed one behind itself, a resource entry that leaked when a held session left by any route but a rebind, and `TAPFLOW_AGENT_GRACE_MS=15s` parsing to a fifteen-*millisecond* window — the typo the documented default invites.

## Tests

Sixteen mutants killed. Three assertions that could not fail were rewritten, including one hidden by the status card's own ordering — the coupling that was also letting a recovery report "Boot failed" while it was going fine.

One mutant survives knowingly and the record says why: the test window is 300 ms, which cannot tell "immediately" from "at expiry" for the identity-change path. At the real 15 s that is the whole difference, and it is covered on the wire rather than by that assertion.

Full suite green against a fresh build: 1914 tests.

## Review

`.work/reviews/feat__agent-reconnect-grace.md` — one design review before code, two channels at PR time, each given the measurements already taken and forbidden from re-running them. Left out of scope with reasons: `session:leave` not checking socket ownership, moving `agentResources` to a `WeakMap`, and the identity fallback in #313.

Closes #426.

## Summary by CodeRabbit

* **New Features**
  * Sessions remain available briefly after an agent disconnects, allowing reconnection without moving viewers.
  * Viewers see an agent-away status while stale device information is hidden.
  * Reconnection restores sessions automatically; expired or mismatched sessions end safely.
  * Temporarily unavailable agents and devices are omitted from listings.
  * Added a configurable grace period, defaulting to 15 seconds.

* **Bug Fixes**
  * Boot requests during agent absence now report that the agent is offline.
  * Prevented stale frames, metadata, and duplicate reconnection notifications.


jo-duchan:
All three taken.

**`Number('')` is 0, not NaN** — and this is the second time this line has been wrong. The pre-PR review flagged three values that silently switch the hold off: `''`, `abc`, and `15s`. My fix caught the last two, missed the first, and the comment I wrote asserted the blank case landed on NaN. It does not: `Number('')` is `0`, which passes `>= 0` and gives a zero-length window — the feature off, with nothing said, from an empty line in `.env`.

Blank is now treated as unset. Since one line has managed to be wrong twice, every value it has ever got wrong is pinned:

| input | window |
|---|---|
| unset, `''`, `'   '`, `abc`, `15s`, `-5` | 15000 (default) |
| `3000` | 3000 |
| `0` | 0 — no hold, deliberately expressible |

Both past versions were reproduced as mutants: the one you found fails 2 of those cases, the earlier unvalidated `parseInt` version fails 5.

**Comment placement** — correct, and the consequence is worse than misattribution: the `session:rebound` block explains its `capabilities` field, and it now sat above a message that has none.

**Korean table** — added, matching the English row.

Full suite green against a fresh build: 1922 tests.

jo-duchan:
Two of the three as asked; the third answered in code instead.

**Unit and `0`** — both into the English and Korean tables. `0` especially: it is not inferable from the name, and it is the only way back to the pre-#426 behaviour.

**"a blank, invalid or negative value uses the default"** — writing that in a reference table documents the failure rather than fixing it. Someone who typed `15s` is looking at their terminal, not at this row. So the fallback now says so:

```
TAPFLOW_AGENT_GRACE_MS="15s" is not a usable number of milliseconds — using 15000
```

Two reasons to prefer that. This parsing has now been wrong twice, and both times the symptom was identical — the hold switched off and nothing mentioned it. Silence *is* the defect here, so covering it with a doc line is the weaker half of the fix. And the repo already warns this way: `lib/config.ts` does it for a malformed webhook entry, an unparseable config file, and a deprecated key.

The tables still carry one clause about the fallback, since a reader who is there deserves to know — but the warning is the part that reaches the person who needs it.

Mutation-checked both directions: dropping the warning fails 3 cases, warning on a value that is fine fails 2.

Full suite green against a fresh build: 1928 tests.

PR #462, opened 2026-08-03
chore: release v0.18.0 — an agent restart no longer costs the tab its place

jo-duchan:
Ten changesets, two of them minor.

| package | | |
|---|---|---|
| `tapflow` · `@tapflowio/protocol` · `agent-core` · `ios-agent` · `android-agent` · `relay` · `flow-runner` · `mcp-server` | 0.17.0 | **0.18.0** |
| `@tapflowio/audiotap-helper` | 0.2.7 | **0.2.8** (dependency-only) |

## The cycle in one line

Most of it is one thing said several ways: the system stopped being quiet when it failed.

A restarted agent hands its sessions back instead of ending them — the tab keeps its place, the app stays on the screen you left it on, and no reinstall wipes what you had entered. An install error reaches the caller that asked for it. A device is not called ready when nothing is streaming. A command names the simulator it means rather than whichever one happens to be up. Full reset erases once, when asked, instead of on every later reconnect.

Full notes in `CHANGELOG.md`.

## Compatibility — minor is correct

`@tapflowio/protocol` is new this cycle, so its union diff is entirely additions: nothing removed, nothing changed shape. The three new browser-facing messages (`session:terminated`, `session:rebound`, `session:agent-away`) are additions to `RelayToBrowser`, and both downstream consumers match messages by predicate (`mcp-server/src/client.ts`, `flow-runner/src/RelayClient.ts`), so an unknown type is dropped rather than an error.

## Three things this release turned up

**`@tapflowio/protocol` was not on npm, and pushing the tag would have broken every install.** The workflow publishes through OIDC trusted publishing, which is configured per package on npmjs.com and cannot be set for a name that does not exist yet. `changeset publish` does not roll back — and `relay` and `mcp-server` declare `protocol` as a real dependency, with the CLI depending on both. One package failing would have put packages on npm pointing at a version that was never published. Seeded manually at `0.17.0` (not `0.18.0`, so the released version keeps its provenance signature) and trusted publishing is now configured. Nothing referenced the seed version: the published `relay@0.17.0` predates that dependency.

**`changeset version` rejected four changesets** that named `@tapflowio/dashboard` — which is `ignore`d — alongside published packages. The dashboard ships inside `relay`'s `public/`, so its notes moved there. Worth knowing: the CI gate cannot catch this, because it only checks that a changeset *exists* and never opens it. The convention is now in `AGENTS.md`; fixing the gate is a separate PR, so this one stays version bumps.

**`changeset:audit` reported two gaps and both are false positives** — verified rather than backfilled. #453's only non-test change is a devDependency on a private package plus a script line in the private root manifest; and the second is a back-merge of main into a feature branch, whose 32 files are main's own commits, each already covered. The audit fixes go with the gate fix above.

## Review

`.work/reviews/release__v0.18.0.md` — independent review skipped with the reason recorded: this is generated version bumps plus a changelog, and every change in it was reviewed in its own PR.

**Do not merge and forget** — `release.yml` fires on a `vX.Y.Z` tag, not on the merge. The tag goes on the merge commit afterwards.

## Summary by CodeRabbit

* **Release**
  * Released version 0.18.0 across the platform.
* **Improvements**
  * Improved recovery after agent disconnects or restarts, preserving sessions and viewer connections.
  * Added clearer, correlated errors for app installation, launch, and device startup issues.
  * Improved iOS simulator targeting, streaming readiness, reset behavior, and shutdown/boot handling.
  * Added typed communication contracts and stronger message validation without changing wire formats.
* **Documentation**
  * Updated changelogs and release documentation.


jo-duchan:
Taken — and it is the more interesting kind of wrong, because both sentences were true when they were written.

That changeset was authored while the receiving half was landing on its own, so "nothing sends `session:rebound`" and "the relay half is the next PR" described that PR accurately. A changelog describes a *release*, and 0.18.0 carries the sender and the grace window too. The result was a document telling a reader the message has no sender a few paragraphs above the entry describing it being sent. `will carry` had gone stale the same way.

Both entries now describe the shipped state. The sequencing stays, because that part is still worth knowing: the receiver had to land first, since a viewer that drops a message it does not recognise has no recovery path to fall back on.

I checked the rest of the release notes for the same shape — per-PR framing that a release cannot honour — and found none.

Worth recording beyond this PR: **a changeset for work split across PRs should not be written from the vantage point of its own PR.** Nothing catches that, because each changeset reads correctly at the moment it merges and only becomes false when the sibling PRs join it in the same release.

PR #474, opened 2026-08-05
fix(android-agent): handle scrcpy server process errors

dr-salman-ahmad:
## Summary

The scrcpy server process spawned in `ScrcpySession.start()` (real-device video streaming path) had no `.on('error', ...)` listener. Node emits `'error'` on a `ChildProcess` (an `EventEmitter`) asynchronously, and an unhandled `'error'` event throws — with no `uncaughtException` handler anywhere in this process, that crash took down the **entire** android-agent process, ending every device session it was managing, not just the one whose server process errored. 

This adds an `.on('error', ...)` handler that logs and continues, matching the pattern already used for every other spawned child process in this codebase (`EmulatorLauncher`, `EmulatorVideo`, ios-agent's `ScreenCaptureStreamer` and `KeyboardHelperDaemon`) — `ScrcpySession` was the one outlier missing it.

- `packages/android-agent/src/scrcpy/ScrcpySession.ts` — attach the error handler right after spawn, before either `.kill()` call site (`start()`'s catch block and `stop()`)
- `packages/android-agent/src/__tests__/ScrcpySession.test.ts` — regression test asserting `serverProc.emit('error', ...)` does not throw; verified it fails without the fix with the exact crash signature
- `.changeset/scrcpy-session-error-handler.md` — patch bump for `@tapflowio/android-agent`
- `CHANGELOG.md` — entry under `## [Unreleased]` → `### Fixed`

## Checklist

- [x] Tests written and passing (`pnpm --filter @tapflowio/android-agent test` — 208/208)
- [x] No `any`
- [x] Interface changes land in `agent-core` first — n/a, no interface change

## Related `.work/` docs

Adversarial review recorded in `.work/reviews/fix__scrcpy-session-error-handler.md` against HEAD `c007606a11f4da13883986215979e033f1a603c4` — two independent-context passes: one verifying the bug claim itself (no global exception handler catches this; every other `spawn()` call in the codebase already guards against it), one reviewing the actual diff (handler placement, test validity — confirmed non-vacuous by reverting and re-running — changeset/CHANGELOG correctness, and completeness of the fix). No issues found.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Android device connection stability by preventing session errors from crashing the agent or interrupting other active sessions.
  * Session-process errors are now safely logged for troubleshooting.

* **Tests**
  * Added regression coverage for session-process error handling.

* **Documentation**
  * Updated the unreleased changelog with the connection stability fix.


jo-duchan (commented):
Thanks for this, and for the unusually thorough write-up — it made the review much easier to follow.

The core diagnosis holds up. `ScrcpySession` really was the one `spawn()` call site in the repo without an `'error'` listener, and the placement (right after `spawn()`, ahead of both `.kill()` call sites) is exactly where it belongs.

I checked the regression test rather than taking it on faith. Reverting the one-line fix and re-running gives:

```
AssertionError: expected [Function] to not throw an error but 'Error: EPIPE' was thrown
 ❯ src/__tests__/ScrcpySession.test.ts:101
```

So it genuinely catches the regression. The full suite is 208/208 locally with the patch applied. The fake-proc timing in that test is easy to get subtly wrong, and you got it right.

Two things I'd like adjusted before this goes in.

### 1. The trigger described isn't the one that fires

The PR body, the changeset, `CHANGELOG.md`, and both new comments all attribute the crash to `kill()` racing an already-exited process. Node doesn't emit `'error'` in that case. `ChildProcess.prototype.kill()` treats `ESRCH` as a no-op and returns `false`:

```js
} else if (err === UV_ESRCH) {
  /* Already dead. */
}
```

(`lib/internal/child_process.js`)

Confirmed on node v24.15.0:

```
kill() on exited proc -> returned: false | error emitted: false
```

The fix is still worth having, to be clear. `spawn()` failures (`ENOENT`, `EACCES`, `EAGAIN` under process pressure) and `EPERM` from `kill()` do emit, and an unhandled one takes the whole agent down exactly as you describe. It's only the specific scenario that doesn't reproduce.

This matters most in `CHANGELOG.md`, since that one is user-facing. "A dropped real-device connection could take down every other Android device with it" states a cause we can't reproduce, and it'll ship in release notes as written. Could you reword the CHANGELOG entry and the changeset around the paths that do emit, or drop the specific scenario and describe it as an unhandled-`'error'` crash risk? The code comment and the test comment deserve the same treatment.

### 2. `logger.warn` → `logger.error`

The neighbouring handlers use `error` for this: `EmulatorLauncher.ts:66` and `XCUITreeReader.ts:127`. The stream is effectively over by the time this fires, so `error` matches both the convention and the severity.

---

One thing that is **not** for this PR, just noting it so it isn't lost: `EmulatorVideo` calls `onDead(...)` on the same event so the session gets torn down, whereas here the process can die while the session still looks alive to the dashboard. The absent `'exit'`/`'close'` handlers are the same gap. That's real, but it's a different problem from the crash you're fixing, and I'd rather keep this PR scoped as it is. I'll open a follow-up issue for it.

CI hasn't run yet — workflows need maintainer approval for first-time contributors. I'll take care of that so you get a real signal on the next push.

Thanks again, this is a good catch and a well-argued one.

dr-salman-ahmad:
Thanks for @jo-duchan  the detailed review — appreciate you verifying against the Node source and confirming the test still catches the regression.

You're right on both points:

1. Reworded the changeset, CHANGELOG.md entry, and both comments (code + test) to describe the actual crash paths — spawn() failures (ENOENT/EACCES/EAGAIN) and EPERM from kill() — instead of the kill()-on-exited-process scenario, which I now see is a documented no-op.
2. Switched logger.warn → logger.error to match EmulatorLauncher.ts and XCUITreeReader.ts.

Pushed in 29bf7f9. Sounds good on keeping the exit/close gap as a separate follow-up.

jo-duchan:
This all looks right. The reworded changeset, CHANGELOG entry and both comments land where I hoped, and rewriting the CHANGELOG headline rather than just the parenthetical was the right call — that was the line that would have shipped. `logger.error` matches the neighbours too.

Verified locally on 4364aa5 with main merged in: android-agent is 208/208 and eslint is clean.

One small leftover, and it's the description rather than the code. The Summary still carries the original attribution:

> A realistic trigger: `serverProc.kill()` racing an already-exited process, e.g. after the physical device disconnects.

Your commit message on 29bf7f9 already sets the record straight, and we merge with a merge commit, so the git history is fine either way. It's just that the PR page is where someone lands from a `git blame` a year out, and it would be a shame for it to still teach the one version we established doesn't happen. Could you edit the description when you get a moment? Housekeeping only — not blocking.

CI still hasn't run: workflows need maintainer approval for first-time contributors, and that's on me rather than you. I'll get it going so you have a real signal.

Thanks also for #479 — I'll review that one on its own thread.

dr-salman-ahmad:
> This all looks right. The reworded changeset, CHANGELOG entry and both comments land where I hoped, and rewriting the CHANGELOG headline rather than just the parenthetical was the right call — that was the line that would have shipped. `logger.error` matches the neighbours too.
> 
> Verified locally on [4364aa5](https://github.com/jo-duchan/tapflow/commit/4364aa50bf946e8f2337d608982d2ecc512272cd) with main merged in: android-agent is 208/208 and eslint is clean.
> 
> One small leftover, and it's the description rather than the code. The Summary still carries the original attribution:
> 
> > A realistic trigger: `serverProc.kill()` racing an already-exited process, e.g. after the physical device disconnects.
> 
> Your commit message on [29bf7f9](https://github.com/jo-duchan/tapflow/commit/29bf7f93234e0fce3c64363fb90689e6ba526eb0) already sets the record straight, and we merge with a merge commit, so the git history is fine either way. It's just that the PR page is where someone lands from a `git blame` a year out, and it would be a shame for it to still teach the one version we established doesn't happen. Could you edit the description when you get a moment? Housekeeping only — not blocking.
> 
> CI still hasn't run: workflows need maintainer approval for first-time contributors, and that's on me rather than you. I'll get it going so you have a real signal.
> 
> Thanks also for #479 — I'll review that one on its own thread.

Good catch, thanks — that line was a leftover from before we corrected the trigger everywhere else. Edited the description to remove it; the rest is unchanged. Appreciate you re-verifying locally and catching this.

Issue #485, opened 2026-08-07
dashboard: input:done / input:error are declared and never handled — the tester is not told an input was dropped


jo-duchan:
Found while reviewing #484 (the #482 fix). Filed separately so that PR could stay inside one package.

## What happens

The relay forwards `input:done` / `input:error` to the browser, and the dashboard drops them on the floor. Both types are **declared** and never handled:

```ts
// packages/dashboard/lib/types.ts:164
| { type: 'input:done'; sessionId: string }
| { type: 'input:error'; sessionId: string; message: string }
```

`grep` for a handler across `packages/dashboard` returns nothing — no `case 'input:error'`, no listener. Compare `device:boot-error`, which is handled in both `SessionList.tsx` and `DeviceViewer.tsx`. The relay's side is fine (`RelayServer.ts`, `TERMINAL_INPUT_TYPES`), so the message arrives and is discarded.

The only consumer today is `mcp-server` (`src/client.ts`), which is the **experimental** path. So the manual testing path — the one AGENTS.md names as the default perspective — is the one that hears nothing.

## Why it matters now

Before #484 the agent lied: a dropped input was acked as `input:done`. After it, the agent tells the truth, and the truth is thrown away before it reaches a human. Concretely, after #484 a tester on a session whose input channel has permanently failed (a missing or wrong-architecture helper binary — #464 — or a helper that cannot start) sees:

- a stream that keeps updating normally
- taps that do nothing
- no indication anywhere in the UI

which is the original #482 symptom minus the recovery. #484 does help the common case without any dashboard work — the helper revives itself, so the next tap lands — but the unrecoverable case is exactly the one where a message exists and nobody shows it.

Worth stating plainly: this makes the wire-level assertions in #484's tests weaker than they look. `expect(browser)` receiving `input:error` is true and proves the agent's half; it does not mean a user learns anything.

## What needs deciding first

Not "add a toast". The interesting part is what an input failure should look like, and the answer is not obvious:

- **Per-input or per-session?** A dead channel produces one `input:error` per tap. Showing each one would spam; showing none is today's behaviour. A session-level "input unavailable" state that the first error latches, and a recovered input clears, is probably closer — but the protocol has no session-level input-health signal, only per-input acks. #455 is holding a related decision (whether to add an error reply or check state first), and the two should be answered together rather than each inventing its own shape.
- **Where?** The viewer, the session list, or both. `device:boot-error` already occupies some of this space and its treatment is worth matching rather than duplicating.
[...]

Issue #457, opened 2026-08-02
mcp-server: a dropped input is reported as success — the optimistic ack fallback outlived its reason


jo-duchan:
## What happens

`TapflowClient.awaitInputAck` waits 2s for `input:done` / `input:error`, and on timeout **returns successfully**:

```ts
// packages/mcp-server/src/client.ts:188
} catch (e) {
  // Only a timeout means an older agent/relay that never acks → assume dispatched.
  if (e instanceof Error && e.message === 'Request timed out') return
  throw e
}
```

So `tap`, `swipe` and every other terminal gesture report success when nothing acked. The LLM agent driving the session is told the tap landed and moves on.

## Why the reason no longer holds

The fallback exists for agents and relays that predate the input-ack protocol. Both current agents ack terminal inputs unconditionally:

- `packages/ios-agent/src/IOSAgent.ts:671`
- `packages/android-agent/src/AndroidAgent.ts:910`

Each sends `input:done` when it dispatched to a booted device and `input:error` when there is no live channel. There is no branch where a current agent stays silent. When both ends are current, a timeout means something is actually wrong — and that is exactly the case the fallback converts into a success.

## Not fixed by the relay work in #426

Stage 2 of #426 makes the relay answer `input:error` itself when it knows the session has no stream (`readySent === false`), which covers the rebind window and pre-boot input. That shrinks the silent window; it does not close it. An agent that holds device state but fails inside the input path still says nothing, and that still reads as success.

## Suggested shape

Capability-gated, so old deployments keep working:

1. `AgentCapability` gains `'input-ack'` (`@tapflowio/protocol`).
2. Both agents add it to their `AGENT_CAPABILITIES` — currently `['clipboard']`, declared separately in each agent (`IOSAgent.ts:14`, `AndroidAgent.ts:47`; worth collapsing into one exported constant while touching them).
3. The relay already echoes capabilities on `session:joined`, and on `session:rebound` after #426 stage 2.
4. `mcp-server` captures the joined session's capabilities — **it ignores them entirely today** — and makes the timeout fatal when `input-ack` is advertised, keeping the optimistic path only when it is not.

Four packages, so this needs a design review before code per AGENTS.md.

## Worth deciding

Whether the 2s window is still right once a timeout is fatal. The comment calls acks "near-instant", but a fatal deadline deserves more headroom than an advisory one.

Issue #482, opened 2026-08-06
ios-agent: a dead touch-helper silently drops every tap for the rest of the session


jo-duchan:
Surfaced while reviewing #479. That PR correctly defends against the *symptom* — `stop()` no longer arms a SIGKILL timer against a process that already exited — but the condition underneath it stays, and it has a user-visible consequence of its own. Filing separately so #479 can merge as scoped.

## What happens

`TouchHelper`'s `'exit'` and `'error'` handlers log and nothing else:

```ts
this.proc.on('exit', (code) => {
  logger.error(`exited with code ${code ?? 'null'}`)
})
this.proc.on('error', (e) => {
  logger.error(`spawn error: ${e.message}`)
})
```
(`packages/ios-agent/src/TouchHelper.ts:24`)

`this.proc` keeps pointing at the dead process. Every write path then guards on the stdin it can no longer reach:

```ts
private write(type: number, x: number, y: number): void {
  if (!this.proc?.stdin?.writable) return
```
(`TouchHelper.ts:130`, and the same line at 54, 66, 84, 108, 120)

So each tap returns early. No throw, no log, no message back to the browser.

Nothing upstream notices either. `IOSAgent.ensureTouchHelper()` tests the *wrapper*, not the process:

```ts
private ensureTouchHelper(state: DeviceState): void {
  if (state.touchHelper) return
  state.touchHelper = new TouchHelper(state.deviceId)
```
(`IOSAgent.ts:638`)

`state.touchHelper` is still a live object, so it returns early and never respawns.

**Net effect: if `touch-helper` dies on its own, that session accepts no further input — and the stream keeps running, so the device looks perfectly healthy.** The viewer taps a screen that updates normally and does nothing. Recovery needs a reconnect, one of the paths that nulls `state.touchHelper` (`IOSAgent.ts:364`, `496`, `614`).

The comment at `IOSAgent.ts:637` says input "self-heals without a fresh `device:boot`" — that holds for a *reconnect*, which reissues the session with `touchHelper: null`. It does not cover the helper process dying while the session stays up.

## `KeyboardHelperDaemon` already solves this

Same package, same kind of helper, complete recovery:

```ts
proc.on('exit', (code) => {
  logger.error(`exited with code ${code ?? 'null'}`)
  this.proc = null                    // ← so the next command respawns
  ...
  if (pending) pending.reject(new Error(`keyboard-helper exited with code ${code}`))
})
```
[...]

PR #490, opened 2026-08-07
feat(protocol): give input:error a machine-readable reason, so a caller can tell retry from reconnect

### LINKED ISSUE #489: ios-agent: a terminal input for a session the agent has lost is answered by nobody — the caller's own fallback reports success
Found by a design review while doing the Android half of #482 (#488). The Android plan claimed this case was Android-specific and that iOS "reaches its handlers only with a live session". The reviewer refuted it, and the code agrees.

## What happens

All four terminal input handlers return silently when the agent holds no state for the session the message names:

```ts
// packages/ios-agent/src/IOSAgent.ts
case 'input:touch:end':  if (!state) break   // :769
case 'input:pinch:end':  if (!state) break   // :793
case 'input:key':        if (!state) break   // :874
case 'input:button':     if (!state) break   // :900
```

No `input:done`, no `input:error`. And the relay only answers on an agent's behalf when the **agent socket is closed** (`RelayServer.ts`, `TERMINAL_INPUT_TYPES` → `'agent offline'`). With the agent online and the session simply unknown to it, nothing answers at all — so `mcp-server` waits out its 2s window and its own fallback reports **success** (#457).

That is the same silent-success shape #484 removed from every other iOS terminal path.

## More reachable on iOS than on Android

`IOSAgent` clears its device state on both teardown paths:

```
IOSAgent.ts:300  deviceStates.clear()   // disconnect()
IOSAgent.ts:315  deviceStates.clear()   // _scheduleReconnect()
```

Android has **no** `deviceStates.delete` and no `clear` outside the same two paths, so on that side an unknown sessionId is genuinely unregistered. On iOS the reconnect path *drops sessions the relay still holds and still routes for* — which is exactly the shape #426 is about (an agent restarts, the dashboard tab is still attached). So the window is wider here, not narrower.

## Why it was not fixed in #488

That PR is `android-agent` only, deliberately — keeping it to one package kept the review surface small and avoided the design-review-for-cross-cutting-changes rule. The Android side now answers `no-session` with its own message; iOS is the remaining half.

## What the fix looks like

Android's shape ports directly, and the interesting part is what to reuse rather than how to write it:

- The Android side introduced an outcome vocabulary (`packages/android-agent/src/inputOutcome.ts`) because it needed five reasons iOS does not have. iOS needs exactly **one** new reason, so copying the whole vocabulary would be over-building — but `ackInput` there takes a `dispatched: boolean` and cannot express a third answer at all, so something has to give. Deciding that is the design question, not the four `break`s.
- Whatever is chosen, the message string should match Android's (`'no active session for this device on this agent'`) or both should change together. Two agents answering the same situation with different text is worse than either wording.
- Worth checking at the same time whether `input:touch:start` / `input:pinch:start` should stay silent. They have no ack obligation, so Android left them alone, and the same reasoning should hold here — but it should be a decision rather than an omission.

## Not the same as #457

#457 is the MCP client converting an ack timeout into success. This is the agent not answering in the first place. Fixing #457 alone would turn this into a *timeout* rather than a false success, which is better but still not an answer; fixing this alone makes the answer arrive while #457's optimistic path stays for genuinely slow dispatches. They are independent.

## Pointers

- `packages/ios-agent/src/IOSAgent.ts:769`, `:793`, `:874`, `:900` — the four silent exits
- `packages/ios-agent/src/IOSAgent.ts:300`, `:315` — where the state is dropped while the relay keeps the session
- `packages/ios-agent/src/IOSAgent.ts` — `ackInput`, which currently takes a boolean
- `packages/android-agent/src/inputOutcome.ts` and `AndroidAgent.ts`'s `ackNoSession` — the shape that landed on the other side
- `packages/relay/src/RelayServer.ts` — `TERMINAL_INPUT_TYPES` and the agent-offline reply, i.e. why nothing else covers this

jo-duchan:
Closes #489. Groundwork for #485 — the dashboard cannot render "why" until "why" is on the wire.

## The problem

`input:error` carried human prose and nothing else, so three situations that call for three different responses arrived looking identical:

| situation | what the caller should do | what it saw |
|---|---|---|
| the input channel is still starting (measured 186–247ms on iOS) | **wait ~200ms and retry** | `input channel not ready` |
| the channel is gone | reconnect | `input channel not ready` |
| this agent does not implement the input | never retry | `input channel not ready` |

With nothing to branch on, a caller could only give up or blindly retry.

This is *not* the same gap as `mcp-server`'s optimistic timeout fallback — that fires when no ack arrives at all, and no field on a message that never arrives could inform it. That is #457. An earlier draft of this PR's description claimed otherwise and was wrong.

## The design

`InputErrorReason` is a closed string-literal union in `@tapflowio/protocol`, and its set is derived from **what a consumer must do differently** — not from how many internal states an agent has. iOS has one input path (a HID helper process); Android has three (a scrcpy socket, an emulator gRPC channel, `adb shell input`). Each agent maps its own states onto the set, so the wire contract stays one size while the platforms stay different.

Three things made this land where it did, after two earlier plans were discarded in design review:

- **The vocabulary belongs in `protocol`, not `agent-core`.** `dashboard`, `mcp-server` and `relay` already depend on `protocol`; none depends on `agent-core`. An earlier plan unified the vocabulary there and would have achieved type agreement *between the two producers* while the consumer still received prose.
- **`protocol`'s entry point must stay runtime-free**, so a shared message table is impossible — and that is the right answer, not a constraint to work around. `reason` is the machine field; `message` stays free prose, which is what lets iOS keep `unknown key code: KeyFoo` while adding `reason: 'unsupported'`.
- **`DeviceAgent` is untouched.** A repo-wide grep found zero production callers of its three touch methods; the acks come from the ws handlers. Threading an outcome through a signature nobody calls would have been a breaking change bought for nothing.

## Two reasons that had no name

**`channel-starting`.** iOS's helper needs 186–247ms after spawn before an injected frame reaches the device, and `device:ready` can arrive inside that window — so a caller tapping as soon as a boot returned was told the channel was gone when it was merely coming up.

**`no-gesture`.** A frame that only means something as part of a gesture, arriving with no gesture behind it, is well-formed on a channel that may be perfectly healthy. The move is to open a new gesture — neither "fix the call" nor "wait". iOS reaches it through its gesture-ownership guard; Android through a terminal frame the viewer sends on any pointerup that did not start on the video.

## Ownership before readiness, and why that is the substance

The pre-PR review found the derivation inverted for its own headline case, and both channels found it independently.

Readiness is about *now*; ownership was settled when the gesture *opened*. A gesture whose opening frame was refused inside the start-up window owns nothing — so by the time its terminal frame arrives the helper reads ready, and a readiness-first derivation answered `malformed`, i.e. **"never retry"**, for exactly the sequence `channel-starting` exists to serve. MCP's `swipe` defaults to 300ms, comfortably past the measured 247ms, so it landed there every time.

The fix consults ownership first for continuation frames, with "no channel at all" outranking both. A consequence worth knowing, now documented: **`channel-starting` is unreachable for a continuation frame**, since owning a gesture requires readiness in the first place. Only standalone inputs — a key, a button — are ever refused merely because the channel is coming up.

The test that should have caught it could not: it pinned the mock's `inputState()` to a constant, so the *transition* between the two frames never happened. The agent layer mocks that class, so the transition is only observable at the unit layer, and that is where the test now lives.

## `TouchHelper` keeps its `boolean` returns

Deliberately. Every member of a string union is truthy, so converting them would silently invert `this.gestureProc = sent ? this.proc : null` — the guard two earlier reviews already fought over — and neither `tsc` nor eslint would report it, since `no-unnecessary-condition` is not enabled and tests are excluded from both. The reason is derived at the ack site instead.

## Not a breaking change

The field is optional: an agent predating it omits it and nothing breaks. Absence therefore means *unknown*, never *fine*, and a consumer meeting an unfamiliar reason must treat it as `channel-unavailable`. Making it required is the breaking step and is not taken here.

## A gate hole this work depended on

The root `typecheck` never ran `@tapflowio/protocol`'s own script, so `typeAssertions.ts` — the file that exists to catch wire-contract drift — had never been compiled, in the pre-commit hook or in CI. Making `reason` required passes the whole test suite and fails only that check, which is how the hole surfaced. Fixed here.

## Review

Two design rounds (which discarded two plans outright) and one pre-PR round, two independent channels each. Recorded in `.work/reviews/feat__input-error-reason.md`.

9 mutations, all now caught. Three survived at first and all three were real: `inputState()` forced to a constant (needed a unit test, since agent tests mock the class), `reason` made required (needed the gate fix above), and the deferred software-keyboard branch's reason replaced by a constant — which would have lost this PR's whole distinction for every key pressed while the keyboard is up.

Also corrected during review: a misplaced `export` keyword that had silently un-exported `IOS_BUTTON_ALIASES`, and two of my own tests that proved themselves — a static check whose regex listed the names it then asserted membership for, and a mapping test comparing two hand-written literals in the same file while claiming the compiler enforced it.

## Out of scope

- **#457** — acting on the reason (retrying `channel-starting`, dropping the optimistic fallback for reasons that say never retry). This PR only carries it.
- **#485** — the dashboard rendering it. That is next, and this is what it needed.
- Making `reason` required; adding it to the relay's own `agent offline` reply; the write-nothing button branch's behaviour on a starting channel. All recorded with reasons in the review record.

## Summary by CodeRabbit

* **New Features**
  * Added optional machine-readable reasons to input errors, including unavailable channels, startup, unsupported input, and gesture issues.
  * iOS and Android acknowledgements now provide consistent, actionable error details while preserving existing messages.
  * MCP errors now include available input error reasons.

* **Bug Fixes**
  * Improved distinction between unavailable, starting, unsupported, and gesture-related input failures.

* **Documentation**
  * Documented input error meanings, compatibility behavior, and cross-platform handling.


jo-duchan:
CodeRabbit's finding is correct and is fixed in d20e997.

**What changed.** iOS's four terminal handlers (`input:touch:end`, `input:pinch:end`, `input:key`, `input:button`) answered nothing when `deviceStates` held no state for the message's session — `if (!state) break`. The relay replies on an agent's behalf only when the agent socket is *down*, so in this shape nothing answered at all: the caller waited out its own timeout, and MCP's fallback reports that as success. They now send `input:error` with `reason: 'channel-unavailable'`, which is what `android-agent`'s `wireReason()` maps the same situation to, so the two agents no longer differ here. Opening frames stay silent — nothing is waiting on them.

**My skip rationale was too strong, and that is the more useful part of this.** The plan and the review record both justified the skip with "`sessionRebind.test.ts:328-340` already refutes reachability". On re-checking, that test does not support the conclusion: `RelayServer`'s rebind path builds `registeredSessions` from `byDeviceId`, a Map keyed by deviceId, so `agent:registered` carries **one entry per device** — and the relay's own comment at that spot says that before rebinding existed one device could not sit behind two sessions but now can, leaving the second one unannounced. So re-seeding is not guaranteed to cover every session.

What actually decided it was not reachability but a self-contradiction in the PR: the scope table claimed "#490 closes #489", and #489 is titled *nobody answers*. Claiming to close it while iOS stayed silent does not hold. The cost asymmetry points the same way — four lines of possibly-dead code against a swallowed terminal input reported as success, which is the exact bug class #482/#484/#488 were about. Reachability is therefore **not claimed** either way; the dispute is recorded verbatim in the code comment and in both agents' AGENTS.md.

**Also corrected** (the second half of the comment): the "unfixed asymmetry" statement at `packages/android-agent/AGENTS.md:46`, plus the same claim in both doc blocks of `packages/android-agent/src/inputOutcome.ts`.

**Coverage.** 5 new tests — four terminal types answer with the reason, one pins that an opening frame stays silent. Three mutations, all caught: reverting the handlers to silent `break` (4 failures), changing the reason to `channel-starting` (4), and answering opening frames too (1). The last one matters: without it, someone could later "fix the asymmetry" by answering every frame.

Full suite 2158 passed + 1 skip, typecheck/lint/build clean. Review record updated at `.work/reviews/feat__input-error-reason.md` with the round and the corrected rationale.

jo-duchan:
Skipping this one, with the measurement.

**The wording is already pinned.** I emptied `INPUT_ERROR_MESSAGES['channel-unavailable']` and ran the iOS suite: **8 tests fail** (`IOSAgent.test.ts` 618, 647, 661, 693, 731, 744, 781, …). "A change that breaks the wording will pass this test" is true of the new test read in isolation and false of the suite, which is what actually guards the string. A ninth assertion of the same literal would be duplication.

**And the premise misdescribes the wording.** It is not Android-compatible and is not meant to be — Android's prose for this situation is `'no active session for this device on this agent'`, and its own test asserts that. Only `reason` is shared. `packages/protocol/AGENTS.md` states the split explicitly: `message` stays free prose per agent and `reason` is the closed contract, which is exactly what lets iOS keep `unknown key code: ${code}`. Pinning this literal in one more place pushes the prose toward being a de-facto contract, so a wording improvement would start breaking tests for no gain.

Recorded in `.work/reviews/feat__input-error-reason.md` as mutation M13, with the 8-failure count as the evidence.

Issue #464, opened 2026-08-03
native helper binaries are arm64-only — Intel (x86_64) Macs fail with "spawn unknown error" (EBADARCH)


aptonline:
**Summary**

On an Intel Mac (x86_64), opening any build in the dashboard fails with a generic "spawn unknown error". Root cause: all five native helper binaries shipped in the npm package are Mach-O arm64 only. macOS refuses to exec them on Intel hosts with EBADARCH, which Node surfaces as Unknown system error -86 → the dashboard shows "spawn unknown error".

Affected binaries (all arm64-only, tapflow 0.18.0)

```
- @tapflowio/ios-agent/bin/touch-helper
- @tapflowio/ios-agent/bin/keyboard-helper
- @tapflowio/ios-agent/bin/rotation-helper
- @tapflowio/ios-agent/bin/screencapture-helper
- @tapflowio/audiotap-helper/bin/audiotap-helper.app/Contents/MacOS/audiotap-helper
```

Verified with file and lipo -info on the installed package.

**Reproduction**

1. Install tapflow 0.18.0 on an Intel Mac (uname -m → x86_64), tapflow start.
2. Upload any iOS build, open it in the dashboard.
3. Session never starts — agent reports "spawn unknown error".

Or directly:
```
node -e "const {spawnSync}=require('child_process'); \
  const r=spawnSync('/usr/local/lib/node_modules/tapflow/node_modules/@tapflowio/ios-agent/bin/touch-helper',[]); \
  console.log(r.error ? r.error.message : 'ok')"
# → error: spawnSync ... touch-helper Unknown system error -86
```
Shell equivalent: Bad CPU type in executable.

**What works (to isolate)**

- xcrun simctl boot <udid> ✓
- tapflow boot "iPhone 17 Pro" ✓ (agent boots sims fine)
- tapflow devices, tapflow status, upload API ✓
- Everything fails at the moment the agent spawns a native helper (screen capture, touch, keyboard, rotation, audio).

**Root cause detail**

The npm packages ship the helpers as prebuilt binaries only — the Swift sources are in the GitHub repo (packages/ios-agent/src/*-helper.swift, packages/audiotap-helper/src/audiotap-helper.swift) but are not included in the published tarball. The mtime-gated recompile paths (ensureCompiled() in ScreenCaptureStreamer.js, ensureHelperApp() in audiotap-helper) explicitly skip when source is absent:
[...]

PR #494, opened 2026-08-08
fix(dashboard): tell the tester when an input never reached the device

### LINKED ISSUE #485: dashboard: input:done / input:error are declared and never handled — the tester is not told an input was dropped

jo-duchan:
Closes #485.

`input:done` / `input:error` were declared in `packages/dashboard/lib/types.ts` and handled nowhere, so the relay forwarded them and the browser dropped them. The only consumer of a failed input was `mcp-server` — the experimental path — leaving manual testing, which AGENTS.md names as the default perspective, hearing nothing.

That mattered more after #484/#488/#490. Before those, an agent reported a dropped input as success; now it reports the truth with a machine-readable `reason`, and the truth was being discarded before a human saw it. A session whose input channel has permanently failed (a helper binary missing or built for the wrong architecture, #464) showed a stream that kept updating, taps that did nothing, and no indication anywhere.

## What a tester sees now

A failed input raises a toast whose copy is chosen by `reason`, so they are told **what to do** rather than only that something failed. `not-booted` gets its own wording because the protocol prescribes a different action for it (boot the device) than for `channel-unavailable` (reconnect). The wire `message` rides along as the description, where its detail (`unknown key code: KeyFoo`) is useful — it is not the headline, because it is free prose each agent owns and cannot be localised.

`channel-starting` and `no-gesture` are shown nowhere: both are fixed by the time the message could be read. Nothing is shown while the agent is away either — the relay answers every terminal input itself in that state, and the viewer already says the session is being held open.

## There is no session-level "input unavailable" state, and that is the interesting part

That was the first design — a latched line on the status card, cleared by `input:done` and the boot lifecycle. A two-channel design review discarded it, and each channel produced a blocker independently. Per-input acks cannot support a latch:

- **No message carries evidence that input is working again.** iOS replaces a dead helper eagerly and is injecting ~200ms later silently. Lifecycle messages do arrive (`session:rebound`, `session:joined`) but neither is evidence of input *health* — with the binary still missing, a latch would clear and re-raise on the next tap. The only signal that means anything is a successful input, and `channel-unavailable`'s own advice is *do not blindly retry*.
- **The acks are unordered.** A dispatch is awaited before its ack while a refusal is not, so an earlier input's `input:done` can land after a later input's `input:error`.
- **An ack does not say which channel answered.** On Android buttons always take the adb path while touch takes the pointer channel, so a tester pressing Home to check whether input works at all succeeds — and that success erased the warning about the dead touch channel. Both review channels found this sequence independently.

The toast's own lifetime carries it instead: repeats reuse one `id`, which sonner refreshes rather than stacks, so it stays up while inputs keep failing and fades on its own when they stop. **Observed rather than stored**, so there is no clear edge to get wrong. Two tests guard the decision, and `AGENTS.md` records why, because a latch will look like the obvious improvement to whoever reads this next.

## Review

`.work/reviews/fix__dashboard-input-error-unhandled.md` — design round (2 channels, discarded the design) plus a pre-PR round (2 channels) against `d74b49d`.

The pre-PR round is worth summarising because of what it found. The first channel was handed the ten mutations I had already run and asked what they *missed* — **four survived**, and all four were things documented but not guarded:

- `duration: 6000` was read by no test, and it is the only state this design has. `Infinity` made the toast permanent — the opposite of the documented behaviour — with nothing failing.
- The two tests said to guard the discarded latch watched only `toast.dismiss`, so a latch rebuilt as `toast.info('Input is working again')` passed both.
- `notice.action` was never checked through the render path, so dropping it passed, as did `not-booted` sharing `channel-unavailable`'s action.
- The `agentAway` interaction was reachable and pinned in neither direction. Now decided and tested.

The second channel checked whether the prose is true and found **five false claims**, including one that named the wrong backend (`socket.writable` is scrcpy's signal, not gRPC's) and one overstatement in the very paragraph written to stop the latch being rebuilt. All corrected; details and evidence in the record.

16 mutations total, 6 of which survived initially. None of my own ten touched those six — they had followed what I implemented rather than what I claimed.

## Scope

Dashboard only. `SimulatorInfoCard`, `IOSViewer` and `AndroidViewer` are untouched. The changeset names `@tapflowio/relay`, since that is the package the built dashboard ships inside.

Two follow-ups recommended in the record: a persistent session-level indicator needs a protocol-level input-health signal (or acks that name their channel and arrive in order), and the dashboard's inbound `RelayMessage` union is the remaining half of the drift `@tapflowio/protocol` was created to remove.

Full suite 2175 passed + 1 skip (dashboard 275 → 292); typecheck, lint, build and a11y-lens clean.

## Summary by CodeRabbit

* **New Features**
  * Added clearer, reason-specific notifications when dashboard input actions fail.
  * Included relay-provided error details in notifications for easier troubleshooting.
  * Reused notifications for repeated failures to reduce duplicate alerts.

* **Bug Fixes**
  * Suppressed transient errors and notifications when the agent is away.
  * Improved handling of unknown or missing error reasons with a fallback notice.
  * Prevented completion events from incorrectly clearing active error notifications.

### DIFF
--- a/.changeset/dashboard-input-error.md
+++ b/.changeset/dashboard-input-error.md
@@ -27,11 +27,15 @@ and `no-gesture` (the gesture is gone; a fresh one works). Reporting an error fo
 fixed by the time it is read is noise.
 
 There is **no session-level "input unavailable" state**, which was the first design and was discarded
-after review. Per-input acks cannot support one: nothing on the wire announces that a replaced helper
-is healthy again, the acks are not ordered (a success awaits a child process while a refusal is sent
-synchronously), and an ack does not say which channel answered — so on Android a working Home button
-would have erased a warning about a dead touch channel. The toast's own lifetime carries it instead:
+after review. Per-input acks cannot support one: no message carries evidence that input is working
+again — a replaced helper announces nothing, and an agent restart is not the same as a healthy channel —
+the acks are not ordered (a dispatch is awaited before its ack while a refusal is not), and an ack does
+not say which channel answered, so on Android, where buttons always take the adb path, a working Home
+button would have erased a warning about a dead touch channel. The toast's own lifetime carries it instead:
 repeats reuse one id, so it stays up while inputs keep failing and fades when they stop.
+
+Nothing is shown while the agent is away, either: the relay answers every terminal input itself in that
+state, and the viewer already says the session is being held open and waiting.
 
 Dashboard-only change, released as part of `@tapflowio/relay` because that is the package the built
 dashboard ships inside.
--- a/.changeset/dashboard-input-error.md
+++ b/.changeset/dashboard-input-error.md
@@ -1,0 +1,37 @@
+---
+'@tapflowio/relay': patch
+---
+
+fix(dashboard): tell the tester when an input never reached the device
+
+The relay forwarded `input:done` / `input:error` to the browser and the dashboard dropped both. They
+were declared in `lib/types.ts` and had no handler anywhere, so the only consumer of a failed input
+was `mcp-server` — the experimental path. Manual testing, which is tapflow's primary use, heard
+nothing.
+
+That mattered more after #484/#488/#490. Before those, an agent reported a dropped input as success;
+now it reports the truth with a machine-readable `reason`, and the truth was being discarded before it
+reached a human. Concretely, a session whose input channel has permanently failed (a helper binary
+that is missing or built for the wrong architecture) showed a stream that kept updating, taps that did
+nothing, and no indication anywhere.
+
+A failed input now raises a toast whose copy is chosen by `reason`, so the tester is told what to do
+rather than just that something failed — reconnect, start the device, report a bug, or that the input
+has no equivalent on this device. `not-booted` gets its own wording because the protocol prescribes a
+different action for it than for `channel-unavailable`. The wire `message` rides along as the
+description, which is where its detail (`unknown key code: KeyFoo`) is useful; it is not used as the
+headline, because it is free prose each agent owns and cannot be localised.
+
+Two reasons are deliberately shown nowhere: `channel-starting` (the input channel is up ~200ms later)
+and `no-gesture` (the gesture is gone; a fresh one works). Reporting an error for something already
+fixed by the time it is read is noise.
+
+There is **no session-level "input unavailable" state**, which was the first design and was discarded
+after review. Per-input acks cannot support one: nothing on the wire announces that a replaced helper
+is healthy again, the acks are not ordered (a success awaits a child process while a refusal is sent
+synchronously), and an ack does not say which channel answered — so on Android a working Home button
+would have erased a warning about a dead touch channel. The toast's own lifetime carries it instead:
+repeats reuse one id, so it stays up while inputs keep failing and fades when they stop.
+
+Dashboard-only change, released as part of `@tapflowio/relay` because that is the package the built
+dashboard ships inside.
--- a/packages/dashboard/AGENTS.md
+++ b/packages/dashboard/AGENTS.md
@@ -56,16 +56,25 @@ A latched "input unavailable" line on the status card was designed and discarded
 like the obvious improvement to whoever reads this next. It cannot be made honest on the current
 protocol, for three independent reasons:
 
-- **Nothing announces recovery.** iOS replaces a dead helper eagerly and is injecting again ~200ms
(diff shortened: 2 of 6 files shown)


jo-duchan:
Fixed in 25e4369.

The finding is right, and the reason I missed it is structural: the **silent** path had a `console.debug` (added by one of the pre-PR review channels) and the **shown** path had none — and an unrecognised reason takes the shown path. So it was normalised to `channel-unavailable`, the tester saw the correct conservative copy, and the value that had arrived went nowhere. A dashboard running behind its agents was indistinguishable from one looking at a genuinely dead channel, which is exactly the situation this union's growth guarantees: the symptom would be "everything says reconnect" with nothing pointing at why.

Absence is deliberately **not** logged. An agent older than #490 omits the field on every single input, so that would be steady noise about a state the protocol documents as normal. Only an unrecognised *value* is news, so the condition is `msg.reason !== undefined && msg.reason !== key`.

Two mutations, both caught: removing the trace (1 failure), and logging on absence too (1 failure) — the second matters as much as the first, since a log that fires for a documented-normal state trains people to ignore it.

Dashboard suite 292 → 294. Full suite 2177 passed + 1 skip; typecheck, lint, build and a11y-lens clean. Review record updated at `.work/reviews/fix__dashboard-input-error-unhandled.md`.

Issue #491, opened 2026-08-08
protocol: the guaranteed half of input:error is the half that is not a contract — invert it


jo-duchan:
## What

`input:error` currently declares:

```ts
{ type: 'input:error'; sessionId: string; message: string; reason?: InputErrorReason }
//                                       ^^^^^^^^^^^^^^^ required, but explicitly NOT a contract
//                                                        ^^^^^^^ the contract, but optional
```

`packages/protocol/AGENTS.md` states the split on purpose: `message` is free prose each agent owns, and `reason` is the closed union consumers branch on. So the field a consumer is guaranteed to receive is the one it must not depend on, and the field it should depend on may be absent.

That inversion was the correct way to *ship* it (#490) — an agent predating the field omits it and nothing breaks. It is not a correct end state, and this issue is the scheduled other half, not a maybe. As long as `reason` is optional, every consumer carries an unknown-reason branch forever, and "absence means unknown" has to be re-derived by each new consumer.

## Two changes, and one that is genuinely conditional

**1. `reason` becomes required.** The whole point of the field is that a caller branches on it; today it cannot branch without a fallback path that exists only because of version skew.

**2. `message` becomes optional.** It currently does two jobs: it carries parameterised diagnostic detail (`unknown key code: KeyFoo` — information `reason: 'unsupported'` loses) and it is display copy. The second job does not belong to an agent — display strings belong in the presentation layer, and agent-authored English cannot be localised, which matters because tapflow's audience is PO / PM / designers / QA. Once the dashboard renders its own copy from `reason` (#485), `message` is a debug/log/forward-compat field, and a debug field should not be required.

**3. `params?: Record<string, string | number>` — conditional, do not add speculatively.** It is the honest way to keep `KeyFoo` once prose is demoted. But it only pays off if a rendered UI actually needs the variable. #485 is what will say whether anyone misses it. Adding it before then means guessing a schema from no requirement.

## Prerequisites

- [ ] The relay stops sending `input:error` with no reason at all — `RelayServer.ts:774` (`'agent offline'`). It is the one producer that knows the answer for certain and the only in-repo site that omits the field. Filed as #492; making `reason` required is impossible while it exists.
- [ ] #485 — the dashboard reads `reason`.
- [ ] #457 — mcp-server acts on `reason` rather than only including it in the thrown message.

[...]

PR #495, opened 2026-08-08
fix(relay): carry a reason on input:error, and stop blaming an agent for a session it never had

### LINKED ISSUE #492: relay: the one producer that knows for certain sends input:error with no reason — and its prose is wrong for half the cases it covers
## What

`RelayServer.ts:774` answers a terminal input the relay cannot dispatch:

```ts
this.sendTo(ws, { type: 'input:error', sessionId: msg.sessionId!, message: 'agent offline' })
```

No `reason`. Every other production send site carries one after #490, and this is the only in-repo producer that omits it — while being the producer that knows the answer with the most certainty. The agents infer a reason from their own state; the relay is looking directly at a socket that is not `OPEN`.

A consumer therefore has to fall back to the documented rule "an unfamiliar or absent reason is treated as `channel-unavailable`" for a message whose reason is not unclear at all.

## The prose is also wrong for one of the two branches

The `else if` fires in two unrelated situations, and the single string only describes one:

- `session.agentSocket.readyState !== OPEN` — the agent really is offline.
- `session === undefined` — the session was evicted or never existed. **The agent may be perfectly healthy.** Telling the caller "agent offline" here is a wrong diagnosis, and it is the same class of inaccuracy #482 / #488 / #490 were about, just one layer up.

Both map to `reason: 'channel-unavailable'` — the caller's move is the same (re-join or reconnect) — so the machine field is correct for both while the prose is correct for one. That is a small, concrete illustration of why `reason` is the field consumers should read (#491).

## Fix

- Add `reason: 'channel-unavailable'` to the send.
- Split the prose so each branch says what actually happened, or make it accurate for both. Do not fold the distinction into a second wire reason: the consumer action does not differ, and the reason set is derived from what a consumer must do differently.

## Why it was skipped in #490

Recorded as T7 in that PR's adversarial review. #490 deliberately kept the relay diff at zero — the change spanned four packages already and the documented fallback covered this case. That reasoning holds for why it was not in #490 and does not survive as a permanent state, because it is a prerequisite for #491 (`reason` cannot become required while a producer omits it).

## Scope

One package (`relay`), one send site plus tests. No protocol change — `reason` is already optional on the type, so adding it here is additive.

jo-duchan:
Closes #492.

The relay answers a terminal input it cannot dispatch — `input:touch:end`, `input:pinch:end`, `input:key`, `input:button` — so an MCP or browser caller fails immediately instead of waiting out its own timeout, which its fallback reports as success. That reply carried no `reason`, which made the relay the last producer of `input:error` without one, and the producer whose answer was least in doubt: an agent infers a reason from its own state, while the relay reads the socket directly.

## Two situations, one of them misdiagnosed

The `else if` covers a session held with a socket that is no longer open, and no such session at all — evicted after the reconnect grace, or never valid. Only the first is the agent's fault; in the second the agent can be perfectly healthy, and `agent offline` sent the reader after the wrong problem.

`device:boot` already told that pair apart, twelve lines earlier in the same file, with the two strings now used here and a comment giving the reason. So this is a missed application rather than a new judgement.

## Both branches keep `channel-unavailable`

Not split into two wire reasons. The set is derived from what a consumer must *do* differently and a reconnect or re-join answers both — the machine field was right for both cases while the prose was wrong for one, which is the concrete argument for branching on `reason` and displaying `message`.

That collapse had a consequence worth following through, and the review caught that I hadn't: the dashboard's copy for this reason told the reader to check the agent on the Mac, which is a wild goose chase for a session the relay simply no longer has — the exact misdirection this PR sets out to remove, reintroduced by the consumer. The action now names no cause and leaves it to the message shown beside it.

## What actually changes downstream

`mcp-server` is what visibly changes; it puts the reason in the error it raises. The dashboard's output does not — its unknown-reason rule already resolved to this same reason, and while the agent is away it suppresses the notice entirely. The real value is that **absence of the field now means an agent older than it and nothing else**, which is what #491 needs before the field can be required.

No protocol change: `reason` has been optional on this message since it was introduced.

## Review

`.work/reviews/fix__relay-input-error-reason.md` — one pre-PR channel against `96b8ce8`, handed my five mutations and asked what they missed. Five findings, all verified:

- **My commit claimed the docs were updated and three of four places were not**, including `DeviceViewer.tsx` — shipped source, and the comment that points a reader at the changed file. The paragraph I *did* edit is a near-copy of it, so I updated the mirror and left the original.
- **The non-terminal guard covered one of seven types.** Widening `TERMINAL_INPUT_TYPES` to include `input:touch:start` or `input:type` survived the whole suite, and both are harmful — the first answers twice per tap, the second answers a message whose waiters key on the `input:type-*` pair and would ignore it. All seven are now listed.
- **My `AGENTS.md` bullet over-generalised**: "no caller waits on a move" is true of moves and false of `input:type`, which is also excluded and *does* have a 15s waiter in two packages. The comment I replaced was correctly scoped.
- **The changeset overstated the benefit** (see above).
- A test comment named `session:terminated` where the awaited-past message is `session:agent-away`.

8 mutations, 2 of which survived initially. One lesson recorded: removing `reason` failed 5 tests, not 6 — the both-branches test compared two reasons for equality and `undefined === undefined` passes. It now names the value.

The plan carried a diagnostic test case — "the existing `clearState.test.ts` expectation of `agent offline` must keep passing; if it breaks, my branch judgement is wrong" — and it passed, because the reconnect grace holds the session. The reviewer confirmed that independently.

Two follow-ups recommended in the record: `input:type` burns its full deadline against an offline agent and needs an `input:type-error` from the relay rather than membership in this set; and a missing `sessionId` drops the required field from this reply, which predates this diff and affects `open-url` and `device:boot` equally.

Full suite 2187 passed + 1 skip (relay 570 → 583); typecheck, lint, build and a11y-lens clean.

## Summary by CodeRabbit

* **New Features**
  * Terminal input errors now include a `channel-unavailable` reason.
  * Error messages distinguish between unavailable agents and missing sessions.
  * Dashboard guidance now recommends rejoining the session when the channel is unavailable.
* **Documentation**
  * Updated guidance to explain error reasons and compatibility with older agents.
* **Tests**
  * Added coverage for terminal input failures, missing sessions, unavailable agents, and non-terminal inputs.

### DIFF
--- a/.changeset/relay-input-error-reason.md
+++ b/.changeset/relay-input-error-reason.md
@@ -10,9 +10,11 @@ own timeout, which its fallback would report as success. That reply carried no `
 relay the last producer of `input:error` without one, and the one whose answer was least in doubt: an
 agent infers a reason from its own state, while the relay is looking straight at the socket.
 
-It now sends `reason: 'channel-unavailable'`. Consumers that switch on the reason — the dashboard's
-per-reason notice, and `mcp-server`, which includes it in the error it raises — no longer have to fall
-back to the "unknown reason" rule for a message whose reason was never unclear.
+It now sends `reason: 'channel-unavailable'`. What visibly changes today is `mcp-server`, which puts the
+reason in the error it raises. The dashboard's output does not change — its unknown-reason rule already
+resolves to this same reason, and while the agent is away it suppresses the notice entirely — so the
+value here is that **absence of the field now means an agent older than it and nothing else**, which is
+what makes it possible to require the field later.
 
 The prose was also wrong for half the cases it covered. Two situations reach that reply: the session
 is held with a socket that is no longer open, or there is no such session — evicted after the reconnect
@@ -24,6 +26,11 @@ Both keep `channel-unavailable` rather than splitting into two wire reasons. The
 what a consumer must do differently, and a reconnect or a re-join answers both — the machine field was
 right for both cases while the prose was wrong for one, which is exactly why consumers should branch on
 `reason` and display `message`.
+
+One consequence of that collapse had to be followed through: the dashboard's copy for this reason told
+the reader to check the agent on the Mac, which is a wild goose chase for a session the relay simply no
+longer has. It now says to rejoin the session and leaves the specific cause to the message shown beside
+it.
 
 No protocol change: `reason` has been optional on this message since it was introduced, so this is
 additive. Absence of the field now means an agent older than it, and nothing else.
--- a/.changeset/relay-input-error-reason.md
+++ b/.changeset/relay-input-error-reason.md
@@ -10,11 +10,12 @@ own timeout, which its fallback would report as success. That reply carried no `
 relay the last producer of `input:error` without one, and the one whose answer was least in doubt: an
 agent infers a reason from its own state, while the relay is looking straight at the socket.
 
-It now sends `reason: 'channel-unavailable'`. What visibly changes today is `mcp-server`, which puts the
-reason in the error it raises. The dashboard's output does not change — its unknown-reason rule already
-resolves to this same reason, and while the agent is away it suppresses the notice entirely — so the
-value here is that **absence of the field now means an agent older than it and nothing else**, which is
-what makes it possible to require the field later.
+It now sends `reason: 'channel-unavailable'`. Two things change visibly: `mcp-server` puts the reason in
+the error it raises, and the dashboard's guidance for this reason is reworded (below). What does *not*
+change is the dashboard's branching — its unknown-reason rule already resolved to this same reason, and
+while the agent is away it suppresses the notice entirely. So the field itself buys something narrower
+and more durable: **absence now means an agent older than it and nothing else**, which is what makes it
+possible to require the field later.
 
 The prose was also wrong for half the cases it covered. Two situations reach that reply: the session
 is held with a socket that is no longer open, or there is no such session — evicted after the reconnect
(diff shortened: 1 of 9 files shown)


jo-duchan:
Correct, and it is a contradiction inside my own changeset — fixed in c3f8ab4.

The paragraph on downstream effects said the dashboard's output does not change, while a later paragraph in the same file describes changing its guidance copy. The first was true when I wrote it and stopped being true when I fixed the review finding about that copy: I added the new paragraph and did not revisit the old one.

Worth naming plainly, because **this is the second instance of the same failure in this PR.** The pre-PR review's first finding was that I claimed the docs were updated where they named the relay reasonless and left three of four places behind — including `DeviceViewer.tsx`, whose comment is a near-copy of the `AGENTS.md` paragraph I did edit. Same shape here: one of two places stating the same fact. So the pattern is not "I forgot a file" but that when I re-read a diff I check whether each sentence is true, and not whether a claim appears more than once. Recorded in the review record rather than just fixed.

The paragraph now separates three things that were being conflated:

- **Visible changes**: `mcp-server` puts the reason in the error it raises, and the dashboard's guidance for `channel-unavailable` is reworded.
- **Not changed**: the dashboard's branching — its unknown-reason rule already resolved to this same reason, and it suppresses the notice entirely while the agent is away.
- **What the field itself buys**: absence now means an agent older than it and nothing else, which is what makes requiring it possible (#491).

Changeset-only change, so the gates are unaffected: full suite 2187 passed + 1 skip, typecheck and lint clean. Review record updated at `.work/reviews/fix__relay-input-error-reason.md`.

Issue #499, opened 2026-08-08
protocol: an input ack carries no correlation id, so a late ack is consumed by the next input and reports the wrong outcome


jo-duchan:
Found while reviewing #457. Pre-existing; that work makes it newly consequential rather than causing it.

## What happens

`input:done` / `input:error` identify only the session, never the input they answer. So every consumer correlates by "the next ack for this session", and that is wrong whenever an ack is late:

```ts
// packages/mcp-server/src/client.ts — the predicate has nothing else to match on
(m['type'] === 'input:done' || m['type'] === 'input:error') && m['sessionId'] === sessionId
```

Sequence:

1. Input **N** is sent. Its ack is slow — `ackInput` awaits a `simctl list` / `adb` device verify on the first input after a boot or reconnect (`IOSAgent.ts:746-752`, `AndroidAgent.ts:917-923`), on the Mac the relay gates at 80% CPU.
2. N's waiter hits its 2s window and gives up.
3. Input **N+1** is sent and arms a new waiter.
4. N's ack arrives and satisfies N+1's waiter.

From there the session is permanently off by one: every input reports the *previous* input's outcome. An `input:error` gets attributed to an input that was fine, and — the direction that matters — an input that was never answered is reported as having landed, because it consumed an older `input:done`.

Demonstrated during the #457 review: with the client's strict gate armed, a single bare `input:done` delivered mid-window let the *unanswered* input resolve as success in under 1500ms.

`flow-runner` correlates the same way (`RelayClient.ts`), so it has the same exposure.

## Why the obvious workarounds do not work

- **Draining stale acks before arming a waiter** has nothing to drain: an unmatched message is dropped in `dispatch`, so a late ack that arrives between inputs is already gone. The damaging case is the one that arrives *after* the next waiter exists, which is indistinguishable at that point from the answer it is impersonating.
- **A shorter or longer ack window** moves the boundary without removing it.
- **Counting outstanding inputs** cannot help either: the client knows how many it sent, but not which ack belongs to which, and it must not block on the difference — an agent that genuinely drops one ack would then stall every input after it.

## The fix has to be on the wire

Give the ack something to correlate on. Two shapes:

- **Echo the input** — the ack carries the `type` it answers (`input:touch:end`, `input:key`, …). Cheap and additive, and enough to catch cross-type misattribution, but not enough for two taps in a row.
[...]

PR #500, opened 2026-08-08
fix(mcp-server): stop reporting an unacknowledged input as success

### LINKED ISSUE #457: mcp-server: a dropped input is reported as success — the optimistic ack fallback outlived its reason

jo-duchan:
Closes #457.

`awaitInputAck` waited 2s for `input:done` / `input:error` and, on timeout, **returned successfully**. So `tap`, `swipe` and every other terminal gesture reported success when nothing had acked, and the model driving the session was told the tap landed and moved on. The fallback existed for agents predating the ack protocol and outlived them — both agents ack every terminal input (#484, #488), and since #495 every producer sends a reason too, so with both ends current silence means something is actually wrong.

## Three decisions, all of which the review changed

**Silence is "could not confirm", not "dropped".** `ackInput` awaits a device verify on the first input after a boot or reconnect, on the same Mac the relay gates at 80% CPU — so an ack past the window can belong to an input that *did* land. Calling that a drop invites a retry, and a retry of a landed input duplicates it. The error says the input may have landed and to check device state instead of repeating.

**Whether silence is fatal is decided by what the session has already done**, not by a negotiated flag. `ackedSessions` records any session that has answered with `input:done`; only those are judged strictly. An agent that never acks is never judged, which needs nothing on the wire. The ledger is written in `dispatch` rather than at the waiter, so an ack that missed its own window still counts — that is the case worth learning from.

**Nothing is retried here.** Each reason carries advice for the caller instead, including a warning that `no-gesture` may already have applied part of the input.

## What the reviews took out

**The design review discarded both halves of the first plan and cut it from four packages to one.**

A per-reason auto-retry was unsafe on both platforms. `no-gesture` means either "nothing reached the device" or "the opening frames landed and only the last was refused" — `TouchHelper`'s ownership record is deliberately conditional on delivery, so both are real — and the wire reason is identical, so a client that retried would sometimes apply a 300ms drag twice with nobody able to see it had. On Android the same reason can follow a `touchDown` that reached the guest, with the retry then going out over a different backend. `TapflowClient` also drives `run_flow`, where a retry makes deterministic replay non-deterministic.

A capability flag (`'input-ack'`) was the wrong mechanism. My sticky-capabilities argument was contradicted by the protocol's own comment — *"a restarted agent may advertise a different set (an upgrade is the usual reason to restart one)"* — and a **rollback** restart would have made the client report every input as dropped to an agent that cannot ack, inviting exactly the double-input the retry cell was rejected for. It was also a version stamp wearing capability clothing: both agents would have had to advertise it, a static check would police the pairing, and one release cycle later it would be permanently inert and unremovable because consumers key on its absence. And it would have pre-decided the fork #491 is explicitly open on, while listing this work as its prerequisite.

**The pre-PR review then found a blocker, and it was mine from earlier the same day.** The ledger counted `input:error` as evidence too — but the relay originates `input:error` to this very socket for a terminal input it cannot dispatch, which is the reply #495 had just given a reason to. One agent-offline blip, or a single input for a session this client never joined, marked a session as acking when its agent may never have answered anything; every later input then threw "this session has acknowledged input before", which was false, at the exact population the optimistic path protects. Counting only `input:done` fixes it and is simpler: nothing in the relay originates one.

It also caught that "bounded to one input" was false (the gap is any session that never gets an answer), that the advice suite pinned uniqueness but not direction so two bodies could be swapped, and that one test sat at 80% of vitest's unconfigured 5s default.

One finding I did not accept as stated: the reviewer read the `dispatch-failed` advice as inverting the protocol, citing `dispatchTo`'s comment that the deadline made a retry safe. That comment had itself been refuted by the analysis in `android-agent/AGENTS.md` written during the PR that reviewed it (#488) — the deadline cancels our call, not a request the emulator applied. So the advice stands and **the stale comment is corrected here**, since it is what misleads a reader. Which side the protocol table should take is #491's.

## Review

`.work/reviews/fix__mcp-input-ack-truthfulness.md` — design round (2 channels, discarded the design) and a pre-PR round against `2dd32ee`. 10 mutations, 2 of them regression guards for what the reviews removed.

## Follow-up filed

**#499** — an input ack carries no correlation id, so an ack arriving after its own input timed out is consumed by the next input's waiter and reports the previous input's outcome. It predates this change and this change makes it consequential; it needs a field on the wire, and `clipboard:*` already sets the precedent with `requestId`. Documented next to the gate rather than papered over, because it is the one way that gate can still be defeated.

Full suite 2200 passed + 1 skip (mcp-server 41 → 63); typecheck, lint, build and a11y-lens clean.

## Summary by CodeRabbit

* **New Features**
  * Input errors now include actionable guidance, including when to retry, reconnect, reboot, or investigate.
  * Input acknowledgement status is tracked independently for each session.
  * Previously acknowledged sessions report unanswered or disconnected inputs as unconfirmed.

* **Bug Fixes**
  * Prevented automatic retries when input delivery is uncertain.
  * Improved handling of input errors, late acknowledgements, and acknowledgement timeouts.

* **Documentation**
  * Clarified timeout behavior and safe retry guidance for input dispatches.

### DIFF
--- a/.changeset/mcp-input-ack-truthfulness.md
+++ b/.changeset/mcp-input-ack-truthfulness.md
@@ -20,10 +20,20 @@ a drop invites a retry, and a retry of a landed input duplicates it. The error t
 input may have landed and to check device state rather than repeat it.
 
 **Whether silence is fatal is decided by what the session has already done.** A session that has
-answered an input before is judged strictly; one that never has keeps the optimistic path, because an
-agent that does not ack at all is exactly what the fallback was for. This degrades in the safe
-direction and needs nothing on the wire. A session's first input stays optimistic, which is the one
-case where silence genuinely does mean "an agent that does not ack".
+answered an input with `input:done` is judged strictly; one that never has keeps the optimistic path,
+because an agent that does not ack at all is exactly what the fallback was for. This degrades in the safe
+direction and needs nothing on the wire.
+
+`input:done` specifically, not any ack: the relay originates `input:error` to the client for a terminal
+input it cannot dispatch, so counting those would let one agent-offline blip mark a session as acking
+when its agent may never have answered anything — and then report every later input as unconfirmed on
+evidence the agent did not produce. A session that has never had an answer keeps the optimistic path
+indefinitely, so #457 is unchanged for an agent whose acks never arrive; what this buys is that once a
+session answers, silence after that is reported.
+
+One gap stays open and is documented rather than papered over: an ack carries no correlation id, so an
+ack arriving after its own input timed out is consumed by the next input's waiter. That needs a field on
+the wire (#499).
 
 **An `input:error` now carries advice, not just prose.** Each reason maps to what the caller should do
 — boot the device, reconnect, send the same input again in a moment, or never retry this one. The
--- a/.changeset/mcp-input-ack-truthfulness.md
+++ b/.changeset/mcp-input-ack-truthfulness.md
@@ -1,0 +1,37 @@
+---
+'@tapflowio/mcp-server': patch
+---
+
+fix(mcp-server): stop reporting an unacknowledged input as success
+
+`awaitInputAck` waited 2s for `input:done` / `input:error` and, on timeout, **returned successfully**.
+So `tap`, `swipe` and every other terminal gesture reported success when nothing had acked, and the
+model driving the session was told the tap landed and moved on.
+
+The fallback existed for agents predating the input-ack protocol and outlived them: both agents ack
+every terminal input unconditionally (#484, #488), and since #495 every producer also sends a
+machine-readable reason. When both ends are current, silence means something is actually wrong — which
+is exactly the case the fallback converted into a success.
+
+**Silence is now reported as "could not confirm", not as a drop.** That distinction is load-bearing:
+`ackInput` awaits a device verify on the first input after a boot or reconnect, on the same Mac the
+relay gates at 80% CPU, so an ack past the window can belong to an input that *did* land. Calling that
+a drop invites a retry, and a retry of a landed input duplicates it. The error tells the caller the
+input may have landed and to check device state rather than repeat it.
+
+**Whether silence is fatal is decided by what the session has already done.** A session that has
+answered an input before is judged strictly; one that never has keeps the optimistic path, because an
+agent that does not ack at all is exactly what the fallback was for. This degrades in the safe
+direction and needs nothing on the wire. A session's first input stays optimistic, which is the one
+case where silence genuinely does mean "an agent that does not ack".
+
+**An `input:error` now carries advice, not just prose.** Each reason maps to what the caller should do
+— boot the device, reconnect, send the same input again in a moment, or never retry this one. The
+`no-gesture` advice warns that part of the input may already have been applied, because that reason
+covers both "nothing landed" and "the opening frames landed and only the last was refused".
+
+Nothing is retried automatically. That was the first design and was discarded after review: the wire
+cannot distinguish those two `no-gesture` cases, so a client that retried would sometimes apply a drag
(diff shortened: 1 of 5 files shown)


jo-duchan:
Correct, and the comment was mine — fixed in bc67d35.

`// A WebSocket close ... means the input was not dispatched` is false for the reason you give: every caller sends its input **before** awaiting the ack — `tap` sends both frames, `swipe` all ten — so by the time the socket closes the input has left this process and the relay may already have forwarded it. Worth naming plainly: **that is the same false certainty this whole PR exists to remove**, and I rewrote that line while preserving the claim inside it.

I took the substance and split the wording rather than sharing one sentence:

- **timeout** → cites the ledger, because the ledger is what decided it
- **close** → cites the close

One message for both would attach the wrong reason to one of them, and the reason is what a model acts on.

And I applied it **independent of the ledger**, as you suggested: a close says nothing about whether the agent acks — only that we stopped being able to hear it — so a session that has never answered an input still gets the truth rather than an optimistic success. That is the one point where the proposed diff was better than gating it on `strict`, and it is in.

Two notes on what I did differently:

- The `'WebSocket closed'` string is matched explicitly rather than treating every non-timeout error as unconfirmed. An error this method cannot account for should still propagate — silently relabelling an unknown failure as "may have landed" would be a new guess in place of the one just removed.
- The existing test pinned the *message* of the false claim (`toThrow(/closed/i)`), so it would have passed a wrong fix. It now pins the substance: unconfirmed, may have landed, do not repeat. One mutation to confirm — rethrowing the close raw fails 2 tests.

mcp-server 63 → 64. Full suite 2201 passed + 1 skip; typecheck, lint, build and a11y-lens clean. Review record updated at `.work/reviews/fix__mcp-input-ack-truthfulness.md`.

PR #501, opened 2026-08-08
refactor(protocol): declare wire payload types once, and add a check that keeps them there

jo-duchan:
First step of a multi-PR program to clean up the cross-package wire contract. The program plan lives at `.work/WIRE-CONTRACT-PLAN.md`; this is its **L2**.

## Why this program exists

Six "silent lie" bugs on the input path were fixed in one day (#482 → #500). Each fix was correct and each revealed the next gap, because the defect was in the **shape** of the contract rather than in any one message. Measured: 4 of the 6 were protocol-shaped, and the other 2 were harder to see because the wire could not express what the agent knew.

Six structural defects came out of that. This PR closes one of them.

## What was wrong

Every wire payload type was declared separately in three to five packages, and they had drifted **in both directions**:

- `@tapflowio/protocol` was missing the `payload` field on `clipboard:error` that both agents send and the viewer reads.
- The dashboard's `session:chrome` declaration was missing three fields its own `DeviceViewer` reads.

Nothing checked either side. This is the exact situation `@tapflowio/protocol` was created to end — it had been half done, and the other half drifted back apart.

Several shapes were also the same thing under different names, which is how the duplication survived a name-based search: `DeviceSummary` was `AgentDevice` in the dashboard and `DeviceInfo` in `mcp-server` and `flow-runner` — and that last name collided with the relay's own `DeviceInfo`, which is a *different* shape. `DeviceReport` was `RawDevice` in the relay. `Point` was `NormPoint` in the dashboard.

## What changed

Protocol owns them; everyone else imports. `agent-core`, `flow-runner`, the relay and the dashboard re-export the names they published, so no consumer of those packages changes — including third-party agents built on `AgentRegistry.register()`. `export type` rather than a bare re-export, so `agent-core` does not gain a runtime import of protocol's `dist`.

**The comments mattered more than the types.** `ChromeData` was described accurately only in `ios-agent`, which produces it: the viewer lays out against an *expanded* composite canvas — the device frame grown by the button margins — and protocol's copy said `compositeWidth` was "full PDF width including devicePadding", a different quantity. Deleting the producer's declaration would have deleted the only correct description of the coordinate space three viewers compute against. Protocol carries it now, with both spaces named explicitly.

## The check

`scripts/__tests__/protocolPayloadTypes.test.mjs` fails if any workspace member re-declares a protocol payload shape. It matches on **field sets, not names**: the inventory that planned this work grepped names and missed two of five copies of one shape, so a name-scoped check would have passed with all five standing and would be bypassed by a rename.

It earned itself repeatedly — five copies nobody had inventoried (the relay's `SessionInfo` and `RawDevice`, `agent-core`'s `Point`, the dashboard's `NormPoint`, plus two inline `{ deviceName, osVersion }` literals in the relay). One legitimate false positive, `UIElementFrame`, is allowlisted with its reason: same four fields as `ChromeRect`, but it is an accessibility-tree element's frame and merging them would couple the UI-tree schema to bezel artwork.

## Review

`.work/reviews/refactor__wire-l2-payload-types.md` — a design round (1 blocker: the clipboard requirement did not hold together, and my inventory was wrong) and a pre-PR round (2 majors), against `80a4063`.

The pre-PR round is worth summarising, because both findings were mine:

- **I deferred half of a fix that did not need deferring.** The design review correctly said typing the clipboard *bridge* needs types L3 creates. I deferred the whole thing, including the one-line field addition on protocol's own member that needs nothing — leaving the moved type referenced nowhere while the commit message named that drift as the motivation. Fixed, and the commit amended so the claim matches the code.
- **The check could be bypassed, and not hypothetically.** The first version required the `interface` keyword and a closing brace at line start. The reviewer planted five duplicates and it reported one: `type X = { … }` aliases, a single-line body, and the `playground` workspace member all walked past — and two inline literals in the relay were passing at that moment. A check whose whole value is being un-bypassable is worse than none when it is bypassable, because its presence reassures. All five forms now fail; verified by replanting them.

7 mutations, 4 of which survived the first version.

## Scope

No behaviour change: declarations, imports and comments. Every package's test count is unchanged — scripts 211 → 213 for the new check's two tests. typecheck, lint, build and a11y-lens clean.

Deliberately left for L3: removing `useClipboardBridge`'s casts, which needs `clipboard:data` / `clipboard:write-done` in protocol — those are agent→browser forwards, a category protocol does not model yet, and that is L3's subject.

## Summary by CodeRabbit

* **New Features**
  * Added shared protocol types for device, session, Chrome, Android, coordinate, resource, and clipboard data.
  * Clipboard error messages can include additional status details.
* **Documentation**
  * Clarified Chrome coordinate-space and expanded-canvas behavior.
* **Bug Fixes**
  * Improved consistency of device and payload data across connected components.
* **Tests**
  * Added validation to detect duplicate payload type declarations.

### DIFF
--- a/.changeset/wire-l2-payload-types.md
+++ b/.changeset/wire-l2-payload-types.md
@@ -1,0 +1,43 @@
+---
+'@tapflowio/protocol': patch
+'@tapflowio/agent-core': patch
+'@tapflowio/flow-runner': patch
+'@tapflowio/mcp-server': patch
+'@tapflowio/relay': patch
+---
+
+refactor(protocol): wire payload types are declared once, and a check keeps them that way
+
+Every wire payload type was declared separately in three to five packages, and they had drifted **in
+both directions**: `@tapflowio/protocol` was missing the `payload` field on `clipboard:error` that both
+agents send and the viewer reads, while the dashboard's `session:chrome` declaration was missing three
+fields its own `DeviceViewer` reads. Nothing checked either side, which is the situation
+`@tapflowio/protocol` was created to end — it had been half done.
+
+Protocol now owns them and everyone else imports: `ChromeData`, `ChromeButton`, `ChromeRect`,
+`AndroidButton`, `AgentResources`, `SessionInfo`, `Point`, and `ClipboardErrorPayload` (which moves out
+of `agent-core`, since neither the dashboard nor `mcp-server` can reach that package). `agent-core`,
+`flow-runner` and the relay re-export the names they published, so no consumer of those packages
+changes — including third-party agents built on `AgentRegistry.register()`.
+
+Three shapes were also the same thing under different names, which is how the duplication survived:
+protocol's `DeviceSummary` was `AgentDevice` in the dashboard and `DeviceInfo` in `mcp-server` and
+`flow-runner`, and that last name collided with the relay's own `DeviceInfo`, which is a *different*
+shape. `DeviceSummary` is now the one name; `flow-runner` keeps exporting `DeviceInfo` as an alias
+because the CLI imports it.
+
+**The comments moved with the types, and that mattered more than the types.** `ChromeData`'s fields
+were described accurately only in `ios-agent`, which produces them: the viewer lays out against an
+*expanded* composite canvas (the device frame grown by the button margins), and protocol's copy said
+`compositeWidth` was "full PDF width including devicePadding" — a different quantity. Deleting the
+producer's declaration would have deleted the only correct description of the coordinate space three
+viewers compute against. Protocol's fields now carry it, with the two spaces named explicitly.
+
+A new check (`scripts/__tests__/protocolPayloadTypes.test.mjs`) fails if any package re-declares a
+protocol payload shape. It matches on **field sets, not names** — the inventory that planned this work
+grepped for names and missed two of the five copies of one shape for exactly the reason above, so a
+name-scoped check would have passed with all five standing and would be bypassed by a rename. It found
+two more copies during implementation: the relay's own `SessionInfo` and `agent-core`'s `Point`.
+
+No behaviour change: type declarations, imports and comments only. Every package's test count is
+unchanged.
--- a/packages/agent-core/package.json
+++ b/packages/agent-core/package.json
@@ -56,6 +56,7 @@
     "lint": "eslint src"
   },
   "dependencies": {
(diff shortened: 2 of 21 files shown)


jo-duchan:
Both correct, and both fixed in cc6ad23. The first one is the more instructive.

## The body capture was truncating, and my own fix for a related finding hid it

`[\s\S]*?` to the first `}` stopped at the first **nested object literal**. So `ChromeData` was guarded as its first 6 fields instead of 11, and `ChromeButton` as 5 instead of 13 — meaning a real duplicate of either would not have matched the truncated key at all. The body is found by counting braces now.

What makes this worth naming: **the pre-PR round had already flagged the floor assertion here, and my fix for it gave false confidence.** I replaced `guarded.size >= 10` with a by-name assertion — "`ChromeData` must be in the guard" — and it passed, because the truncated 6-field key was mapped to that name just fine. A name being present says nothing about its field set being complete. I traded a weak assertion for a differently weak one and moved on.

So the fix is not just brace counting; it is an assertion that would have caught this: the field counts of the shapes that *have* nested literals, since those are exactly what a lazy match gets wrong. `ChromeData` 11, `ChromeButton` 13, `DeviceSummary` 7, `SessionInfo` 4.

## Coverage could shrink without failing

`catch { continue }` swallowed every error from the walk, so a permission error or a broken symlink would skip a whole workspace member while the test reported green. Only `ENOTDIR` / `ENOENT` is tolerated now.

And the roots come from `pnpm-workspace.yaml` rather than a hardcoded `playground`, so a new top-level member is scanned without an edit here — plus an assertion that the roots really were scanned. A member silently dropped from the walk is the other way this check goes quietly useless, and the comment above it claimed "every workspace member" while the code named one.

## Verification

Three mutations, all failing: a full 11-field `ChromeData` copy (**which the previous version would have missed** — that is the finding, reproduced), reverting to the truncating capture, and dropping a member from the scan.

scripts 213 → 214. Full suite 2203 passed + 1 skip; typecheck, lint and build clean.

One thing worth reporting rather than swallowing: the first full-suite run after this had `relay/src/__tests__/static-compression.test.ts > serves the brotli sibling` fail. Judged a one-off under parallel load and unrelated — that test builds its own `mkdtempSync` directory so it does not touch anything here, this PR's relay change is type-only, and it passed both in isolation (583) and on a second full run. Recorded in the review record as a first data point in case it recurs, not filed as an issue.

PR #503, opened 2026-08-08
refactor(protocol): complete the browser-inbound surface by direction

jo-duchan:
Layer 3 of the wire-contract program (`.work/WIRE-CONTRACT-PLAN.md`). Follows #501 (L2).

## What was wrong

A browser receives **28** message types. `@tapflowio/protocol` declared 17 — one of which does not go to a browser at all — and the dashboard declared its own hand-copy of 24 in `lib/types.ts`.

The twelve an agent sends were in **neither**. The relay forwards them with `JSON.stringify(msg)`, so nothing on its typed send path mentions them and no compiler could notice they were missing.

The two copies had drifted in four places and nothing reported any of it:

| | dashboard | protocol |
|---|---|---|
| `app:install-error`, `app:launch-error`, `device:boot-error` | `sessionId?` | `sessionId` |
| `session:joined.capabilities` | optional | required |
| `app:install-done`, `app:launch-done`, `device:booting`, `device:shutdown-done` | no `sessionId` | absent entirely |
| `app:clear-state-*`, `input:type-*` | absent | absent |

## What this does

- **`AgentToBrowser`** — the twelve forward-only messages. Shapes derived from both agents' send literals, since routing gives only the label.
- **`RelayOrAgentToBrowser`** — the ten with *both* producers, declared **once** and referenced by both directions rather than written into each. Two copies of one message drift; that is what this PR is cleaning up.
- **`BrowserInbound`** — what a consumer uses. The dashboard's `RelayMessage` is deleted; 12 files import this instead. The name also collided with the relay's own `RelayMessage`, which is its *inbound* type.
- **`RelayToStream`** — `stream:registered` goes to an agent's stream socket, not a browser.

`sessionId` stays required everywhere it already was. `session:chrome` / `session:deviceInfo` / `device:ready` get `sessionId?` because the two producers genuinely differ: both agents stamp it, the relay's replay to a re-joining viewer does not.

## Two silent drops surfaced

The clipboard bridge declared its own `{ type: string; payload?: unknown }`, wide enough that a `clipboard:write-done` answering a read parsed as "no text" and cancelled the claim **saying nothing**; likewise `clipboard:data` answering a write. Both report now, with tests.

Five fixtures were sending a `device:booting` with **no `sessionId`** — a message the wire does not produce, and one that bypasses the viewer's session scoping. They compiled because the injection helper ended in `as never`.

## Checks

`scripts/__tests__/browserInboundRouting.test.mjs` — 10 assertions. Routing ↔ union in both directions, the field set of all 28 members pinned with optionality, a parser-blind-spot assertion, no consumer re-declaring a wire-message union, and `DeviceViewer`'s clipboard routing list derived from both sides and compared. Plus 5 new `@ts-expect-error` assertions in `typeAssertions.ts`, because `RelayOutbound` must reject a message only an agent produces.

`AgentToBrowser` needs a script check at all because the compiler cannot see a forwarded message. `RelayToBrowser` does not — `sendTo(socket, msg: RelayOutbound)` already holds it.

## Not in scope

`msg.sessionId!` itself (#444), relay forward narrowing (L4), named interfaces (L1).

## Review

Three rounds, **10 findings, all dispositioned** — one skipped with its reasoning recorded and handed to L4.

- **Design**, before implementing: two channels over the plan. Both returned a blocker and both named the same thing — the draft had carried forward a measurement taken before #501 landed. The plan was rewritten.
- **Code**: two independent channels (protocol change → escalation), on shapes/partition and on dashboard runtime behaviour. 7 findings. **Two were mutations that passed all 296 tests**, and one corrected the *reason* behind a conclusion that survived — which is what sharpened #444.
- **CodeRabbit**: 3 findings. Two caught 25 fixture casts that bypassed the union — the rule this PR itself adds to `packages/dashboard/AGENTS.md`, left violated in the files it touched.

Fixing those casts cost `capabilities ?? []` its only coverage, because every `session:joined` fixture had been omitting the field. That is now one deliberately non-conforming fixture in a test whose subject is the defense. It also invalidated an item the code reviewer had *cleared* — a reminder that a cleared list ages with the diff.

13 mutations, all caught. Dispositions and grounds: `.work/reviews/refactor__wire-l3-browser-inbound.md`.

One finding corrected a reason rather than a conclusion, and it matters for #444: there is **no guard** rejecting a message with no `sessionId`, so a client sending `{"type":"input:touch:end"}` bare reaches `sessions.get(undefined)` and the relay answers with the key dropped. Required `sessionId` is therefore a specification the relay does not yet meet — which makes #444 "send `{type:'error'}` instead", not "delete the `!`".

## Summary by CodeRabbit

- **New Features**
  - Added consistent handling for messages received from relay and agent connections.
  - Added session-aware forwarding and stream registration support.
  - Improved clipboard response reporting for successful, timed-out, and failed operations.

- **Bug Fixes**
  - Prevented malformed or unexpected clipboard responses from being treated as empty data.
  - Improved routing and session filtering for inbound device messages.

- **Tests**
  - Expanded coverage for message routing, clipboard errors, and session-scoped events.


jo-duchan:
Thanks — two of the three were right, and the first one caught a rule this PR itself added.

## Fixture casts (both comments) — fixed

Twenty-five `as BrowserInbound` casts removed across five files. `session:joined` was being injected without `sessionId` or `capabilities` in eleven places, and `session:chrome` went through a two-step `as unknown as`. This PR added exactly that rule to `packages/dashboard/AGENTS.md` and left it violated in the files it touched, so citing the coding guidelines was correct.

One cast stays, in `DeviceViewer.inputError.test.tsx:50`: posting a `reason` this build does not know about **is** the subject of that test, so disagreeing with the union is deliberate there. It is documented on the helper.

**Removing them had a consequence worth recording.** `setAgentCapabilities(msg.capabilities ?? [])` lost its only coverage — every `session:joined` fixture used to omit the field, so the fallback ran everywhere and was guarded nowhere. With the fixtures typed correctly, deleting the `?? []` leaves all 297 tests green while a non-conforming agent throws on `agentCapabilities.includes('clipboard')` and blanks the tab. So one fixture is now *deliberately* non-conforming, in a test whose subject is that defense, with a comment saying why. Verified by mutation: the mutation survived before the new test and fails after it.

## `AgentToBrowser` accepting the three shared messages without `sessionId` — skipped, and the doc comment was wrong

The observation is correct, and it found a real error next to it: the `AgentToBrowser` doc comment claimed *every* member carries a required `sessionId`. Three inherited from `RelayOrAgentToBrowser` are `sessionId?`. That comment is rewritten.

The remedy is deferred to L4 (the layer that types the relay's own sends), for three reasons:

1. **Declaring the three twice reintroduces what this PR removes.** Two copies of one message drift — that is the finding this PR is built on: four divergences between `protocol` and the dashboard's hand-copy, none of them reported by anything.
2. **The mapped alternative breaks a dependency.** `Omit<T,'sessionId'> & { sessionId: string }` turns those members into intersections, so `Extract<BrowserInbound, { type: … }>` stops resolving to a single member. `useClipboardBridge` reads its three replies without a cast precisely because each `Extract` yields one declaration.
3. **No consumer can read the stricter claim today.** `BrowserInbound` merges both directions and a union merge collapses to the looser field, and nothing consumes `AgentToBrowser` on its own — the relay forwards with `JSON.stringify(msg)`, untyped. The claim would be unenforced until the forward path is narrowed, which is L4's job.

The looseness is not incidental: the relay genuinely replays these three from its own cache without a `sessionId` (`RelayServer.ts:1079,1082,1089`) while both agents stamp one. `sessionId?` is what one declaration can honestly say about two producers that differ. L4 brings the replay up to the agents' shape and tightens the field in the same change; it is recorded there as a requirement rather than left implicit, and the reasoning is in the union's doc comment.

Partially satisfied now: a positive assertion pins the shape an agent actually sends (`agentStampsSession` in `typeAssertions.ts`). A negative assertion is not expressible while the shared declaration is optional — that is the same fact stated twice.

Full dispositions and the 13 mutations: `.work/reviews/refactor__wire-l3-browser-inbound.md`.

PR #505, opened 2026-08-09
refactor(protocol): give every wire message a name

jo-duchan:
Layer 1 of the wire-contract program (`.work/WIRE-CONTRACT-PLAN.md`). Follows #503 (L3).

## What was wrong

58 messages were anonymous members of a union, and that blocked two things.

**A single message could not be referred to.** Consumers reached for `Extract<Union, { type: 'x' }>` — one of them needing `extends { payload: infer P }` just to get at a payload.

**Shared structure had nowhere to live.** Eight session-scoped failures carry the same `{ sessionId, message }` contract with no place to say so, and the note explaining that contract floated above a union line.

## What this does

Each message is an `export interface`; the unions are unions of those names. `interface` rather than a `type` alias, because the alias cannot be `extends`ed and the intersection workaround stops `Extract<Union, { type: 'x' }>` resolving to one member — which `useClipboardBridge` relies on to read a reply without a cast.

`SessionError` is the base for the eight session-scoped failures. Not for DRY at two fields, but so the family relation is something a reader and a check can see. `error` does not inherit — it has no `sessionId`, which is that member's whole purpose.

**Naming forced a split anonymity had hidden.** `app:install` and `app:launch` travel in both directions with *different* shapes: the browser sends `buildId`, the relay resolves it into `payload: { filePath, bundleId }`. They become `AppInstallToRelay` / `AppInstallToAgent`. `device:shutdown` is identical both ways, so it is **one** interface both unions reference — which now states that the relay forwards it untouched.

## The names were generated, not typed

The conversion's real hazard is a copy-paste that leaves two interfaces holding each other's literal, and `AgentToBrowser` has seven members whose shape is identical apart from the literal. **No type-level check sees that** — a union is a set, so which name owns which literal is not part of the comparison. Measured: the swap produced zero errors from the nine equivalence assertions.

So names were derived from the literals mechanically, and the relation is asserted two ways because the cheap one is the same fact twice:

- a per-message binding in `typeAssertions.ts` — catches an author who edits one copy;
- a source check that the name is **derivable** from the literal — catches one who edits both, or who regenerates the binding table from the declarations it is meant to check. Measured: that regeneration left all 228 assertions green.

## Guards, all mutation-verified

`protocolMessageNames.test.mjs` (new) — bindings exist and agree, names derive, the eight failures declare their base, and `SessionError` still carries both fields. That last one matters: the signature table resolves `extends` and sorts, so moving `sessionId` out of the base into all eight subclasses is otherwise invisible.

Bodies are brace-counted. The first version matched a run of two-space lines, so **one blank line inside an interface hid a message and satisfied the count that was meant to prove coverage.**

`browserInboundRouting.test.mjs` — the parser resolves `| Name` to the declaring interface. Its pinned signatures now cover all five unions rather than the browser-inbound three: 58 bindings read like per-message coverage while pinning only a literal, and `ScreenshotRequest.requestId` / `DeviceBoot.sessionId` could both be made optional with everything green.

`protocolPayloadTypes.test.mjs` — messages are excluded from the duplicate-shape check (a message is identified by its literal, not its field set), and so is `SessionError` — but only because a *message* extends it. Excluding every base also excluded `DeviceReport`, which is a real payload.

## Two costs, recorded rather than reverted

Neither was visible to the equivalence proof, and both are in `packages/protocol/AGENTS.md`:

- **A named `interface` has no implicit index signature**, so a message is no longer assignable to `Record<string, unknown>`. Nothing breaks today — the three such sinks here only ever receive fresh object literals — and L4 is exactly the work that would hand one a typed value. The fix there is to type the sink, not to widen the message.
- **An `interface` can be reopened** by a consumer via `declare module`, which an anonymous union member could not.

## Review

Two design rounds and one code round, two channels each (protocol change → escalation): **11 findings, all dispositioned.**

The design round found a blocker I had missed — the equivalence net does not see names, which is the layer's entire output. The code round found three guards that passed while broken, each measured. One reviewer answer refuted my own analysis: I claimed moving a field into `SessionError` would be caught via the other seven signatures, and missed the reverse direction.

A conversion side effect is also fixed here: my generator's region parser swallowed four union JSDoc blocks, including 24 lines of rationale written for #503. Same failure as every other parser bug in this program — a region that does not know where it ends.

Record with dispositions and the 13 mutations: `.work/reviews/refactor__wire-l1-named-interfaces.md`.

## Summary by CodeRabbit

- **New Features**
  - Added clearly named protocol message types to improve integration clarity and tooling support.
  - Introduced shared session and generic error representations with more consistent message definitions.
  - Added validation to keep message names and protocol values aligned.

- **Bug Fixes**
  - Standardized clipboard error payload handling without changing runtime behavior.

- **Documentation**
  - Documented protocol message naming, inheritance, compatibility, and representation guidelines.

### DIFF
--- a/.changeset/named-message-interfaces.md
+++ b/.changeset/named-message-interfaces.md
@@ -1,0 +1,47 @@
+---
+'@tapflowio/protocol': patch
+'@tapflowio/relay': patch
+---
+
+refactor(protocol): give every wire message a name
+
+58 messages were anonymous members of a union, which meant two things could not be done. A single
+message could not be referred to — consumers reached for `Extract<Union, { type: 'x' }>`, one of them
+needing `extends { payload: infer P }` to get at a payload. And shared structure had nowhere to live:
+eight session-scoped failures carry the same `{ sessionId, message }` contract with no place to say so,
+and the comment explaining it floated above a union line.
+
+Each message is now an `export interface`, and the unions are unions of those names.
+
+- **`SessionError`** — the eight session-scoped failures extend it. Not for DRY (two fields) but so
+  that "this is a failure addressed to a session" is something a reader and a check can see, and the
+  contract note has a home. `error` does not inherit: it has no `sessionId`, which is the whole point
+  of that member.
+- **Direction suffixes where a literal means two things.** `app:install` and `app:launch` travel in
+  both directions with *different* shapes — the browser sends `buildId`, the relay resolves it and
+  sends `payload: { filePath, bundleId }`. Naming forced the split: `AppInstallToRelay` /
+  `AppInstallToAgent`. `device:shutdown` is identical in both directions, so it is one interface both
+  unions reference, which now says out loud that the relay forwards it untouched.
+- `GenericError`, because `Error` would shadow the global.
+
+**Names were derived from the `type` literals mechanically, not typed by hand.** The conversion's real
+hazard is a copy-paste that leaves two interfaces holding each other's literal, and `AgentToBrowser`
+has seven members whose shape is identical apart from the literal. A type-level equivalence check
+cannot see it — a union is a set, so which name owns which literal is not part of the comparison, and
+the routing check compares membership, so the literal set is unchanged. Measured: the swap produced
+zero errors from the nine equivalence assertions.
+
+So `typeAssertions.ts` carries one binding per message (`_InputDone: InputDone['type'] = 'input:done'`),
+and `scripts/__tests__/protocolMessageNames.test.mjs` asserts every message has one — plus that the
+eight failures declare `extends SessionError`, which no type can state about itself because every
+object with `{ sessionId, message }` is assignable to it.
+
+The conversion itself was proven with `Equals<Union, UnionOld>` against verbatim snapshots and then
+deleted along with them; that net was for the conversion window, not for keeping.
+
+**Two properties were given up, neither visible to that proof.** A named `interface` has no implicit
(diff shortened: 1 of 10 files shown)


jo-duchan:
Both correct, and the first one was a guard that would have gone quiet exactly when it mattered.

## The bridge derivation truncated at the first newline — fixed

`ClipboardBridgeMessage` is one line today, so the regex looked safe. It is not: a fourth member wraps the declaration, the capture takes only the first line, and `wanted` silently loses the trailing members. If the author also forgot to add them to `DeviceViewer`'s routing condition, `routed` is short by the same ones and **the two sets match** — which is precisely the omission this guard exists to catch.

The `expect(wanted.length).toBe(3)` I had added does not help either: the truncation lands on exactly the old count.

Now reads to the next blank line, the way `unionBody` does. Reproduced before and after — a fourth member added to the bridge and not to the viewer passes the old regex and fails the new one.

## Comment counts aligned on 58

`56-of-57` (messageNames:17), "57 came from L1's conversion" (:88), the pin at 58, and `57-of-58` (typeAssertions:104). The prose is what the next author reads before touching the pin, so it should not have to be reconciled from four places.

---

Verified: `tsc -b` 0; `dashboard`, `mcp-server`, `test-utils` `tsc --noEmit` 0 each (the root `tsc -b` references only eight projects and covers none of those three — a correction from an earlier review round); assertions project 0; scripts 231; dashboard 297; all package suites pass.

Dispositions and the 13 mutations: `.work/reviews/refactor__wire-l1-named-interfaces.md`.

PR #506, opened 2026-08-09
fix(dashboard): answer the three ways a session:start can fail, and declare every message's disposition

jo-duchan:
Layer 6 of the wire-contract program (`.work/WIRE-CONTRACT-PLAN.md`). Follows #505 (L1).

## The bug

The relay sends `error` with three wordings. The dashboard handled two.

`Session busy` reached the viewer and **did nothing** — so two testers opening the same device left the second tab waiting on a `session:joined` that cannot arrive. Nothing reported it, because from the outside `error` *was* a handled type: the viewer branched on the free-prose `message`.

## The fix, and what review found on top of it

`error` now carries a closed `reason`, the same split #491 gave `input:error`. **Required** here — that one's producer set is open by design (a third-party platform registers via `AgentRegistry.register()`), while this one has a single producer, the relay. So `sendTo` enforces it, and the compiler walked the change: protocol → three relay sites → the viewer's prose comparison → the copy record.

Two independent review channels then found that **handling a case is not the same as ending it**, in three more places:

| | |
|---|---|
| **`agent-resources-exhausted` only toasted** | The relay `return`s after sending it, so nothing follows — the tab sat on "Starting device…" indefinitely. Making `reason` required stopped a case being *unhandled* without making the handled cases *end*. |
| **`session-busy` does not mean "someone else"** | The relay answers it whenever the session's browser socket still reads OPEN, and the commonest cause is the tester's **own** previous socket: a sleeping laptop reconnects in 2s while the relay takes up to a heartbeat to notice the old one died. My first wording asserted a colleague who does not exist — in exactly the case `DeviceViewer`'s own comment calls routine (#439). |
| **`SessionList` was dropping `error` too** | `handleShutdown` sends `session:start` on its own socket; a refused join is answered there, and `device:shutdown-done` — the only message that clears `shutting` — never comes. The badge read "Shutting down..." permanently and the gate on it hid both buttons. |

The relay also checks occupancy **before** the resource gate (with both true, the tester got the less actionable answer and the real reason went unreported), skips that check for a socket re-joining the session it already holds, and no longer folds `join()`'s two exceptions into a `catch {}` that discarded the error.

## The layer

Browser-inbound is 28 types; the dashboard handled 22 and dropped 6 — and the three reasons for dropping (handled elsewhere, deliberately ignored, nobody wrote it) were **indistinguishable in code**, because all three look like an absent branch.

`lib/inboundDisposition.ts` states one per message under `satisfies Record<BrowserInbound['type'], Disposition>`, so a message added to the wire **breaks the file** until someone picks a category. Measured: `TS1360` at the table.

Deriving the *reachable* subset and obliging only that was designed first and rejected by two channels for different reasons, both recorded in the table: `send()` is shared by four sockets that each open their own WebSocket, and **a reply does not go to whoever asked** — the relay forwards to whichever socket holds the session now, so "we never send `input:type`" is true per session, not per socket.

## The guard had to be fixed twice, and that is the interesting part

`scripts/__tests__/inboundDisposition.test.mjs` covers what a type cannot state. Its first version used `src.includes("'<type>'")` and **certified a false claim**: `error` was listed as handled in `SessionList`, which had no branch for it, and three unrelated `'error'` strings made it green while the message really was being dropped. Deleting a genuine branch and leaving the literal in a comment passed too.

So the table **created one of the three indistinguishable states it exists to remove**, and the check signed it off. Strengthening the predicate to an actual `.type` comparison immediately surfaced three more false `at` claims.

The parser also sliced to end-of-file rather than the table's closing line, so a sibling map below it could restore an entry count that a skipped key had lowered — the third time in this program a region parser has not known where its region ended.

## Not in scope

The session-ownership defects behind `session-busy` — evicting a holder the heartbeat has already lost, and `handleShutdown` leaving the first session permanently `busy: true`. Those change who owns a session and do not belong in a typing layer; filed separately with the review's evidence.

## Review

Two design rounds and one code round, two channels each: **11 findings, all dispositioned.** The first design was rejected whole — its rationale is kept in the plan, because a context reset will propose it again. Record and the 11 mutations: `.work/reviews/refactor__wire-l6-consumption-obligation.md`.

## Summary by CodeRabbit

* **New Features**
  * Added structured reasons for session-start failures, including busy sessions and exhausted resources.
  * Added distinct notifications for busy, overloaded, and unavailable sessions.
  * Added clearer error notifications when session shutdown is refused.

* **Bug Fixes**
  * Improved session rejoining and resource-check behavior.
  * Prevented shutdown requests before joining or while another shutdown is in progress.
  * Improved handling of unexpected session-start failures.

* **Documentation**
  * Clarified device information compatibility and usage.

### DIFF
--- a/.changeset/inbound-disposition-and-error-reason.md
+++ b/.changeset/inbound-disposition-and-error-reason.md
@@ -1,0 +1,60 @@
+---
+'@tapflowio/protocol': patch
+'@tapflowio/relay': patch
+---
+
+fix(dashboard): tell the second tester the device is in use, and make an unhandled message a compile error
+
+`Session busy` reached the viewer and did nothing. The relay sends it when another browser socket already
+holds the session, so **two testers opening the same device** — the likeliest collision in a product whose
+premise is that the whole team opens a browser — left the second tab waiting on a `session:joined` that
+cannot arrive.
+
+Nothing reported it, because from the outside `error` *was* a handled type. The viewer branched on the
+free-prose `message` and covered two of the three wordings the relay sends.
+
+`error` now carries a closed `reason` — the same split #491 gave `input:error`: `message` stays prose the
+producer owns, the machine field is closed. **Required here, unlike `InputErrorReason`**, because that
+one's producer set is open by design (a third-party platform registers through `AgentRegistry.register()`
+and may predate the field) while this one has a single producer: the relay, at three sites. So `sendTo`
+enforces it. The viewer switches on `reason` exhaustively, and a fourth reason is a compile error instead
+of another silent case.
+
+`busy-elsewhere` and `mac-overloaded` are dashboard-local stop reasons rather than new
+`SessionTerminatedReason` values, because in both cases the session is **alive** — widening the protocol
+vocabulary would let `session:terminated` carry a reason it can never mean. The copy record is keyed on the
+union, so it forced both wordings.
+
+The busy wording deliberately does **not** name another person. The relay answers `session-busy` whenever
+the session's browser socket still reads OPEN, and the commonest cause is the tester's *own* previous
+socket: a sleeping laptop reconnects in 2s while the relay takes up to a heartbeat (30s) to notice the old
+one died. "Someone else is testing this device" would be false in exactly the case the viewer's own comment
+calls routine. Relatedly, the relay now checks occupancy *before* the resource gate — with both true, the
+tester was told to pick a different Mac while the real reason went unreported — and skips the check for a
+socket re-joining the session it already holds, which is what `SessionList` does before a shutdown.
+
+`agent-resources-exhausted` also gained an exit. It only toasted, and the relay `return`s after sending it,
+so the tab sat on "Starting device…" indefinitely: making `reason` required stopped a case from being
+unhandled without making the handled cases *end*.
+
+**`SessionList` was dropping `error` too**, and the same shape of bug: `handleShutdown` sends
+`session:start` on its own socket, a refused join is answered there, and `device:shutdown-done` — the only
+message that clears `shutting` — never arrives. The badge read "Shutting down..." permanently and the gate
+on it hid both buttons, so the row went inert.
+
+## The layer this belongs to
+
+Browser-inbound is 28 message types; the dashboard handled 22 and dropped 6. The three reasons for
+dropping — handled elsewhere, deliberately ignored, nobody wrote it — were **indistinguishable in code**,
+since all three look like an absent branch. `lib/inboundDisposition.ts` states one per message under
+`satisfies Record<BrowserInbound['type'], Disposition>`, so a message added to the wire breaks the file
+until someone picks a category. Measured: adding one produces `TS1360` at the table.
+
+Deriving the *reachable* subset and obliging only that was designed and discarded, and it is the obvious
+idea, so both reasons are recorded in the file: `send()` is shared by four sockets that each open their
+own WebSocket, and a reply does not go to whoever asked — the relay forwards to whichever socket holds the
+session now, so "we never send `input:type`" is true per session, not per socket.
+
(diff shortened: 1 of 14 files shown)


jo-duchan:
Both confirmed, and the first one made a message this PR added into a false statement.

## The shutdown went through after a refused join — fixed

`handleShutdown` sent `session:start` and `device:shutdown` back to back, and you are right that nothing stops the second frame: **no browser→agent forward checks `session.browserSocket === ws`**. The relay forwards on the strength of the session existing and the agent socket being open. So a refused join was followed by a shutdown that landed — another tester's device went down while this list reported that nothing had happened, which is exactly what the toast added in this PR claims.

That is the same mistake as an earlier finding in this branch, where the busy copy asserted "someone else is testing this device" without verifying it. Twice now the fix stated something it had not established.

The shutdown now waits for `session:joined`. Verified in both directions: a refused join sends no `device:shutdown`, an accepted one does.

**The general defect is not fixed here.** Ownership is unchecked on every session-scoped forward, so any socket that knows a session id can drive it. That changes who owns a session rather than how messages are typed, and it joins the ownership work this PR already deferred — evicting a holder the heartbeat has lost, and `handleShutdown` leaving the first session permanently `busy: true`. Filing those together with your evidence rather than reaching into the forward path from a typing layer.

## Concurrent shutdowns — fixed, and the comment was the tell

`setShutting({})` was justified in a comment saying one shutdown is in flight at a time. The code did not enforce that; it was an assumption written as a fact. Now a second shutdown while one is unanswered is refused, and the refusal handler clears only the row it was waiting on — waiting for `session:joined` supplies the request identity, so both of your findings close with one change.

## One thing your review surfaced indirectly

Adding a `session:joined` branch to `SessionList` left `inboundDisposition`'s entry naming only two files, and **the check did not notice** — `at` was never obliged to be complete. That is a hole an earlier reviewer had flagged as harmless, and it went stale inside the same PR that introduced the table. There is now a reverse assertion: every module's `.type` comparisons must be named in the table's entry for that type.

---

Verified: `tsc -b` 0; dashboard/mcp-server/test-utils `tsc --noEmit` 0 each; assertions project 0; scripts 237; dashboard 302; relay 584; all package suites pass. Lint clean — it caught a TDZ on `send` inside its own `useRelay` callback, now using the `sendRef` idiom `DeviceViewer` already uses.

Dispositions and the 13 mutations: `.work/reviews/refactor__wire-l6-consumption-obligation.md`.

jo-duchan:
Both fixed. The first one is the third time on this branch that a fix asserted something it had not established, which is worth naming.

## `session:joined` is now matched to the request, not counted

The handler accepted **any** join as the answer to its pending shutdown. Only this list sends `session:start` from here, and a second one is refused while the first is unanswered, so a mismatch is unreachable today — but "unreachable today" is precisely what the two previous versions of this handler asserted and got wrong:

- the busy copy claimed "someone else is testing this device", which is false whenever the holder is the tester's own pre-reconnect socket;
- the refusal toast claimed the device "was not shut down", while the shutdown frame went through anyway.

Three for three. The comparison is one line, so it goes in regardless of reachability. A mismatched join now leaves the request pending, and there is a test for both halves: `s2` joining fires nothing, and the subsequent `s1` still works.

## The matcher no longer depends on formatting

`msg.type==="session:joined"` would have gone unseen, and the table could have drifted incomplete without failing — the same class as every other parser this branch has had to harden. Both directions now tolerate either quote style and any spacing around the operator.

It still has to reject a bare literal, since that was the original defect: verified by deleting a branch and leaving only a double-quoted literal in a comment, which fails.

---

Verified: `tsc -b` 0; scripts 237; dashboard 303; lint clean.

Dispositions and the 15 mutations: `.work/reviews/refactor__wire-l6-consumption-obligation.md`.

Issue #508, opened 2026-08-09
android-agent: a JPEG screenshot request returns PNG bytes labelled image/jpeg


jo-duchan:
An Android screenshot requested as JPEG comes back as **PNG bytes labelled `image/jpeg`**.

Found while deriving the shape of `screenshot:done` for the wire-contract program: the field's *type* is identical on both platforms, which is why nothing has caught it. The disagreement is in what the value means.

## The three steps

1. **Android always produces PNG.** `AdbWrapper.screenshot()` runs `screencap -p` and takes no format argument (`packages/android-agent/src/AdbWrapper.ts:149-151`).
2. **The reply echoes the request anyway.** `AndroidAgent.ts:1227-1231` sends `format: format ?? 'png'` — the value the caller asked for, not the value it produced.
3. **The relay turns that into a Content-Type.** `RelayServer.ts:1216`: `const contentType = fmt === 'jpeg' ? 'image/jpeg' : 'image/png'`.

iOS does honour the request: `this.simctl.screenshot(shotState.deviceId, format ?? 'png')` (`IOSAgent.ts:1043`).

So the field is a **request echo on Android and a result description on iOS**, and only the second is what the relay reads it as.

## Why the type system cannot see it

`format: 'png' | 'jpeg'` is correct on both sides. A consumer, a static check, and a compiler all agree. The only way to find it is to read what each agent does with the value — which is what shape derivation forced.

## Impact

Mild but real. Most decoders sniff magic bytes and will render the PNG regardless of the header, so a viewer usually shows the right image. What breaks is anything that trusts the label: a strict image pipeline, a stored artefact whose extension is derived from Content-Type, or an MCP client that hands the bytes to a library configured by MIME type.

Not a data-loss bug and not worth blocking on — but it is a lie on the wire, and the fix is small.

## Options

- **Report what was produced.** `format: 'png'` unconditionally on Android, since that is the truth today. One line, and the relay's Content-Type becomes correct.
- **Honour the request.** Convert with the emulator's tooling or re-encode; more work, and the PNG→JPEG conversion cost lands on every screenshot.

The first is the honest minimum. The second is a feature, and only worth it if a caller actually needs JPEG — the MCP `screenshot` tool is the one to ask.

## Not the same as the ceiling on `data`

Unrelated to payload size. This is purely the mismatch between the format asked for, the format produced, and the format announced.

Issue #507, opened 2026-08-09
relay: session-scoped commands are authorised by the session existing, not by who owns it


jo-duchan:
Three defects with one root: **the relay decides what a browser socket may do from the session's existence, never from who owns it.** Found while reviewing #506, which fixed the messages these produce but deliberately left the ownership semantics alone — that PR was a typing layer, and this changes who may drive a session.

## 1. No session-scoped forward checks ownership

`RelayServer.ts` forwards `device:boot`, `device:shutdown`, `open-url`, `app:install`, `app:launch`, `app:clear-state`, every `input:*` and both clipboard requests on the strength of two things: the session exists, and its agent socket is open. None of them checks `session.browserSocket === ws`.

So **any authenticated browser socket that knows a session id can drive that session** — shut the device down, install a build, inject input — while another tester is using it.

Reachable without guessing an id: `session:start` is refused for a session someone else holds (`session-busy`), and the refusal does not stop the next frame. #506 fixed the one in-repo caller that did this (`SessionList` now waits for `session:joined`), but the relay still accepts it from anyone.

## 2. A holder the heartbeat has already lost is never evicted

`SessionManager.join()` refuses when `session.browserSocket?.readyState === WebSocket.OPEN`. The relay's own liveness signal is separate: `HEARTBEAT_MS` is 30s and a socket is terminated only after it misses a pong, so a browser that vanished — sleeping laptop, Wi-Fi blip, both documented as routine in `DeviceViewer`'s comments (#439) — leaves `readyState` OPEN on the relay for **30–60 seconds**.

During that window the same tester reconnects (`useRelay` retries after 2s), re-sends `session:start`, and is refused for holding the session themselves. Frames keep going to the dead socket in the meantime, and if the tester does not get back before `clearBrowser`'s idle timer fires the relay asks the agent to shut the device down.

`wsAlive` already tracks what is needed to tell a lost holder from a live one.

## 3. `handleShutdown` can leave a session permanently `busy`

`SessionList.handleShutdown` sends `session:start` before the shutdown, so the list socket becomes the session's `browserSocket`. `join()` overwrites `browserSocketIndex` for the new session (`SessionManager.ts:149-151`) but the previous session keeps pointing at that same socket.

[...]

PR #509, opened 2026-08-09
refactor(protocol): declare the agent→relay direction, and type every agent send

jo-duchan:
Layer 4a of the wire-contract program (`.work/WIRE-CONTRACT-PLAN.md`). Follows #506 (L6).

## The gap

An agent's outbound literal was **the one part of the wire contract no compiler could see.** The relay forwards replies with `JSON.stringify(msg)`, so nothing typed ever re-creates them, and each agent handed its literal straight to `ws.send`.

#489 (an agent answering nobody) and #490 (a missing `reason`) both came out of that, and `scripts/__tests__/inputErrorReason.test.mjs` exists because a script had to stand in for a compiler.

Seven message types were declared nowhere at all — `agent:register`, `agent:resources`, `screenshot:done`, `screenshot:error`, `stream:register`, `ui:tree:response`, `ui:tree:error`. **With this, every direction on the wire is declared.**

## What the compiler found the moment it could see

| | |
|---|---|
| **30 sends passed an optional `sessionId` into a required field** | The dispatcher took `{ sessionId?: string }` while every type it dispatches is session-scoped. Now required, with the check at the socket boundary — and **25 `msg.sessionId!` assertions went away as a consequence**, plus the now-unreachable `if (!sessionId) return` in each `ackNoSession`. That is the payoff #444 is after on the relay side, arriving here first. |
| **`requestId` was optional in the clipboard cases, required in the screenshot and ui:tree ones** | Same wire guarantee, same file, two spellings. |
| **`respond` took `object`** | So all three clipboard replies were unchecked — the last unchecked send path in that code. |

## Two decisions worth reading

**`AgentControlOutbound = AgentToRelay | AgentToBrowser`** — `StreamToRelay` is deliberately outside the bound the agents' helpers take. A draft merged all three and handed back the exact hazard the direction split exists to avoid: `case 'stream:register'` calls `setStreamSocket(session.id, ws)` with no role gate, so a control socket that could type-check that message could take over the session's video path. Both review channels found this independently.

**`sendOn(ws, msg)` takes the socket as an argument.** Its call sites sit behind an entry guard and are compiler-proven non-null; reading `this.ws` inside the helper — or asserting it with `!` — would make those guards optional to the compiler and turn deleting one from a compile error into a runtime `TypeError`. One of them has a test whose subject is the window it covers.

`screenshot:error` and `ui:tree:error` are the first `*-error` messages that do **not** extend `SessionError`: they are request-scoped, resolved by `requestId` alone. `SessionError` is for a failure a *session* is waiting on, and naming these draws the boundary rather than widening it.

## The guard had to be rewritten four times

`agentSendTyped.test.mjs` matches **serialization**, not any spelling of `.send`. Three drafts keyed on `this.ws` were bypassed in review by renaming the socket — and `streamWs.send(JSON.stringify(…))` is not hypothetical: `streamWs` is in scope in `startBinaryStream`, and the relay dispatches a text frame from a stream socket through the same agent cases. A commented-out copy of the canonical helper also satisfied the positive assertion while the real one took `msg: object`, leaving all 66 of that agent's sends unchecked with the check green.

## Two mistakes the record keeps

Inserting the helpers before rewriting the call sites let the rewrite hit the helper bodies, so `sendMsg` called itself. **The suite hung rather than failed**, and it was found by establishing that the baseline passed rather than by assuming the tests were at fault.

And the boundary guard's test took three attempts. The first asserted a non-terminal input does nothing — true with or without the guard, which is the Potemkin shape the plan itself warned about. The second used a terminal input and still passed, because **the relay's own forward gate already blocks that frame** — the same fact the required declaration rests on. So the guard is for a client that reaches an agent without going through the relay, and the test now drives it that way.

## Not in scope

Relay forward narrowing (L4b), the untyped `Record<string, unknown>` sinks (L4c), and enforcement for third-party platform agents — a source scan cannot reach another package, and the structural answer is a shared helper in `agent-core`, which `DeviceAgent` being an interface rules out for now.

## Review

Two design rounds and one code round, two channels each: **14 findings, all dispositioned.** The design round found **both of the plan's premises false**, including a correction I had already made myself that was also wrong. Record and the 10 mutations: `.work/reviews/refactor__wire-l4a-agent-producer.md`.

Byproducts filed separately: #507 (session ownership) and #508 (Android returns PNG bytes labelled `image/jpeg`).

## Summary by CodeRabbit

* **New Features**
  * Added shared protocol support for agent, stream, screenshot, UI-tree, clipboard, and resource messages.
  * Improved validation for session and request identifiers.
  * Added typed handling for control and streaming communication.
  * Standardized UI element types across integrations.

* **Bug Fixes**
  * Invalid or sessionless relay messages are now ignored safely.

* **Tests**
  * Added coverage for message validation and communication serialization.

### DIFF
--- a/.changeset/agent-producer-typing.md
+++ b/.changeset/agent-producer-typing.md
@@ -15,7 +15,8 @@ and `scripts/__tests__/inputErrorReason.test.mjs` exists because a script had to
 
 Seven message types were declared nowhere: `agent:register`, `agent:resources`, `screenshot:done`,
 `screenshot:error`, `stream:register`, `ui:tree:response`, `ui:tree:error` — the last undeclared direction.
-They are now `AgentToRelay` and `StreamToRelay`, and `AgentOutbound` is what a typed send takes.
+They are now `AgentToRelay` and `StreamToRelay`, and `AgentControlOutbound` is what the agents' typed send
+helpers take.
 
 `stream:register` is its own direction rather than part of `AgentToRelay`, mirroring `RelayToStream`: the
 relay assigns the role `'stream'` from it, not `'agent'`, so folding it in would make the union's name
@@ -25,8 +26,8 @@ inbound is narrowed by role.
 ## Two helpers, and why not one
 
 ```ts
-private sendMsg(msg: AgentOutbound): void { this.ws?.send(JSON.stringify(msg)) }
-private sendOn(ws: WebSocket, msg: AgentOutbound): void { ws.send(JSON.stringify(msg)) }
+private sendMsg(msg: AgentControlOutbound): void { this.ws?.send(JSON.stringify(msg)) }
+private sendOn(ws: WebSocket, msg: AgentControlOutbound): void { ws.send(JSON.stringify(msg)) }
 ```
 
 `sendOn` takes the socket **as an argument**. Seven call sites sit behind an entry guard (`if (!this.ws)
--- a/.changeset/agent-producer-typing.md
+++ b/.changeset/agent-producer-typing.md
@@ -1,0 +1,80 @@
+---
+'@tapflowio/protocol': patch
+'@tapflowio/agent-core': patch
+'@tapflowio/ios-agent': patch
+'@tapflowio/android-agent': patch
+'@tapflowio/mcp-server': patch
+---
+
+refactor(protocol): declare the agent→relay direction, and type every agent send
+
+An agent's outbound literal was the one part of the wire contract no compiler could see. The relay forwards
+replies with `JSON.stringify(msg)`, so nothing typed ever re-creates them, and each agent handed its literal
+straight to `ws.send`. #489 (an agent answering nobody) and #490 (a missing `reason`) both came out of that,
+and `scripts/__tests__/inputErrorReason.test.mjs` exists because a script had to stand in for a compiler.
+
+Seven message types were declared nowhere: `agent:register`, `agent:resources`, `screenshot:done`,
+`screenshot:error`, `stream:register`, `ui:tree:response`, `ui:tree:error` — the last undeclared direction.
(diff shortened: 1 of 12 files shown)


jo-duchan:
Both correct, and both are the same failure: prose that stopped describing the code.

## Fixed

The helpers take `AgentControlOutbound`. The union was split *during* the review round — precisely so a control socket cannot type-check `stream:register` and take over a session's video path — and the changeset and `packages/protocol/AGENTS.md` were both still naming the merged one. `AGENTS.md` also still said `screenshot:error` and `ui:tree:error` carry an optional `sessionId`, which the same round made required.

So the two documents were describing the design this PR *rejected*, right next to the reasoning for rejecting it. Both now state the implemented contract, including why the exclusion exists and why the required field is not optional.

## Worth naming, because it is the third time on this branch

Each time, a fix changed the code and left the reason written beside it false:

- `AgentToBrowser`'s doc claimed every member carried a required `sessionId` after three of them became optional (caught in the #503 round);
- the `sessionId?` justification here was falsified by a **later hunk of the very change that relied on it** — the dispatchers became `sessionId: string`, so "the agents pass through an optional one" stopped being true mid-commit;
- and now the rename and the optionality.

All three were caught by review rather than by me. It is a different failure from the one #504 documents — that rule is about a *reviewer's* clearance expiring when the findings are fixed; this is about *my own recorded reasoning* going false when I fix them. The rule that would have caught all three: **after applying review fixes, re-read the prose each fix touched — the justification is part of the change, not documentation of it.**

Happy to add that to `contributing/adversarial-review.md` as a follow-up if you want it; leaving it out of this PR since it is a rules change and this one is a typing layer.

---

Verified: `tsc -b` 0; scripts 243; lint clean.

Dispositions: `.work/reviews/refactor__wire-l4a-agent-producer.md`.

Issue #512, opened 2026-08-09
mcp-server, flow-runner: the relay's replies are read as prose, so the machine fields are discarded


jo-duchan:
`mcp-server` and `flow-runner` both read what the relay sends them as `Record<string, unknown>`. One root, four consequences: **they discard the machine-readable fields the wire now carries, and read prose instead.**

Found while planning L4c (typing these two clients' *outbound*, #509's counterpart). Their outbound is the layer's subject; this is the other direction and needs its own reasoning, so it is filed rather than folded in.

**Not the same as #444.** That one is the relay trusting unvalidated JSON *from* browsers and agents — a robustness problem whose fix is a runtime validator. This is two clients not reading fields that exist and are required, in messages produced by code that is now typed end to end. Different direction, different packages, different fix.

## 1. An `error` match on a field that cannot exist — wrong session reported as failed

`packages/mcp-server/src/client.ts:185` and `packages/flow-runner/src/RelayClient.ts:123`:

```ts
(m['type'] === 'error' && (m['sessionId'] === undefined || m['sessionId'] === sessionId))
```

`GenericError` has **no `sessionId` field at all** — that is the whole point of the member: it is what the relay sends when a failure cannot be attributed to a session.

So the second branch is dead and can never be true, and the surviving `m['sessionId'] === undefined` matches **any** `error` regardless of which session caused it. With one client joined to several sessions — an LLM connecting two devices is the ordinary case for `TapflowClient` — an `error` raised by session B is caught by session A's in-flight `connectDevice` waiter, and A's join is reported as failed.

Narrowing the inbound would have made this a compile error (`TS2339: Property 'sessionId' does not exist on type 'GenericError'`).

**This one needs a decision, not just a fix.** An `error` with no `sessionId` genuinely cannot be correlated, so a waiter cannot know it is theirs. The options are the same shape as #445: fail the wait on any uncorrelated `error`, ignore them and let the deadline speak, or have the relay stop sending uncorrelated ones (it already has `session-not-found` / `session-busy` / `agent-resources-exhausted` — see #506, which required a machine field on this exact message).

## 2. `GenericError.reason` is ignored and the prose is read instead

`client.ts:188` and `RelayClient.ts:127`:

```ts
if (msg['type'] === 'error') throw new Error((msg['message'] as string) ?? 'Connect failed')
```

[...]

PR #513, opened 2026-08-09
fix(protocol): app:clear-state must carry a bundleId, and every browser-role send is checked against the contract

jo-duchan:
Layer L4c of the wire-contract program. The remaining untyped browser-role producer is typed, one message whose declaration was looser than every producer *and* every consumer is tightened, and the static check that defends both is anchored where its sibling learned to anchor.

Follows #501 (L2), #505 (L1), #503 (L3), #506 (L6), #509 (L4a). L4b was retired — narrowing a relay *forward* checks nothing, because no literal is constructed there.

## What changed

**`flow-runner`'s `RelayClient.send` takes `BrowserToRelay`** instead of `Record<string, unknown>`, so its 15 outbound literals are checked. `mcp-server`'s equivalent was already typed in `7637be3` — the plan claimed otherwise and was wrong; its source was this program's own stale handoff item, which cited a line that turned out to be `Waiter.resolve`. `RelayMsg` survives in both files as the **inbound** alias.

**Zero compile errors came out of it, and that is the honest result.** Every `sessionId` in that file arrives as a required `string` parameter, so the root cause that produced 30 errors when the agents were typed (#509) does not exist here. The value is regression defence.

**`AppClearState.payload` is now `{ bundleId: string }`** — the layer's one real defect. It was `payload?: { bundleId?: string }`: `mcp-server` and `flow-runner` are its only senders and both supply a string, and both agents answer `app:clear-state-error` with `'bundleId missing'` when it is absent. The declaration permitted a message whose only possible outcome was that error, and it compiled.

**`engine.ts` no longer reaches `driver.clearState` through `flow.appId!`.** `parseFlow` rejects a bare `clearState` with no flow-level `appId`, but `runFlow` is exported and does not parse — an embedder can hand it a `Flow` built by hand, and the assertion laundered that into `payload: {}` for the device to reject a network hop later.

**`scripts/__tests__/clientOutboundTyped.test.mjs`** holds all three browser-role producers to the contract.

## The check took two attempts, and that is most of this PR

Its first draft asserted the signature `private send(msg: BrowserToRelay)` while its own header claimed it did **not** match a spelling of `send(`. Review defeated it with the bypass the header named:

```ts
private sendRaw(msg: RelayMsg): void { this.ws?.send(JSON.stringify(msg)) }
this.sendRaw({ type: 'session:leaev', sessionId, whatever: 1 })
```

7/7 assertions passing, `tsc` clean, package suite green, misspelled type on the wire. `agentSendTyped.test.mjs` had already lost three drafts to a renamed socket; this one cited that lesson and then reproduced it.

It now anchors where the sibling does — `JSON.stringify` appears once per file, and the function enclosing it takes `BrowserToRelay` — which constrains the sink instead of its name. Three further gaps review found, all measured:

- An in-file `type BrowserToRelay = RelayMsg` passed every assertion. The import is asserted now and a local declaration refused.
- Nothing bound the union's *contents*. Appending `| Record<string, unknown>` to `BrowserToRelay` passed all 250 static assertions while making every literal in all three files accept anything.
- The file list was two hardcoded paths, silently excluding `dashboard/hooks/useRelay.ts` — already typed, and the busiest of the three (47 send sites against 32 for both clients). It is derived by inspection now, and walked from disk rather than `git ls-files`: a new sender file escapes while it is untracked, and "just added" is exactly when it is untracked. `agentSendTyped.test.mjs` has the same hole; it is recorded, not fixed here, because that check sweeps three agent packages and its mutation set would have to be re-run.

## Prose corrected where this work falsified it

The rule #510 added, applied to its own next PR — and it caught three places:

- `AppClearState`'s comment offered "the relay forwards without filling anything in" as *grounds*. The two sibling messages carry a `buildId` the relay resolves from the builds table, so blind forwarding is an absence of relay-side resolution, not a design. It also overstated what required buys: `bundleId: ''` type-checks, so the agents' guard stays live.
- `protocol/AGENTS.md` called `mcp-server` and `flow-runner` "the untyped senders" — true when written, falsified by the work that typed them — and listed three `Record<string, unknown>` sinks when two are now typed.
- The changeset's unconditional "not a breaking change" is scoped to in-repo producers. `AppClearState` is a published export and this package advertises `declare module` re-opening, so the next tightening should not cite this as precedent for "required fields are patches".

## Verification

`tsc -b` 0 · protocol assertions project 0 · lint 0 errors · manifests untouched.

| suite | |
|---|---|
| scripts (static checks) | **255** (250 before; the new file went 7 → 12) |
| flow-runner | **57** (56 + the engine guard) |
| mcp-server | 64 |
| dashboard | 303 |
| relay | 583 passed / 1 skipped |
| agent-core | 137 |

Eight mutations run and caught: second helper writing to `this.ws`; in-file `BrowserToRelay` shadow; `| Record<string, unknown>` appended to the union; new sender file while untracked; `send` renamed-but-loosened; `flow.appId!` restored; dashboard's send loosened; serialization hoisted out of the typed send via `const p = JSON.stringify(m)`.

## Adversarial review

Two channels — tightening correctness, and guard adequacy — plus an earlier design pass whose findings are the plan's "what the first draft got wrong" section. Record: `.work/reviews/refactor__wire-l4c-client-producer.md`.

**8 findings, 7 of them defects in my own fixes, 1 original bug** — the same ratio as L6 (15 and 1). The record also logs a third occurrence of this program's D19: `git checkout HEAD -- <file>` destroyed two uncommitted edits during the mutation round, because HEAD was the commit under review.

## Out of scope

- **Inbound narrowing for both clients** → #512, which also carries the four live defects it would catch. Ordering constraint (after #444) is recorded there.
- **Relay-side `bundleId` resolution for `app:clear-state`** — new feature, not contract alignment.
- **Relay inbound validation, the remaining `!`, replay `sessionId`** → #444.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clear-state operations now fail early with a clear error when no app ID is provided.
  * Outbound browser-to-relay messages are validated more reliably.
  * App clear-state messages now require a bundle ID, preventing incomplete requests.

* **Tests**
  * Added regression coverage for invalid clear-state steps and outbound message typing.

* **Documentation**
  * Clarified message validation behavior and updated related technical guidance.

### DIFF
--- a/.changeset/client-outbound-typing.md
+++ b/.changeset/client-outbound-typing.md
@@ -1,0 +1,34 @@
+---
+'@tapflowio/protocol': patch
+'@tapflowio/flow-runner': patch
+---
+
+fix(protocol): app:clear-state must carry a bundleId, and flow-runner's sends are checked against the contract
+
+`AppClearState` declared `payload?: { bundleId?: string }` — looser than every producer **and** every consumer.
+`mcp-server` and `flow-runner` are its only senders and both supply a string; the relay forwards without filling
+anything in; and both agents answer `app:clear-state-error` with `'bundleId missing'` when it is absent. So the
+declaration permitted a message whose only possible outcome was that error, and it compiled. Now
+`payload: { bundleId: string }`.
+
+Not a breaking change: the two in-repo senders already comply, and the dashboard does not send this message at
+all.
+
+`flow-runner`'s `RelayClient.send` takes `BrowserToRelay` instead of `Record<string, unknown>`, so its 15
+outbound literals are checked. `mcp-server`'s equivalent was already typed in `7637be3`; this finishes the pair.
+
+**Zero compile errors came out of it, and that is the honest result.** Every `sessionId` in that file arrives as
+a required `string` parameter, so the root cause that produced 30 errors when the agents were typed (#509) does
+not exist here. The value is regression defence, and `scripts/__tests__/clientOutboundTyped.test.mjs` is what
+carries it — reading the *signature* rather than any spelling of `send(`, because the agents' equivalent check
+was bypassed three times in review by renaming a socket.
+
+Two comments corrected while passing through, since a stale reason argues for the thing that was removed:
+
+- `test-utils`'s `SocketMessage` said its looseness accommodates each importer's richer view of the wire via
+  `waitForType<T extends SocketMessage>`. #505 broke that — named interfaces have no implicit index signature, so
+  no protocol type satisfies the constraint, and all 25 call sites already violate it invisibly because
+  `src/__tests__` is outside the package's tsconfig. The looseness now blocks the richer views rather than
+  accommodating them; it stays only because a narrowing nothing type-checks is decoration.
+- `mcp-server`'s AGENTS.md presented its loose *inbound* as a settled decision. It is a deferral, tracked in
+  #512 along with the live defect narrowing would catch.
--- a/.changeset/client-outbound-typing.md
+++ b/.changeset/client-outbound-typing.md
@@ -6,13 +6,20 @@
 fix(protocol): app:clear-state must carry a bundleId, and flow-runner's sends are checked against the contract
 
 `AppClearState` declared `payload?: { bundleId?: string }` — looser than every producer **and** every consumer.
-`mcp-server` and `flow-runner` are its only senders and both supply a string; the relay forwards without filling
-anything in; and both agents answer `app:clear-state-error` with `'bundleId missing'` when it is absent. So the
-declaration permitted a message whose only possible outcome was that error, and it compiled. Now
-`payload: { bundleId: string }`.
+`mcp-server` and `flow-runner` are its only senders and both supply a string, and both agents answer
+`app:clear-state-error` with `'bundleId missing'` when it is absent. So the declaration permitted a message whose
+only possible outcome was that error, and it compiled. Now `payload: { bundleId: string }`.
 
-Not a breaking change: the two in-repo senders already comply, and the dashboard does not send this message at
-all.
+**No in-repo producer changes** — both senders already comply and the dashboard does not send this message at
+all. It is a `patch` on that basis, not on a claim about every consumer: `AppClearState` is a published export and
+this package advertises `declare module` re-opening, so adding a required field is source-breaking for an
+out-of-repo producer that omits it. None is known to exist; the wording is scoped so the next tightening does not
+cite this as precedent for "required fields are patches".
+
+`flow-runner`'s `engine.ts` no longer reaches `driver.clearState` through `flow.appId!`. `parseFlow` rejects a
+bare `clearState` with no flow-level `appId`, but `runFlow` is exported and does not parse — an embedder can hand
+it a `Flow` built by hand, and the assertion laundered that into `payload: {}` for the device to reject a network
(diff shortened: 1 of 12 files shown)


jo-duchan:
All three were valid and all three are fixed in `17fe13c`. They were also all three defects in the fixes from the previous review round, not in the original change — worth stating, because it is the pattern this layer keeps producing.

**Scope-aware sink detection.** Confirmed exactly as described. The regex required the parameter to be named `msg`/`m`/`message` and to be the sole parameter, so `sendRaw(payload: RelayMsg, audit: boolean)` did not match and the backward scan walked past the function boundary to report `BrowserToRelay`. The previous round's mutation `blast(m: RelayMsg)` was caught only because `m` happened to be in the allowlist — so the difference between caught and not caught was what the parameter was called, in a file whose header disclaims exactly that brittleness.

It uses the TypeScript AST now rather than another hand-rolled brace walker. Two of this file's three bypasses in one round came from text heuristics reporting a fallback instead of nothing, so a third heuristic was the wrong tool; `typescript` is a root devDependency for it (one line in `package.json`, three in the lockfile).

With the parser in hand the requirement went up a step — from "the enclosing function takes `BrowserToRelay`" to "the argument handed to `JSON.stringify` is a parameter of the enclosing function, and *that parameter* is `BrowserToRelay`". That catches a mutation every earlier draft passed:

```ts
private send(msg: BrowserToRelay): void {
  const raw: RelayMsg = { ...msg, extra: 1 }
  this.ws.send(JSON.stringify(raw))
}
```

It deliberately does not pin the parameter's name: the declared type is only resolvable when the serialized name *is* a parameter, so the name never has to be asserted.

Your negative case is now a mutation and fails with `serializes payload, which the enclosing function does not declare as BrowserToRelay: expected 'RelayMsg' to be 'BrowserToRelay'`.

**Alias widening in the union guard.** Also confirmed. Members are resolved to their declarations now: an `export interface` must carry a literal `type` field, an `export type` must be a union and is recursed into, and a single-member alias is refused. Recursion turned out to be required rather than defensive — `ClipboardRequest` is a real nested union (`ClipboardRead | ClipboardWrite`, `index.ts:775`), which the first version of this fix failed on before the resolver existed.

Three mutations added: `| UnsafeOutbound` with `type UnsafeOutbound = Record<string, unknown>` beside it; an alias swapped in *inside* the nested union; and an interface with `type: string` instead of a literal.

**`minor` instead of `patch`.** Correct, and `CONTRIBUTING.md:71,73` is the policy you inferred — any breaking change is a `major`, relaxed to `minor` before `v1.0.0`.

The specific mistake is worth naming: the changeset already said the change is source-breaking for an out-of-repo producer, and then argued `patch` from the in-repo census. The census is the wrong question, because the consumers a bump exists to warn are exactly the ones it cannot see. Every package in the fixed group moves to `0.19.0` with it, which is the signal rather than a side effect.

**Not done, with a reason:** `scripts/__tests__/agentSendTyped.test.mjs` pins whole helper bodies as text and has a comparable weakness plus the same `git ls-files` blind spot for untracked files. Rewriting it means re-running its mutation set across three agent packages, which does not belong in this diff. It is recorded in the review record and in the plan's follow-ups.

Fourteen mutations now run and caught in total. Full detail in `.work/reviews/refactor__wire-l4c-client-producer.md`.

PR #518, opened 2026-08-10
feat(protocol): open-url carries a requestId, and both replies echo it

jo-duchan:
First pair of the request/response correlation work (L5a). `open-url` carries a `requestId` and both replies echo it, so a caller matches its own reply instead of matching on `sessionId` + message type and taking whichever arrives first.

Nine pairs correlate that way today, and it is the shared root of #499, #512's first finding, and #444's seven relay reply sites. Three others — `screenshot`, `ui:tree`, `clipboard` — already carry a `requestId` and have none of those problems. This extends what works rather than inventing anything.

Part of the wire-contract program (#501, #505, #503, #506, #509, #513, #516). Breaking-change approval was given for the required field on the request.

## Required on the reply, not optional

A first draft made it optional so an agent predating the field would not falsify the declaration. The design review measured that and reversed it:

- Required yields **complete, precise** in-repo compile errors — ten, at exactly the ten production sites for this pair, nothing else. L4a's typed agent sends are why the write side is covered at all.
- Optional needs a static check to stand in for the compiler, and that check **cannot exist.** Presence is checkable; the property is *provenance*. A check built and run against the clipboard family — 100% correlated today — produced seven false positives, because `respond({ sessionId, requestId, ...body })` puts the `type` literal and the id in different object literals. And it passed when an echo was replaced with a freshly minted id.
- Absence would mean two things wanting opposite handling: "an old agent", and "not a reply at all". The relay's `device:ready` replay is a permanent producer of the second, and reading it as the first is the `{booted: true}` for a boot that never happened that #516 measured and refused to ship.
- The repo had already decided this seven times: of the eleven messages declaring `requestId`, seven are agent-produced replies and all are required.

There is deliberately **no fallback** to `sessionId` + type. The `fixed` version group locks protocol, agent-core, both agents and the relay together, so the in-repo skew window is zero.

## What enforces the echo, precisely

This is the part the code review rewrote, and the numbers are worth stating: **13 attacks, the type caught 3.**

- **Omitting the correlator is a compile error** — from `requestId: string` being required on the reply interfaces, reached through the agents' typed send helper.
- **A freshly minted id written as a literal at the `respond(...)` call is an excess property** — that one is `OpenUrlReplyBody`.
- **`...body` is spread first.** With the ids last, a body *variable* carrying a `requestId` would override the real one; excess-property checking does not fire on variables. An earlier draft had the ids first and let a wrong id win.

Two claims in the first commit were false and are corrected in the code, the comments and the changeset: "the send helper is the only place the id can come from" (`sendMsg` accepts any `string`, so bypassing `respond` type-checks) and "a body cannot declare `requestId`". Each remaining pair needs its own echo tests; this helper does not remove that work.

**The mutation nothing caught:** hoisting the correlator out of per-request scope — a plausible "share `respond` across handlers" refactor — compiles clean, passed all 370 iOS tests, and answers two in-flight requests with the second one's id. That is precisely the class this layer exists to remove, and concurrency was the only reason this pair needs a correlator at all. There is a test for it now.

## The relay had the opposite policy to the agents for the same input

It answered an id-less request with `requestId: msg.requestId!`, which is **not** the `sessionId!` beside it in kind. `sessionId!` feeds a read, so a miss still produces a visible error. `requestId!` feeds a write into an outbound frame — `JSON.stringify` drops the key and ships an `open-url:error` whose required correlator is absent, which every correlating consumer discards, turning "agent offline" into a caller waiting out its full 15s.

Measured: `{"type":"open-url:error","sessionId":"nonexistent-session","message":"agent offline"}`. It declines to answer now, matching both agents. Three relay fixtures were sending id-less requests and asserting only type+message, so the suite was blessing that hole; they carry ids and assert the echo now, and a fourth pins the drop.

## The dashboard was a producer nobody had counted, and correlating it fixes a live bug

It sends `open-url` from `DeepLinkDialog`, and `DeviceViewer` toasts the reply — but a reply goes to whichever socket holds the session, not to whoever asked, so the viewer was showing "Deeplink opened" for `mcp-server`'s deeplinks. The viewer mints and records the id now and toasts only its own. The id comes from `getRandomValues`, not `crypto.randomUUID`, which is secure-context only and therefore absent on the plain-HTTP LAN deployment that is tapflow's primary path — `useClipboardBridge` had already found that and written it down 12 lines from the code this would have copied.

That half shipped with **no tests** in the first commit, which is the same gap that let the whole change pass the iOS suite. Six now, driven through ⌘K → type → Enter rather than by calling `openUrl`, so what is pinned is that the id recorded is the id sent. Verified against three mutations: removing the gate (2 fail), recording an id different from the one sent (5 fail), clearing the record on `session:joined` (1 fail).

## Two additions removed as inert

`RequestReplies` / `ReplyOf` had zero consumers repo-wide — deleting them left `tsc` and all 255 static assertions green, which is exactly the unenforced-mapping property its own doc comment criticised. And `OpenUrl` in `RelayToAgent` enabled nothing: neither agent consumes that union, both still read a hand-written inbound literal that this work widened by hand, and removing the member changed no measurement. Both return when a second pair needs them and something checks them.

## Verification

`tsc -b` 0 · lint 0 errors · `changeset:check` passes.

| suite | |
|---|---|
| scripts (static checks) | 255 |
| relay | **585** / 1 skipped (+1) |
| dashboard | **310** (+6) |
| mcp-server · flow-runner | 64 · 57 |
| ios-agent | **371** (+1) |
| android-agent | 252 |

`pnpm typecheck` is `tsc -b && …`, so a `tsc -b` error short-circuits and `dashboard` / `mcp-server` are never reached in exactly the runs that have errors. Both were run directly.

## Adversarial review

A design pass across two channels ran before any code and reversed the central judgment, plus found that **two of the seven closures I claimed were already closed** — #445 a week earlier by #449, and `clipboard:error`'s `requestId` already required and filled. Then two code channels, worktree-isolated, with briefs narrowed to one blocking question each after the previous rounds cost 50 and 55 minutes; these took 25 and 33.

Record: `.work/reviews/feat__wire-l5a-open-url-correlation.md`.

## Known and deliberately not fixed here

- **Reply routing is unchanged.** Replies still go to whichever socket holds the session rather than to the requester, so an `mcp-server` deeplink on a dashboard-held session goes from "toasted as a lie" to "dropped silently, and mcp-server times out". Right at both endpoints, still misaddressed in the middle. The relay already keys replies by `requestId` for `screenshot` / `ui:tree` via a pending map; the same shape fixes it.
- **Union *membership* is pinned nowhere** — adding a bogus member to `RelayToAgent` leaves all 255 static assertions green, because the check's loop iterates fields rather than enumerating unions. That guard would have caught the inert addition above. Follow-up.
- **A third-party agent predating the field** gets its reply dropped, and an independently installed older `mcp-server` / `flow-runner` will time out `openUrl`. Recorded in the changeset as the upgrade note.

## Summary by CodeRabbit

* **New Features**
  * Added reliable request tracking for deep-link and URL-opening actions.
  * Multiple simultaneous URL-opening requests are handled independently.
  * Dashboard notifications now respond only to matching requests and ignore duplicates or unrelated replies.
  * Added accessible labels for URL inputs and Android system controls.

* **Bug Fixes**
  * Requests without valid tracking IDs are safely ignored.
  * Improved success and error response matching across supported platforms.
  * Pending requests remain correctly tracked across reconnects.


jo-duchan:
Both fixed in `c15146a`. The Major was half right, and chasing it surfaced something worse that I had just written.

**The door check — correct, and it was the same inconsistency I had fixed one commit earlier.** The guard sat inside the offline branch, so an id-less `open-url` was still forwarded to a connected agent: two policies for one input. Both in-repo agents drop it, so nothing observable changes for them, but an agent predating the field would execute the request and reply uncorrelated, and then nothing downstream can attribute that reply. It is now before either branch.

Worth separating from #444 explicitly, since the changeset had leaned on that: general inbound validation is #444's, but this is **one required field on one message** whose absence the relay can act on locally, and leaving it to the receiving agent is what made the guarantee rest on the peer.

**Validating the reply forward — declined, with a reason.** `open-url:done` / `open-url:error` are forwarded from a **shared 14-case block**, and only those two of the fourteen carry a `requestId` at all — the rest arrive in later pairs of this layer. Special-casing a message inside a generic forward is the shape a previous layer of this program (L4b) was retired for, and an uncorrelated reply is already inert at every in-repo consumer, so the benefit is zero and the cost is a per-message branch in a path that deliberately has none. Recorded in the review record rather than skipped silently.

**And your finding caught four tests of mine that were passing for the wrong reason.** This is the useful part.

The drop tests I had just added all read:

```ts
const reply = waitForTypeOrNull(sock, type, 0)
send(...)
await barrier(sock)
expect(await reply).toBeNull()
```

A 0 ms deadline resolves `null` on the next tick — before the round-trip completes — so they passed whether the drop happened or not. The proof is exactly your finding: I restored the arrangement you flagged, expecting the new relay test to fail, and it stayed green.

Corrected to the order the existing suite already used — send, `await barrier`, *then* read the recording with a 0 ms deadline. A barrier is evidence the peer finished with the frame; a deadline is a guess about timing. All four now fail on the mutation they exist for:

| mutation | fails |
|---|---|
| door check moved back into the offline branch | relay forward test |
| door check removed entirely (invalid frame answered) | both relay tests |
| iOS guard removed | iOS drop test |
| Android guard removed | Android drop test |

The general lesson is now in this program's decision log: a test asserting *presence* is self-verifying, because a wrong value fails it. A test asserting **absence** is not — passing when nothing happens is its definition, so it cannot distinguish "the drop worked" from "the test never looked". `toBeNull` / `not.toHaveBeenCalled` / `not.toBeInTheDocument` need the mutation before the commit, not after. In this same round the one test I *did* mutate — the concurrency test — was the only sound one.

**The markdownlint MD018** is fixed, and the sentence it was in had gone stale anyway (it said validation at the door was deferred, which this round changes), so it was rewritten rather than escaped.

Verification after both: `tsc -b` 0 · static suite 255 · relay 586 / 1 skipped · dashboard 310 · ios-agent 371 · android-agent 252 · lint 0 errors.

PR #519, opened 2026-08-11
feat(protocol): the app commands carry a requestId, and the relay carries it across a rebuild

jo-duchan:
Second slice of the request/response correlation work. `app:install`, `app:launch` and `app:clear-state` now correlate by `requestId` like `open-url` (#518), so a caller matches its own reply instead of matching on `sessionId` + message type and taking whichever arrives first.

Part of the wire-contract program (#501, #505, #503, #506, #509, #513, #516, #518). Breaking-change approval was given for the required field on the three requests.

## The new part: a transformation has to carry the correlator

`open-url` was re-serialised whole, so the id rode along for free. Two of these three are different:

```
browser → { buildId }
relay   → DB lookup → { payload: { filePath, bundleId } }   ← a different message
agent   → app:install-done / -error → forwarded generically, never inspected
```

If the id does not reach the agent, nothing downstream can attribute the reply. `AppInstallToAgent` / `AppLaunchToAgent` declare it required, so *dropping* the copy is a compile error.

**Nothing type-checks that the copied value is the request's**, and that was measured rather than assumed. Four candidate guards were built and broken: a branded `CorrelatedId` is laundered by any cast to the brand, because a brand names a *kind* while provenance is a property of the *instance* and TypeScript has no value-dependent types; a generic `Omit`-body helper does not compile without a cast of its own, which is worse than the literal it replaces since a literal at least gets its whole shape checked. So the reply side keeps `<Pair>ReplyBody` — worth it there, with ~20 literal sites — and the request side is held by one test per handler.

## `device:shutdown` was in the plan and came out

Two properties this shape cannot express, both found by trying:

- **The relay originates one.** It sends `device:shutdown` itself when a browser socket closes, and `DeviceShutdown` is a single interface shared by both directions — so a required correlator forces the relay to invent an id for a request nobody made, which is exactly what the door checks exist to prevent.
- **The reply is also a broadcast.** `SessionList` reads `device:shutdown-done` as device status, not as a reply to its own request. A consumer that discards on a correlator mismatch stops learning about shutdowns it did not initiate — the cross-requester delivery that is a bug for `open-url` is the feature here.

It also cannot have a meaningful concurrency test: both agents' shutdown handler returns early once state is gone, so a second request produces no second reply to correlate. It joins `device:boot` in the slice that decides what a relay-originated request and a dual-role reply mean.

`clipboard:read`/`write` were pulled *in* despite not being pairs of this slice — they had the identical defect (`requestId: msg.requestId!`, removed for `open-url` one slice earlier and still live next door), and leaving it would have made this change's own claim of one policy at the door false on landing.

## The review found that the entire reply direction was unverified

Both relay `fail()` closures, the clear-state error exit and all six agent `respond` helpers could emit a fabricated correlator with every suite holding its baseline exactly — relay 591, ios 371, android 252, dashboard 310.

The specific mistake: the plan's test case for this was discharged as "the 16 existing fixtures need editing", and **adding `requestId` to a *request* fixture asserts nothing about the *reply*.** More permissive by omission.

Worse than misattribution, because consumers now gate strictly: a wrong echo is a **loss**. The dashboard discards the reply, nothing clears `installing`, and the Launch control never appears for the life of the mount; `mcp-server` burns 60s and `flow-runner` 120s. That is a new single point of failure inside the very file created for #445.

Added, each verified against the mutation it exists for:

| added | mutation result |
|---|---|
| echo assertions on all **nine** relay error exits | 0 → **13 failures** |
| echo tests per pair on both agents | fabricated correlator: **2 per agent** |
| **six concurrency tests** — required by the plan, absent from the first version | hoisting the correlator compiles clean and passed everything else; now fails its pair |
| dashboard tests for both gates | deleting the install gate / launch gate / launch `add()` each left 310/310; now **2 / 1 / 2** |
| `mcp-server` correlator matching incl. `clearState`, which had no test at all | reverting the predicates to `sessionId` left 64/64; now **2** |

The dashboard gap was a documented repeat: `DeviceViewer.openUrl.test.tsx` opens by saying those tests exist because the previous slice shipped without them.

## Two corrections to reasoning used twice

**The `fixed` version group does not mean what I claimed.** It makes packages *release* together, not *install* together, and the agent runs on a tester's Mac installed separately from the relay. An agent older than this field puts `app:install-done` on the wire with the key absent, the dashboard discards it, and "Installing…" persists with no Launch control — and nothing logged at any of three hops. For `open-url` the same skew cost a toast; here it costs the primary manual-testing flow. The relay's door checks log now, and a test pins the cost rather than a claim that it cannot happen.

**A new non-null assertion slipped in** — `launchApp` had `buildId!`, because hoisting the send out of `installed && buildId ? …` lost the render-condition narrowing. In the slice that removed `msg.requestId!` from the relay. It is a guard now.

## Two things the review left, both included

**"One policy at the door" was prose.** True today — all six browser requests declaring a required `requestId` are gated — but nothing would say so the day a seventh was added ungated, and that is not hypothetical given the clipboard defect above. `scripts/__tests__/correlatedRequestsGated.test.mjs` derives the request set from the protocol and accepts either gate form: an inline `isCorrelated(msg)`, or dispatch to a handler whose `msg` is narrowed to carry the correlator, where the compiler enforces it. Unlike the echo obligation this one *is* checkable, because the property is the presence of a gate rather than the provenance of a value.

Its own first version fell into the trap it warns about — it missed both clipboard members because `ClipboardRequest` is a nested union, and only the count assertion caught that; every per-type assertion would have passed vacuously on a set of four.

**The policy moved next to the types.** Root AGENTS.md requires a landed layer's rationale to live in the relevant AGENTS.md, and neither this slice nor the previous one had done it. `packages/protocol/AGENTS.md` now carries: why the reply is required and what optional would have cost, which mechanism catches what and that **each pair still owes its own echo and concurrency tests**, why there is no request-direction body type, that a transformation carries the correlator, and the two properties that put a pair outside this shape.

## Verification

`tsc -b` 0 · `dashboard` and `mcp-server` typechecks run directly (both off the composite graph, and `pnpm typecheck` short-circuits on a `tsc -b` error) · lint 0 errors.

| suite | |
|---|---|
| scripts | **263** (+8, the gate check) |
| relay | **591** / 1 skipped |
| dashboard | **316** (+6) |
| mcp-server | **67** (+3) |
| flow-runner | 57 |
| ios-agent | **377** (+6) |
| android-agent | **258** (+6) |

## Adversarial review

A design pass across two worktree-isolated channels ran before any code and **cut the slice from four pairs to three**, answered the request-direction-guard question with "there is none", and corrected three of the plan's counts. Then two code channels, same isolation, briefs narrowed to one blocking question each.

Record: `.work/reviews/feat__wire-l5b-app-command-correlation.md`.

## Known and deliberately not fixed here

- **Reply routing is unchanged** — replies go to whichever socket holds the session, not to the requester. Correlation lets a non-requester discard; it does not deliver.
- **A fifth failure exit is uncovered**: a throw out of the build lookup unwinds to the message-loop catch and answers nothing. Pre-existing, and the reason the comment does not say "every exit carries the request's id".
- **`device:shutdown` and `device:boot`** — the next slice, which needs its own design pass.

## Summary by CodeRabbit

* **New Features**
  * Added request correlation for app installation, launch, and clear-state actions.
  * Improved concurrent command handling by matching responses and errors to the correct request.
  * Updated dashboards and automation clients to track command status reliably.
  * Improved validation and handling of incomplete or uncorrelatable commands, including clipboard and URL actions.

* **Documentation**
  * Documented request and response correlation requirements across supported device actions.


jo-duchan:
Both valid, both fixed in `9f7e0b8a`.

**The install/launch records outlived the boot cycle that made them.** Correct, and the mechanism is exactly as described: the install is issued on `device:ready`, and `device:booting` resets `installing`, `installed`, `installError`, `bootError` and `chrome` — everything a new cycle invalidates except the ids. So cycle 1's `app:install-done` could arrive while cycle 2's install was in flight and set `installed`, showing a Launch control for an app not yet on the device.

Worth stating precisely: this is not a regression. Before the correlator, *any* install reply set that flag, including another client's, so the new behaviour is strictly narrower than what it replaced. What the change did was add a record without giving it a lifetime — and the correlator is what makes the precise fix possible at all.

Cleared in the handler that already owns the boundary rather than with a generation counter, since `device:booting` *is* the generation boundary. On the second half of your point — queued rebinds producing several ids — each rebind cycle passes through `device:booting`, so at most one id per cycle survives now. The case that remains is two `device:ready` in one cycle, which the relay's replay can produce; there both ids install the same build, so the first reply clearing `installing` is imprecise rather than wrong. That is written down at the fix rather than left implied.

There is a test, and it fails when the clear is removed.

**The clipboard door test synchronised on the wrong connection.** Also correct, and the reasoning is the one that matters more than the fix: a barrier on the browser proves the relay processed something sent on the *browser*, and says nothing about it having seen the agent go. Until it has, the ungated request is forwarded to a still-open socket, no error is produced, and the assertion holds for a reason unrelated to the gate.

It awaits the close event now — the form `appCommandErrors.test.ts` already used for the same setup, which makes this a consistency fix as much as a correctness one.

This is the third time across two slices that a barrier ended up on a connection other than the one carrying the trigger, and the second time this specific instance was raised: my own review channel flagged it as low severity and I accepted "10 out of 10 detection today" as sufficient. That was the wrong call — latent is not fixed, and a passing mutation shows a test *can* fail, not that it always will. Re-measured after the change: still 10/10, now for a stated reason rather than a timing coincidence.

Verification: `tsc -b` 0 · static suite 263 · relay 591 / 1 skipped · dashboard 317 · lint 0 errors. Both mutations checked — not clearing the records fails the new stale-cycle test; removing the clipboard gate fails the door test in all ten runs.

PR #520, opened 2026-08-11
perf(dashboard): split large vendor bundles

### LINKED ISSUE #168: perf(dashboard): code splitting to reduce initial bundle size
## Background

Vite build currently outputs a single JS chunk of ~1,093 kB (318 kB gzipped), triggering the "Some chunks are larger than 500 kB after minification" warning. All routes are loaded eagerly on first visit.

## Impact

- Slower first load, especially on slower connections
- No benefit from browser caching on per-route code

## Proposed approach

Use React lazy + Suspense for route-level code splitting:

```tsx
const AppCenter = lazy(() => import('./pages/AppCenter'))
const QASession = lazy(() => import('./pages/QASession'))
// ...
```

And configure `build.rollupOptions.output.manualChunks` for vendor splitting (react, react-dom, radix-ui, etc.) if needed.

## Acceptance criteria

- No single chunk exceeds 500 kB after minification
- All routes still load correctly
- Vite build warning disappears

Fhatu12:
## Summary

Closes #168

This reduces the dashboard's largest production JavaScript bundle by splitting only the stable third-party dependency families that stay on the right loading path with Vite/Rollup `manualChunks`.

The dashboard already lazy-loads most routes, and the narrowed chunking keeps route/UI/form code lazy while moving stable app-shell and chart dependencies into reusable vendor chunks.

The final groups are:

- `vendor-react`
- `vendor-router`
- `vendor-charts`

Application code, lazy route chunks and the tinyh264 worker are left to the existing build behaviour rather than being forced into generic vendor chunks.

## Build result

Original base (`a61d9d5d1cb660bd166b3ef4efad7b6c4f97d20d`):

- largest JavaScript chunk: 541.99 kB
- first-load JavaScript: 541,990 B raw / 145,181 B Brotli
- Vite large-chunk warning: present
- JavaScript chunks: 20

Current narrowed solution:

- largest JavaScript chunk: 285.08 kB
- first-load JavaScript: 540,934 B raw / 147,099 B Brotli
- first-load Brotli delta vs base: +1,918 B (+1.32%)
- Vite large-chunk warning: absent
- JavaScript chunks: 23
- largest-chunk reduction vs base: 256.91 kB (~47.4%)

Significant final chunks included:

- application entry: ~285.08 kB
- `vendor-react`: ~217.28 kB
- `tinyh264.worker`: ~173.47 kB
- `QASession`: ~109.80 kB
- `vendor-charts`: ~77.25 kB
- `vendor-router`: ~38.58 kB
- `dialog`: ~25.16 kB
- `AppCenter`: ~14.89 kB
- `MacResources`: ~7.20 kB
- `tabs`: ~4.15 kB
- `alert-dialog`: ~3.75 kB

`vendor-charts` is emitted separately and is not modulepreloaded by `index.html`; it remains lazy and keeps `@visx` out of duplicated route chunks. `vendor-ui` and `vendor-forms` are intentionally not split because they made lazy UI/form code part of first load.

Generated filenames/hashes may vary between builds.

## Regression guard

Added `scripts/__tests__/dashboardFirstLoadBudget.test.mjs`, which builds the production dashboard, parses `packages/dashboard/dist/index.html`, collects the entry script plus `modulepreload` JavaScript assets, sums the `.br` files served by the relay, and enforces a 155,000 B first-load Brotli budget.

The guard passed at 147,099 B for the current solution. Temporarily restoring the earlier eager `vendor-ui` / `vendor-forms` split made it fail at 159,588 B, with those chunks listed as first-load assets.

## Validation

- `pnpm exec vitest run --config scripts/vitest.config.mjs scripts/__tests__/dashboardFirstLoadBudget.test.mjs`
- negative guard check against commit `252262bad79766fbffe52682a6c7c9eed57f73ee`'s `packages/dashboard/vite.config.ts`: failed as expected at 159,588 B / 155,000 B budget
- temporary base measurement against `a61d9d5d1cb660bd166b3ef4efad7b6c4f97d20d`'s `packages/dashboard/vite.config.ts`
- `pnpm test`
- `pnpm --filter @tapflowio/protocol build`
- `pnpm --filter @tapflowio/dashboard build`
- `pnpm --filter @tapflowio/dashboard test` (31 files / 310 tests passed)
- `pnpm --filter @tapflowio/dashboard typecheck`
- `pnpm --filter @tapflowio/dashboard lint`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm --filter @tapflowio/relay build`
- `pnpm changeset:check`
- `git diff --check`

The existing root lint still reports two warnings outside this change.

No router source was changed. Existing lazy route chunks continue to be emitted separately. The `tinyh264.worker` output remains isolated from dashboard `manualChunks`.

## Release notes

- added a patch changeset for `@tapflowio/relay`, which ships the dashboard
- updated the root `[Unreleased]` changelog

## Checklist

- [x] Tests passing
- [x] No `any`

## Related `.work/` docs

None.

## Summary by CodeRabbit

* **Performance**
  * Improved dashboard loading by splitting vendor dependencies into smaller, optimized chunks.
  * Organized React, routing, and charting libraries into dedicated bundles for more efficient loading.
  * Improved cache reuse across releases, reducing the amount of data downloaded after updates.
  * Added safeguards to help keep the dashboard’s initial load within its performance budget.

* **Changelog**
  * Documented the dashboard bundle and caching improvements in the upcoming release notes.

### DIFF
--- a/.changeset/dashboard-vendor-chunks.md
+++ b/.changeset/dashboard-vendor-chunks.md
@@ -2,4 +2,4 @@
 '@tapflowio/relay': patch
 ---
 
-Improve dashboard loading performance by splitting large application and vendor bundles.
+Split stable dashboard vendor dependencies into smaller chunks to reduce maximum bundle size and improve cache reuse across releases.
--- a/.changeset/dashboard-vendor-chunks.md
+++ b/.changeset/dashboard-vendor-chunks.md
@@ -1,0 +1,5 @@
+---
+'@tapflowio/relay': patch
+---
+
+Improve dashboard loading performance by splitting large application and vendor bundles.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Improve dashboard loading performance by splitting large application and vendor bundles.
+
 ### Fixed
 
 - **A scrcpy server process error could take down every Android device the agent manages, not just the one session.** The scrcpy server process spawned for a real-device session had no error handler; an unhandled error on it (e.g. the server process failing to spawn, or a permission error on kill) crashed the whole android-agent process, ending every session it was managing. It's now logged instead.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Improve dashboard loading performance by splitting large application and vendor bundles.
+- Split stable dashboard vendor dependencies into smaller chunks to reduce maximum bundle size and improve cache reuse across releases.
 
 ### Fixed
 
--- a/packages/dashboard/vite.config.ts
+++ b/packages/dashboard/vite.config.ts
@@ -27,17 +27,11 @@ export default defineConfig({
             return 'vendor-router'
           }
 
-          if (id.includes('/react-hook-form/') || id.includes('/@hookform/') || id.includes('/zod/')) {
-            return 'vendor-forms'
-          }
-
-          if (id.includes('/@radix-ui/') || id.includes('/lucide-react/') || id.includes('/sonner/')) {
-            return 'vendor-ui'
-          }
-
           if (id.includes('/@visx/') || id.includes('/d3-array/')) {
             return 'vendor-charts'
           }
+
+          return undefined
         },
       },
     },
--- a/packages/dashboard/vite.config.ts
+++ b/packages/dashboard/vite.config.ts
@@ -16,6 +16,9 @@ export default defineConfig({
     outDir: 'dist',
     rollupOptions: {
       output: {
+        // Only split families the app shell already loads synchronously, plus charts.
+        // Grouping @radix-ui / react-hook-form here was measured and reverted: it turned
+        // lazy dialog and form code into modulepreloads, adding ~14 kB Brotli to first paint.
         manualChunks(id) {
           if (!id.includes('/node_modules/')) return undefined
 
--- a/packages/dashboard/vite.config.ts
+++ b/packages/dashboard/vite.config.ts
@@ -14,6 +14,33 @@ export default defineConfig({
   },
   build: {
     outDir: 'dist',
+    rollupOptions: {
+      output: {
+        manualChunks(id) {
+          if (!id.includes('/node_modules/')) return undefined
+
+          if (id.includes('/react/') || id.includes('/react-dom/') || id.includes('/scheduler/')) {
+            return 'vendor-react'
+          }
+
+          if (id.includes('/react-router/') || id.includes('/react-router-dom/')) {
+            return 'vendor-router'
+          }
+
+          if (id.includes('/react-hook-form/') || id.includes('/@hookform/') || id.includes('/zod/')) {
+            return 'vendor-forms'
+          }
+
+          if (id.includes('/@radix-ui/') || id.includes('/lucide-react/') || id.includes('/sonner/')) {
+            return 'vendor-ui'
+          }
+
+          if (id.includes('/@visx/') || id.includes('/d3-array/')) {
+            return 'vendor-charts'
+          }
+        },
+      },
+    },
   },
   // ESM worker (tinyh264.worker imports tinyh264) — 'es' format so the worker chunk
   // can code-split its static imports. Default 'iife' breaks on code-split workers.
--- a/scripts/__tests__/dashboardFirstLoadBudget.test.mjs
+++ b/scripts/__tests__/dashboardFirstLoadBudget.test.mjs
@@ -1,0 +1,95 @@
+// Protect the metric #168 actually cares about: JavaScript fetched before first paint.
+//
+// Vite's large-chunk warning only watches individual raw chunks. The first vendor split looked
+// good by that signal while adding eager modulepreloads. Build the real dashboard output, read
+// index.html, and budget the precompressed Brotli bytes that the relay serves.
+import { beforeAll, describe, expect, it } from 'vitest'
+import { execFileSync } from 'child_process'
+import { existsSync, readFileSync, statSync } from 'fs'
+import { dirname, relative, resolve } from 'path'
+import { fileURLToPath } from 'url'
+
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..', '..')
+const DASHBOARD_DIST = resolve(ROOT, 'packages', 'dashboard', 'dist')
+const INDEX_HTML = resolve(DASHBOARD_DIST, 'index.html')
+const FIRST_LOAD_BROTLI_BUDGET = 155_000
+
+function buildDashboardDist() {
+  const env = { ...process.env, NODE_ENV: 'production' }
+  execFileSync('pnpm', ['--filter', '@tapflowio/protocol', 'build'], { cwd: ROOT, env, stdio: 'inherit' })
+  execFileSync('pnpm', ['--filter', '@tapflowio/dashboard', 'build'], { cwd: ROOT, env, stdio: 'inherit' })
+}
+
+function attributes(tag) {
+  const out = new Map()
+  const pattern = /([:\w-]+)\s*=\s*(['"])(.*?)\2/g
+  for (const match of tag.matchAll(pattern)) out.set(match[1].toLowerCase(), match[3])
+  return out
+}
+
+function distAssetPath(assetPath) {
+  const withoutOrigin = assetPath.replace(/^[a-z]+:\/\/[^/]+/i, '')
+  const withoutQuery = withoutOrigin.split(/[?#]/, 1)[0]
+  const relativeAsset = withoutQuery.replace(/^\/+/, '')
+  const resolved = resolve(DASHBOARD_DIST, relativeAsset)
+  if (relative(DASHBOARD_DIST, resolved).startsWith('..')) {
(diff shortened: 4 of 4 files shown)


jo-duchan:
Thanks for picking this one up, and for the unusually careful PR description — the build numbers made it easy to verify. I rebuilt both configurations locally against the same source and reproduced your before/after chunk sizes exactly, down to the byte. The measurement is solid. My one concern is which number we're optimizing for.

## The largest chunk shrank, but the first load grew

`manualChunks` regroups modules, it doesn't remove them. What #168 is really about ("slower first load, especially on slower connections") is the total JS the browser fetches before first paint — the entry `<script>` plus every `modulepreload` Vite injects into `index.html`. The relay serves the precompressed `.br` files, so brotli is what actually crosses the wire:

| | largest chunk | first-load JS (raw) | first-load JS (brotli) | Vite warning |
|---|---|---|---|---|
| `main` | 541.99 kB | 541,990 B | 145,320 B | present |
| this PR | 217.28 kB | 592,651 B | **159,621 B** | absent |

That's +14.3 kB brotli (+9.8%) before first paint.

The cause is `vendor-ui` and `vendor-forms`. Both land in `index.html` as `modulepreload`, which pulls code onto the critical path that was previously loading lazily. The lazy chunks that shrank show where it went:

- `dialog` 25.12 → 5.19 kB
- `tabs` 4.11 → 1.91 kB
- `alert-dialog` 3.72 → 2.06 kB

Radix's dialog code is only needed once someone opens a dialog, but it now downloads on the login screen.

## `vendor-charts` is a clear win

Worth calling out separately, because this one is doing exactly what you'd want: `vendor-charts` is *not* preloaded, so it stays lazy, and it de-duplicates `@visx`, which was previously emitted into both `MacResources` and `QASession` (`MacResources` 84.09 → 7.27 kB).

## Suggestion

Keep `vendor-react`, `vendor-router` and `vendor-charts`; drop `vendor-ui` and `vendor-forms` and let Rollup's automatic splitting handle Radix and the form libraries. I built this variant too:

| | largest chunk | first-load (brotli) | Vite warning |
|---|---|---|---|
| narrowed | 285.08 kB | 147,089 B (+1.2%) | absent |

It still clears the acceptance criteria in #168 — no chunk over 500 kB, warning gone — keeps the charts de-duplication, and leaves first-paint bytes essentially where they were. React and the router are used synchronously by the app shell anyway, so grouping them eagerly costs nothing, and they're also the chunks that benefit most from staying cached across releases, which was part of the motivation here.

```ts
manualChunks(id) {
  if (!id.includes('/node_modules/')) return undefined
  if (id.includes('/react/') || id.includes('/react-dom/') || id.includes('/scheduler/')) return 'vendor-react'
  if (id.includes('/react-router/') || id.includes('/react-router-dom/')) return 'vendor-router'
  if (id.includes('/@visx/') || id.includes('/d3-array/')) return 'vendor-charts'
  return undefined
}
```

## Smaller notes

- **Changelog wording.** "Improve dashboard loading performance" isn't quite what the change does, even in the narrowed form. Something about chunk size and cache reuse across releases would be more accurate. The changeset naming `@tapflowio/relay` rather than the dashboard is right, and it's the detail people most often get wrong on a first dashboard PR.
- **Nothing guards this.** The only signal that the chunking still works is the Vite warning, and it doesn't fail the build. One new dependency and this quietly regresses. A test pinning the first-load byte total would be worth having — say the word and I'll point at where it fits.
- The checklist has "Tests written and passing" ticked but no test was added. Not a problem by itself, just worth keeping the checklist honest.
- `d3-array` is listed, but `@visx` also pulls `d3-scale`, `d3-shape`, `d3-path` and others. I checked: widening the match to `/d3-` produces a byte-identical `vendor-charts`, so there's no practical difference. Purely a consistency thing if you care.
- Style nit: the first branch returns `undefined` explicitly while the function falls through implicitly at the end. Worth picking one.

I also confirmed the worker build is untouched — `tinyh264.worker` keeps the same hash, so `build.rollupOptions` isn't leaking into `worker.rollupOptions`. That was the thing most likely to break here, and it didn't.

Great first contribution to tapflow — small, scoped, and properly validated. Once `vendor-ui` and `vendor-forms` come out I think this is ready to land.

Fhatu12:
Thanks for the detailed review and for checking the numbers on your side too. The first-load comparison makes sense, and I can see why the UI/forms chunks are pushing too much onto the initial load.

I’ll narrow it down to React, router and charts, update the wording, and add a small regression check so this doesn’t quietly creep back later.

I’ll rerun everything and update the PR once that’s done.

jo-duchan:
Rebuilt this locally and the output matches yours exactly — the same chunk hashes I got when I tried the narrowed variant on my side (`index-DYsf78cu`, `vendor-react-AJpf4YZ0`, `vendor-router-ChFy8CZ1`, `vendor-charts-DyX5CBcf`), so we're looking at identical bytes.

| | largest chunk | first-load (raw) | first-load (brotli) | Vite warning |
|---|---|---|---|---|
| `main` | 541.99 kB | 541,990 B | 145,320 B | present |
| first revision | 217.28 kB | 592,651 B | 159,621 B | absent |
| **now** | 285.08 kB | 540,934 B | 147,092 B (+1.2%) | absent |

Everything that got pulled forward is back where it belongs: `dialog` (25.16 kB), `tabs` (4.15 kB) and `alert-dialog` (3.75 kB) are lazy again, `vendor-charts` is absent from `index.html`'s preloads so it stays lazy, and `MacResources` keeps the de-duplication win at 7.20 kB. `tinyh264.worker` still hashes to `DiJ50Hai`, same as `main`.

What I appreciated more than the config change itself: you rewrote the "Build result" section to lead with first-load bytes instead of largest chunk, and you wrote down *why* `vendor-ui` and `vendor-forms` are deliberately left out. That second part is what keeps this from being quietly undone by someone who later sees an obviously incomplete pattern. Ticking the checklist down to "Tests passing" rather than leaving the inaccurate line was a good call too.

Two things, and then I'm happy to merge.

## 1. Move that reasoning next to the code

The PR body explains why UI and forms aren't split, but nobody opening `vite.config.ts` a year from now reads PR bodies. A couple of lines above `manualChunks` would carry it:

```ts
// Only the families the app shell already loads synchronously, plus charts.
// Grouping @radix-ui / react-hook-form here too was measured and reverted: it turned
// lazy dialog and form code into modulepreloads, adding ~14 kB brotli to first paint.
```

This repo leans on that convention fairly hard — the reason written beside the code counts as part of the change, not as documentation of it.

## 2. The regression check you mentioned

You said you'd add one, and I offered to point at where it goes, so here it is.

Repo-level structural guards live in `scripts/__tests__/`, run from the root `pnpm test` via `vitest run --config scripts/vitest.config.mjs`. `testsReadSource.test.mjs` is the closest precedent — it's the existing test that reads build output rather than source.

The assertion that matters is the **first-load Brotli total**, not the largest chunk. Largest chunk is exactly the metric that looked healthy while the first revision regressed, so a guard built on it would have passed that revision. Concretely: parse `packages/dashboard/dist/index.html`, collect the entry `<script src>` plus every `<link rel="modulepreload">`, sum the matching `.br` files, and assert the total stays under a ceiling. Something like 155,000 B leaves room for normal growth without hiding another 10% slide.

One trap worth naming up front: the test needs `dist/` to exist, and the obvious workaround — skipping when it's absent — produces a test that passes most confidently exactly when it has verified nothing. Either build inside the test (the dashboard build is ~3s) or fail with a message telling the runner to build first. Your call which.

If you'd rather land the config now and take the guard as a follow-up, that's genuinely fine — say the word and I'll merge this and open an issue for it.

---

Either way, this is a strong first contribution. You got a review saying your headline number measured the wrong thing, went and verified that independently, and came back with a smaller change and better numbers. That's rarer than it should be, and it's the reason this went two rounds instead of ten.

If tapflow turns out to be useful to you, a ⭐ on the repo helps other people find it. No obligation at all.

Fhatu12:
Thanks for pointing me in the right direction on the guard.

Both final items are in now — I added the reasoning next to `manualChunks` and the first-load Brotli budget test under `scripts/__tests__/`.

The guard passes the current build at 147,099 B against the 155,000 B budget, and I also checked it against the earlier UI/forms split where it fails as expected at about 159.6 kB.

Everything is green on my side, including the root test suite, and the GitHub CI checks have passed too. Ready from my side.

jo-duchan (approved):
Both items landed, and I verified the guard rather than taking it on faith.

- Current build: 147,092 B first-load Brotli, passes in 6.3s.
- Checked out the first revision's `vite.config.ts` and re-ran it: **fails at 159,621 B** — the same number I measured independently in the first review, to the byte. The guard catches the exact regression it was written for.

Thank you for the way you handled this. The `beforeAll` build is the right call and not the obvious one: reusing an existing `dist/` would have been faster and would have quietly measured whatever was last built. This repo has been bitten by exactly that before, so building unconditionally is worth the ~6s. You also put the measurement into the comment above `manualChunks`, which is what stops someone from "completing" the pattern later and undoing the whole thing.

Three rounds, and every point came back addressed with your own verification attached rather than just applied. That is a genuinely good way to work, and it made reviewing this easy.

One thing I found while testing, which is **not** a change request for this PR: if `manualChunks` is deleted entirely, first-load drops to 145,320 B, the budget passes, and the 541.99 kB chunk plus the Vite warning come back unnoticed. The guard protects the regression this PR could have caused, but not the fix it makes. That gap predates your work and is a separate few lines, so I've opened #524 for it rather than holding this up.

Approving. I'll merge shortly. Thanks again for taking on a `good first issue` and treating it seriously.

Fhatu12:
Thanks again for taking the time to review this so thoroughly and for merging it. I really appreciated the way you walked me through the first-load trade-off — I learned a lot from that review, and the final solution ended up much better because of it. I’ll have a look at #524 next since it follows on naturally from this work.

PR #521, opened 2026-08-11
feat(protocol): correlate device:boot and device:shutdown

jo-duchan:
## Summary

The lifecycle pair joins the four already correlated by `requestId` (`screenshot`, `ui:tree`, `clipboard`, the app commands). It is the first whose correlator is **optional on every reply**, and that is the design rather than a concession: `device:ready`, `device:boot-error` and `device:shutdown-done` all have producers that answer no request at all, so absence has a real and permanent meaning — *this frame is not the answer to anything*. Not "an old agent".

`device:boot` is **required**. Nothing originates a boot but a browser, and a request passes *through* the relay, so one door gates and logs every sender at once. `device:shutdown` cannot be required: the relay sends it itself from the idle timer. Breaking-change approval for the required field was given before implementing.

## What correlation actually buys here

Narrower than the obvious claim, and review caught a draft making the obvious one. It does **not** let two concurrent boots both resolve — the agents answer a superseded boot with nothing at all (`bootSeq` returns silently at every checkpoint), so one of two overlapping boots times out either way. Both clients' `dispatch` resolves the first waiter whose predicate matches and then stops, so on `sessionId` + type alone that single reply went to the boot registered **first**, the superseded one — and the boot that actually happened was reported as a failure. The correlator sends the reply to the request it answers. Same one timeout, correct attribution.

## What the design review deleted

The plan's central premise. A first draft made `DeviceReady.sessionId` required and had the relay stamp its replay, then argued a forced choice: correlate strictly and break every agent predating the field, or keep a fallback and let a replayed ready satisfy an in-flight boot. Measured against a real `RelayServer`, that choice was manufactured by the draft itself — **the discriminator today is that the replay carries no `sessionId`**, and both boot waiters already depend on it. Requiring and stamping it is what deletes the working guard.

Dropped. `sessionId` on `device:ready` stays optional; closing #516's deferral is its own later slice, and it must carry a test pinning the *value* stamped, because there is none: with the stamp mutated to `session.deviceId` — a plausible slip on a line whose payload reads `{ deviceId: session.deviceId }` — every relay and dashboard test and the whole static suite still passed.

The review also found the producer inventory wrong: `AndroidAgent.restartVideoStream` sends `device:boot-error` for a video stream that died mid-session, with no `device:boot` behind it. So that message earns its optional correlator **on its own merits**, and `DeviceViewer` must not gate on it — it is the only surface reporting a dead stream (#426).

## Optional means tests are the entire enforcement

`<Pair>ReplyBody` cannot be built for a field an object may omit, and `scripts/__tests__/correlatedRequestsGated.test.mjs` derives its set only from *required* declarations — so it cannot see the replies, including at the position that matters most: **the relay is itself a producer of `device:boot-error`**, answering a boot it cannot hand to an agent. An uncorrelated diagnosis there is read as unsolicited and discarded, and the caller waits out its deadline instead of failing. That exact defect shipped twice from agent code (`open-url:error`, then `clipboard:error` a slice later); here it sits where no check reaches, so a relay test holds it.

`case 'device:boot': case 'device:shutdown':` was one fall-through clause and is now two — not cosmetic. A gate written into the shared body would have stopped the dashboard's four shutdown senders and the relay's own idle timer from reaching the agent, silently, in the one direction nothing replies to.

## Promoting the request side hung two suites rather than failing to compile

42 fixtures across `IOSAgent.test.ts` and `AndroidAgent.test.ts` send `device:boot` through a real `RelayServer` from inside `JSON.stringify({ … })`, where no annotation exists for a compiler to check. The three **documented** recipes those were copied from are updated too (`ios-agent/AGENTS.md`, `test-utils/src/socket.ts`, `test-utils/AGENTS.md`) — leaving them would hand the next contributor the same mystery. The dashboard has a rule against untyped injected fixtures; the agent packages have no equivalent. Worth one, out of scope here.

## Review

Design review first (two channels over the plan, since this spans several packages), then the pre-PR review — two independent channels in parallel, one with mutation authority and one strictly read-only. **8 findings, all mine, all dispositioned.** 23 author-side mutations before review; all 23 probed what the gate does *with* an id, and review found the blind spot that left — four mutations on how ids enter and leave `bootIdsRef` survived the whole suite, the worst being a `clear()` in the `device:booting` branch, which is invited by the comment above it and breaks every boot while the spinner clears and the device looks healthy.

One finding's conclusion was re-graded on verified grounds, and one of my own claims was false in a way worth naming: I wrote that promoting the field produced "zero compile errors — every sender already minted an id", which argues the compiler cannot see the request side. It can — all four boot senders go through a sink typed `BrowserToRelay` and every one would have errored; they did not only because earlier hunks of the same commit had added the ids.

Record: `.work/reviews/feat__wire-l5b-prime-lifecycle-correlation.md` (HEAD `a669e0a76d0bb94c223efa415632ae175569d50e`), including the cleared lists re-checked after the fixes and three findings skipped with reasons.

## Verification

`tsc -b` clean; `protocol` (incl. `tsconfig.assertions.json`), `dashboard`, `mcp-server`, `test-utils` typecheck clean; lint and a11y-lens clean.

static **264** · agent-core **137** · relay **601** (1 skipped) · dashboard **329** · mcp-server **76** · flow-runner **63** · ios-agent **382** · android-agent **263** · cli **246** · audiotap-helper **4**.

## Summary by CodeRabbit

* **New Features**
  * Added request correlation for device boot and shutdown operations across supported device workflows.
  * Boot requests now use unique request IDs to match responses to the correct operation.
  * Added protection against stale, mismatched, or replayed device-ready responses.
  * Legacy responses without correlation IDs remain supported where applicable.

* **Bug Fixes**
  * Improved handling of boot failures, shutdown completion, reconnects, and unsolicited restart errors.

* **Documentation**
  * Clarified lifecycle message correlation rules and supported fallback behavior.

### DIFF
--- a/.changeset/lifecycle-correlation.md
+++ b/.changeset/lifecycle-correlation.md
@@ -15,8 +15,16 @@ whole design — `device:ready`, `device:boot-error` and `device:shutdown-done` 
 request at all, so absence has a real and permanent meaning: *this frame is not the answer to anything*.
 
 Consumers correlate when the id is present and fall back to `sessionId` + type when it is not
-(`mcp-server`, `flow-runner`). That fallback is not compatibility slack that expires — it is what lets the
-relay's own replay and Android's mid-session stream failure keep reaching the surfaces that report them.
+(`mcp-server`, `flow-runner`). Precisely what that fallback carries is worth separating, because the two
+id-less cases arrive by different routes and only one of them is permanent:
+
+- **Android's mid-session `device:boot-error`** carries a `sessionId`, so it reaches a client waiter through
+  the fallback. There is no request behind it and never will be, so this half does not expire.
+- **An agent predating the echo** likewise, on either reply. This half is compatibility slack.
+- **The relay's `device:ready` replay** does *not* reach a client at all — it carries no `sessionId`, so the
+  comparison ahead of the fallback excludes it, which is the point of the "not satisfied by the replay" test
+  in both clients. It reaches the **dashboard**, and it gets there because `'sessionId' in msg` is false, not
+  because of anything the correlator or the fallback does.
 
 What correlation buys on this pair is narrower than the obvious claim, and review caught the first draft
 making the obvious one. It does **not** let two concurrent boots both resolve: the agents answer a
--- a/.changeset/lifecycle-correlation.md
+++ b/.changeset/lifecycle-correlation.md
@@ -1,0 +1,112 @@
+---
+'@tapflowio/protocol': minor
+'@tapflowio/relay': patch
+'@tapflowio/ios-agent': patch
+'@tapflowio/android-agent': patch
+'@tapflowio/mcp-server': patch
+'@tapflowio/flow-runner': patch
+---
+
+feat(protocol): correlate device:boot and device:shutdown, with an optional correlator
+
+The lifecycle pair joins the four already correlated by `requestId` (`screenshot`, `ui:tree`, `clipboard`, the
+app commands). It is the first one whose correlator is **optional on every reply**, and that difference is the
+whole design — `device:ready`, `device:boot-error` and `device:shutdown-done` all have producers that answer no
+request at all, so absence has a real and permanent meaning: *this frame is not the answer to anything*.
+
+Consumers correlate when the id is present and fall back to `sessionId` + type when it is not
+(`mcp-server`, `flow-runner`). That fallback is not compatibility slack that expires — it is what lets the
+relay's own replay and Android's mid-session stream failure keep reaching the surfaces that report them.
+
+What correlation buys on this pair is narrower than the obvious claim, and review caught the first draft
+making the obvious one. It does **not** let two concurrent boots both resolve: the agents answer a
+superseded boot with nothing at all (`bootSeq` returns silently at every checkpoint), so one of two
+overlapping boots times out either way. Both clients' `dispatch` resolves the first waiter whose predicate
+matches and then stops — so on `sessionId` + type alone, that single reply went to the boot registered
+**first**, the superseded one, and the boot that actually happened was reported as a failure. The
+correlator sends the reply to the request it answers. Same one timeout, correct attribution.
+
+## What the design review changed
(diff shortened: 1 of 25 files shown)


jo-duchan:
CodeRabbit's finding is accepted, and it is right in a way worth stating rather than patching. Pushed as `98d08cc`.

The changeset said the clients' `sessionId` fallback "is what lets the relay's own replay … keep reaching the surfaces that report them". It does not — and this PR contains the tests that say so: *"is not satisfied by the relay's replay"* in both `mcp-server` and `flow-runner`. The replay carries no `sessionId`, so the comparison **ahead of** the fallback excludes it. It reaches the **dashboard**, and only because `'sessionId' in msg` is false there. Two different mechanisms, one sentence.

The prescription was "keep the fallback and fix the wording". The fallback was never at risk — it is tested from both directions, and reverting `bootDevice` to `sessionId` + type fails four tests. So the wording was the whole of it.

**Beyond what was reported**: the identical conflation sat in both clients' `correlatesWith` doc comment, which called absence "a real and permanent case" for *both* replies. Measured — the only in-repo producers of an id-less `device:ready` are the relay replay (excluded from a client) and nothing else, since both agents now echo. So at a client the two halves differ and the comment flattened them:

- `device:boot-error` id-less is **permanent** — Android's `restartVideoStream` reports a stream that died mid-session, with no request behind it, and it carries a `sessionId` so it does reach a client waiter.
- `device:ready` id-less can only be **an agent predating the echo** — compatibility slack, which will expire.

Corrected in all three places. No code changed; `tsc -b` clean, static 264, mcp-server 76, flow-runner 63.

This is the third instance in this slice of one failure mode — prose that argues for the design while misstating the mechanism. Two were found by the pre-PR review channels, one by CodeRabbit, one by me. The review record notes it as a pattern rather than three separate slips.

Review record refreshed against the new HEAD: `.work/reviews/feat__wire-l5b-prime-lifecycle-correlation.md`.

PR #525, opened 2026-08-11
test(scripts): guard dashboard chunk size regression

### LINKED ISSUE #524: test(scripts): first-load budget guard misses manualChunks being removed
## Background

#520 added `scripts/__tests__/dashboardFirstLoadBudget.test.mjs`, which builds the dashboard and asserts the first-load Brotli total (entry `<script>` + every `modulepreload`) stays under 155,000 B. It works: reverting to that PR's first revision makes it fail at 159,621 B, exactly as intended.

But it only guards one half of #168.

## The gap

Delete `build.rollupOptions.output.manualChunks` from `packages/dashboard/vite.config.ts` entirely and run the guard:

```
dist/assets/index-YWsnsJWL.js   541.99 kB │ gzip: 169.91 kB
(!) Some chunks are larger than 500 kB after minification.
 Test Files  1 passed (1)
      Tests  1 passed (1)
```

First-load drops to 145,320 B — *below* the budget, so the assertion is happier than before — while the 541.99 kB chunk and the Vite large-chunk warning both come back.

The two metrics move in opposite directions when the chunking is removed, which is why one of them cannot stand in for the other. #168's acceptance criteria were:

- No single chunk exceeds 500 kB after minification
- Vite build warning disappears

Neither is currently guarded. The Vite warning is printed, but it does not fail the build, so nothing reports this.

## Proposed fix

Add a second assertion to the same test file: no emitted `dist/assets/*.js` exceeds 500 kB raw. That encodes #168's criterion directly and mirrors Vite's own threshold, so the test fails at the same point the warning appears.

The build is already run in `beforeAll`, so this needs no new setup — a `readdirSync` over `dist/assets`, a `statSync` per `.js`, and one assertion. Current headroom: the largest chunk is `index` at 285.08 kB, and the next is `vendor-react` at 217.28 kB.

Keep both assertions. First-load Brotli catches bytes moving onto the critical path; largest-chunk-raw catches the chunking being removed. Each is blind to the failure the other sees.

## Acceptance criteria

- Removing `manualChunks` from `packages/dashboard/vite.config.ts` fails `pnpm test`
- Re-splitting `@radix-ui` / `react-hook-form` into eager vendor chunks still fails the existing first-load budget
- The failure message names the offending chunk and its size

## Notes for whoever picks this up

Run the guard on its own with:

```sh
pnpm exec vitest run --config scripts/vitest.config.mjs dashboardFirstLoadBudget
```

Do not reach for `npx vitest` — see the test hygiene section in [AGENTS.md](../blob/main/AGENTS.md).

Verify the fix by actually breaking the config both ways and watching the right assertion fail each time. A guard that has never been observed failing is not yet known to guard anything.

Fhatu12:
## Summary

Closes #524

Extends the dashboard build regression guard added in #520 with a second, independent assertion for emitted JavaScript chunk size.

The existing check protects first-load Brotli bytes, but removing `manualChunks` lowers that metric while restoring the 541.99 kB application chunk and Vite warning. This adds a raw-size guard so the two failure modes are covered separately.

## What changed

- scan emitted `packages/dashboard/dist/assets/*.js` files dynamically
- fail if any JavaScript chunk exceeds 500,000 B raw
- include the offending chunk name and byte size in the failure message
- keep the existing 155,000 B first-load Brotli budget unchanged

## Validation

Healthy configuration:

- first-load Brotli: 147,129 B / 155,000 B
- largest JS chunk: 285,077 B / 500,000 B
- targeted guard: 2 tests passed

Negative checks:

- removing `manualChunks`
  - first-load Brotli guard still passes
  - raw chunk guard fails at 541,990 B as expected

- restoring the rejected eager UI/forms split
  - existing first-load Brotli guard fails at 159,602 B as expected
  - confirms the two guards protect different regressions

Full validation:

- `pnpm exec vitest run --config scripts/vitest.config.mjs dashboardFirstLoadBudget`
- `pnpm test` — 180 files passed / 1 skipped; 2370 tests passed / 1 skipped
- `pnpm typecheck`
- `pnpm lint`
- `git diff --check`

The existing unrelated lint warnings remain unchanged.

## Scope

Only:

- `scripts/__tests__/dashboardFirstLoadBudget.test.mjs`

No production configuration or dependency changes.

## Summary by CodeRabbit

* **Tests**
  * Added validation to ensure dashboard JavaScript assets stay within a 500,000-byte size limit.
  * Added checks that the dashboard assets directory exists and reports details for oversized chunks.
  * Preserved the existing Brotli-compressed first-load budget validation.

### DIFF
--- a/scripts/__tests__/dashboardFirstLoadBudget.test.mjs
+++ b/scripts/__tests__/dashboardFirstLoadBudget.test.mjs
@@ -5,7 +5,7 @@
 // index.html, and budget the precompressed Brotli bytes that the relay serves.
 import { beforeAll, describe, expect, it } from 'vitest'
 import { execFileSync } from 'child_process'
-import { existsSync, readFileSync, statSync } from 'fs'
+import { existsSync, readFileSync, readdirSync, statSync } from 'fs'
 import { dirname, relative, resolve } from 'path'
 import { fileURLToPath } from 'url'
 
@@ -13,6 +13,7 @@ const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..', '..')
 const DASHBOARD_DIST = resolve(ROOT, 'packages', 'dashboard', 'dist')
 const INDEX_HTML = resolve(DASHBOARD_DIST, 'index.html')
 const FIRST_LOAD_BROTLI_BUDGET = 155_000
+const MAX_RAW_JS_CHUNK_SIZE = 500_000
 
 function buildDashboardDist() {
   const env = { ...process.env, NODE_ENV: 'production' }
@@ -80,6 +81,22 @@ function firstLoadBrotliReport() {
   return { total, details: lines.join('\n') }
 }
 
+function oversizedJavascriptChunks() {
+  const assetsDir = resolve(DASHBOARD_DIST, 'assets')
+  if (!existsSync(assetsDir)) {
+    throw new Error('packages/dashboard/dist/assets is missing after dashboard build')
+  }
+
+  return readdirSync(assetsDir)
+    .filter((name) => name.endsWith('.js'))
+    .map((name) => {
+      const asset = resolve(assetsDir, name)
+      return { name, size: statSync(asset).size }
+    })
+    .filter(({ size }) => size > MAX_RAW_JS_CHUNK_SIZE)
+    .sort((a, b) => a.name.localeCompare(b.name))
+}
+
 describe('dashboard first-load JavaScript budget', () => {
   beforeAll(() => {
     buildDashboardDist()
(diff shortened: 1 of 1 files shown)


jo-duchan:
Ran all three acceptance criteria from #524 against your branch instead of reading them off the diff:

| scenario | first-load budget | raw chunk guard | result |
|---|---|---|---|
| current `main` | pass | pass | 2 passed |
| `manualChunks` removed | pass | fail — `index-YWsnsJWL.js: 541990 B` | caught by the new assertion only |
| `@radix-ui` / `react-hook-form` re-split eagerly | fail — 159,621 B | pass | caught by the existing assertion only |

The middle row is really the point of this PR, and it behaves exactly as intended. First-load drops to 145,320 B and comfortably clears the budget, while a 541.99 kB chunk comes back. Neither guard would have caught that on its own, and the third row shows that isn't just an accident of ordering—each one is blind to exactly what the other catches. That's what makes keeping both necessary rather than redundant.

My negative-case numbers match yours to the byte on the removal case and are within 19 B on the eager-split case, which looks like normal source drift between our checkouts rather than any disagreement.

A few things I checked specifically:

- `MAX_RAW_JS_CHUNK_SIZE = 500_000` lines up with Vite's default `chunkSizeWarningLimit`, so the test fails at the same point the warning appears instead of at an arbitrary nearby value.
- Walking the emitted assets with `readdirSync` instead of relying on a hardcoded chunk list means any new oversized chunk gets caught automatically.
- Throwing when `dist/assets` is missing instead of treating that as "no oversized chunks." Same discipline as the guard you added in #520, and exactly the failure mode that would have made this test decorative.
- The failure message includes both the chunk name and its size, so whoever breaks it knows where to look without reproducing it locally.

CI is green across both Node versions and both build architectures. The only red is the Vercel check, an unrelated authorization issue on this repo.

Nothing to change here. Thanks for turning the follow-up around so quickly, and for checking both negative cases yourself before pushing. That made this a one-round review.

Fhatu12:
Thanks for merging this one! Nice to close the loop from #520 and get both regression cases covered properly. Appreciate it.

Issue #561, opened 2026-08-14
changeset audit: report merges that shipped source without touching the root CHANGELOG


jo-duchan:
The new CHANGELOG gate runs at PR time, so it cannot help with anything already on `main` — the same limitation `--audit` exists for on the changeset side, and for the same measured reason: four merged PRs once got as far as release preparation with no changelog entry between them.

Three paths reach `main` today with a user-facing change and no CHANGELOG demand:

- a **bot PR** — the CI `changeset` job is skipped for bots and a skipped required check counts as passing, so nothing in the chain asks. The `### Security` entries already in `CHANGELOG.md` are exactly this class;
- a branch using the `<!-- no-changeset: reason -->` opt-out, where the reason answers the changeset question and nothing asks the changelog one;
- a release branch, which is exempt by design and where `/release` owns the section rename instead.

Extend `--audit` to report merges whose diff shipped published source and did not touch `CHANGELOG.md`, with the same `Backfills: #NNN` escape the changeset audit already understands. `/release` runs the audit, which is where this belongs.

Found by the adversarial review on the backfill-and-gate branch.

Issue #560, opened 2026-08-14
changeset gate: the changelog: internal marker is honoured inside fenced code blocks


jo-duchan:
`changelogEntryOwed` in `scripts/check-changeset.mjs` regexes the raw changeset body with no markdown awareness, so a marker quoted inside a ``` fence silently releases a user-facing change from the CHANGELOG demand.

It is reachable rather than theoretical: the natural changeset for a follow-up to the gate itself quotes the marker to document it.

Every other near-miss fails in the safe direction — different casing, an indented marker in a list or blockquote, trailing whitespace or CRLF, the marker split across two lines all fail to match and ask for an entry that may already have been written. Only the fence, and a marker inside the `---` frontmatter (self-limiting, since `changeset version` would reject the YAML), go the dangerous way.

Strip fenced regions before matching, or require the marker in the last non-empty paragraph.

Found by the adversarial review on the backfill-and-gate branch.

PR #562, opened 2026-08-14
fix: backfill the root CHANGELOG, and give it the gate the per-package ones never needed

jo-duchan:
## Summary

Two changelogs exist and only one had enforcement. The per-package files are written by `changeset version` and cannot be forgotten. The root `CHANGELOG.md` is hand-written, is what a self-hoster reads to decide whether to upgrade, and nothing checked it — so it had been untouched for four days across **22 merged PRs, every one of which carried a changeset**. The rule was in CONTRIBUTING and the enforcement was in neither place.

**One of the missing entries was breaking.** `node-22-floor` — "Requires Node.js ≥ 22" — landed after the v0.18.0 tag with nothing in the release notes. A self-hoster on Node 20 would have upgraded straight into it. The review then found four more upgrade costs that never reached `### Breaking Changes`, the sharpest being the correlator: an agent from before this release does not echo it, so a dashboard install sits on "Installing…" with no Launch control on a mixed-version install.

`## [Unreleased]` goes from 3 entries to 13, grouped by what a user experiences rather than one line per changeset — the six input-truthfulness changesets are one story and read as one.

The gate is a branch of the existing check, not a new script: a branch that touches a changeset and ships published source must touch `CHANGELOG.md`, unless the changeset opts out on a line of its own with `<!-- changelog: internal — reason -->`. What it does **not** cover is stated rather than implied: bot PRs, `no-changeset` branches, release branches, and whether the entry is any good.

## Checklist

- [x] Tests written and passing — 4 unit + 5 integration driving the real CLI against a throwaway repo; 303 in the scripts suite; 6 mutations run, all killed
- [x] No `any`
- [x] Interface changes land in `agent-core` first — n/a

## Related `.work/` docs

`.work/reviews/docs__changelog-backfill-and-gate.md`.

Worth naming from it: the backfill reproduced the failure it exists to end — #533 had no entry, and a file that reads as complete is one nobody re-derives. And the input-truthfulness entry **invented a third disposition** ("nothing reached the device") which is the exact wording the fix it summarises removed as wrong; a reader told the tool can say that would retry an input that may have landed.

Deferred: **#560** (the marker is honoured inside a fenced code block) and **#561** (extend `--audit`, since a PR gate cannot help with what is already on main).

## Summary by CodeRabbit

- **Documentation**
  - Documented breaking changes, including Node.js 22 support and synchronized agent/relay upgrades.
  - Added release notes for improved session diagnostics, request tracking, input failure reporting, boot readiness, screenshot handling, flow-runner errors, iOS video labeling, and process recovery.

- **Chores**
  - Strengthened release checks to ensure applicable changes are reflected in the root changelog.
  - Added validation for internal-only changes and changelog exceptions.


jo-duchan:
Both actionable findings are fixed in `f36dff8`, and the third is declined with a reason.

**The success path was half-widened — fixed.** Widening the CHANGELOG obligation to `AMR` and leaving the success path on `added` meant a branch that amended an existing changeset *and* wrote its entry satisfied the demand and then fell through to "adds no changeset". That is a false failure, which is the direction that teaches people a gate is noise. Both now read `touchedChangesets`, and an integration test covers the amended-and-satisfied case.

**A deletion counted as writing — fixed.** `changed` comes from a bare `--name-only`, which lists deletions like any other path, so removing `CHANGELOG.md` satisfied the demand to write in it. The check asks for an added or modified status now, with a test that deletes the file and expects exit 1.

Both mutations confirmed: reverting the success path kills `passes an amended changeset that does write its entry`; widening the filter to `AMD` kills `does not accept deleting the CHANGELOG as writing an entry`.

**Declined: removing "7.6 seconds".** It is measured, not asserted. `.changeset/ios-boot-readiness.md` names the device and the method — `xcrun simctl bootstatus` reported `Finished` 7.6 seconds after `boot` had returned, on an iPhone 17 Pro / iOS 26.5 — and `IOSAgent.ts:628` and `SimctlWrapper.ts:139` both carry the figure beside the code that waits for it. The number is what makes the entry actionable: "announced ready early" invites a shrug, "7.6 seconds early" explains why an install issued straight after a boot failed.

Worth noting the contrast, since it is the same class of check landing on opposite sides: an earlier review had me drop "the relay knew within a second" from the neighbouring entry, and that was right — no changeset stated a timing and I had invented it.

Issue #352, opened 2026-07-02
Publish relay as a Docker image (Docker Hub) + document Docker deployment


jo-duchan:
## Context

A `Dockerfile` already exists at the repo root — it builds the relay (plus the dashboard) into a runtime image (`EXPOSE 4000`, `/app/.tapflow/data` volume). So the **self-build path already works** (`git clone` + `docker build` + `docker run`).

What's missing is the **polished distribution tier**: a published image + user docs so a team can start with a one-liner instead of cloning and building.

> Note: this is relay-only (a single Node process). The iOS/Android agents need macOS and are out of scope.

### Where the container is meant to run

Docker does not conflict with tapflow's self-hosted, no-external-cloud ethos — a registry is only a distribution channel, and users run the image on their own infrastructure. But the container belongs **on the same Mac, or on a box on your own network — not on a public cloud VM.** Every video and audio frame between an agent and a browser passes through the relay (`packages/relay/src/RelayServer.ts`), so a relay on a cloud VM sends app screen data outside your network, and it inserts a detour the tap-to-glass latency goal cannot absorb. An internal fly.io deployment was used at one point purely to check that the image boots; it is not a recommended topology, and the docs below should say so plainly.

The docs also have to cover the topology this image actually unlocks: **relay on an always-on LAN box, Macs joining as agents.** That path is currently undocumented and pulls in `publicUrl`, TLS and auth. It should be exercised for real before the Compose docs are written, because that is where the remaining unknowns are.

## Scope

- [x] GitHub Actions workflow (`.github/workflows/docker-publish.yml`) — build + push a **multi-platform** image (`linux/amd64`, `linux/arm64`) on native runners. Publishes `edge` and `sha-*` from `main`, semver tags from `v*`. Extended with a runtime smoke test in #565.
- [x] Docker Hub repository — `tapflow/tapflow`, public.
- [ ] Docker Hub image README (the repository overview page).
- [ ] Docker Compose section in `guide/self-hosting.md` (EN + KO), added once the image is published (`docker pull tapflow/tapflow` one-liner) and covering the relay-only scope and the placement rule above.
- [ ] Secrets setup guide (`DOCKERHUB_USERNAME`, `DOCKERHUB_TOKEN`) — for anyone publishing from a fork.
- [ ] Verify the first real publish (below).

## Prerequisite (maintainer) — done

[...]

PR #565, opened 2026-08-15
ci(docker): smoke test relay image runtime

Fhatu12:
## Summary

Docker CI previously proved the relay image could build, but not that the resulting image could boot. This PR progresses #352 by smoke-testing the exact image produced for each architecture: validation-only builds load the local architecture image, and credentialed publish builds smoke-test the pushed digest before manifest assembly.

The smoke run mounts /app/.tapflow/data on a disposable Docker volume, checks the Tapflow dashboard response and /api/v1/auth/status, and cleans up the container/volume on success or failure. .dockerignore now excludes stale *.tsbuildinfo files because dist is also excluded from the Docker context, and stale incremental metadata can suppress required TypeScript emits during clean image builds.

Both linux/amd64 and linux/arm64 remain required for #352 and are structurally guarded by the scripts regression test. Registry selection/publication, Docker Hub README, EN/KO Compose docs, and credential documentation remain outside this PR and still belong to #352.

Remaining uncertainty: local runtime evidence is amd64 only; GitHub native arm64 runners will provide real arm64 evidence, and Docker Hub credentialed publish/manifest behavior cannot be fully exercised until maintainer registry credentials/resources exist.

## Checklist

- [x] No `any`

## Validation

- No-cache Docker build: `docker build --no-cache -t tapflow-qa-352-runtime-smoke:local .` passed.
- Real amd64 Docker Desktop runtime passed: dashboard HTTP response contained `<title>tapflow</title>`, `/api/v1/auth/status` returned initialized-state JSON, JWT secret persisted across restart without revealing the secret, and forced-failure cleanup removed the test container/volume.
- Architecture mutation tests: removing `linux/arm64` failed the focused structural test; removing `linux/amd64` failed the focused structural test; both mutations were restored.
- Focused regression test: `pnpm exec vitest run --config scripts/vitest.config.mjs scripts/__tests__/dockerPublishSmoke.test.mjs` passed, 6 tests.
- Workflow lint: `go run github.com/rhysd/actionlint/cmd/actionlint@latest .github/workflows/docker-publish.yml` passed.
- Scripts suite: `pnpm exec vitest run --config scripts/vitest.config.mjs` passed, 24 files / 319 tests.
- Workspace gates already established during QA/pre-commit: `pnpm typecheck` passed; `pnpm lint` exited 0 with existing warnings only; `pnpm test` passed end-to-end.
- `git diff --check origin/main...HEAD` passed.
- Secret scan of the committed diff found no credential material.
- Independent QA verdict before this repair was PASS WITH NOTES; the noted arm64 structural-test gap is now addressed.

## Related `.work/` docs

- `.work/reviews/fix__docker-runtime-smoke.md`

## Summary by CodeRabbit

* **Chores**
  * Improved Docker image handling for pull requests and builds without publishing credentials.
  * Excluded unnecessary TypeScript build metadata from Docker build contexts.
  * Enhanced container validation with readiness checks, endpoint verification, diagnostics, cleanup, and configuration persistence checks.

* **Tests**
  * Added automated coverage for supported platforms, image selection, container startup, application endpoints, configuration persistence, cleanup, and excluded build artifacts.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,8 @@
 node_modules
 **/node_modules
 **/dist
+# Package-root tsconfig.tsbuildinfo sits outside **/dist; stale incremental metadata
+# can make a local pnpm build -> docker build skip required emits.
 **/*.tsbuildinfo
 **/.tapflow
 playground
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,6 +7,7 @@
 node_modules
 **/node_modules
 **/dist
+**/*.tsbuildinfo
 **/.tapflow
 playground
 !playground/package.json
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -70,9 +70,74 @@ jobs:
           cache-from: type=gha,scope=${{ env.PLATFORM_PAIR }}
           cache-to: type=gha,mode=max,scope=${{ env.PLATFORM_PAIR }}
           # Publish: push a per-architecture image by digest (merged into a manifest
-          # list in the merge job). Otherwise (PR / no credentials): build for
-          # validation only, keeping just the cache.
-          outputs: ${{ (github.event_name != 'pull_request' && env.HAS_DOCKERHUB == 'true') && format('type=image,name={0},push-by-digest=true,name-canonical=true,push=true', env.IMAGE) || 'type=cacheonly' }}
+          # list in the merge job). Otherwise (PR / no credentials): load the
+          # built image into Docker so the runtime smoke test exercises it.
+          outputs: ${{ (github.event_name != 'pull_request' && env.HAS_DOCKERHUB == 'true') && format('type=image,name={0},push-by-digest=true,name-canonical=true,push=true', env.IMAGE) || format('type=docker,name=tapflow-smoke:{0}', env.PLATFORM_PAIR) }}
+
+      - name: Select image for smoke test
+        run: |
+          if [ "${HAS_DOCKERHUB}" = "true" ] && [ "${GITHUB_EVENT_NAME}" != "pull_request" ]; then
+            echo "SMOKE_IMAGE=${IMAGE}@${{ steps.build.outputs.digest }}" >> "$GITHUB_ENV"
+          else
+            echo "SMOKE_IMAGE=tapflow-smoke:${PLATFORM_PAIR}" >> "$GITHUB_ENV"
+          fi
+
+      - name: Smoke test image runtime
+        run: |
+          set -euo pipefail
+
+          container="tapflow-smoke-${PLATFORM_PAIR}-${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          volume="${container}-data"
+          root_body="${RUNNER_TEMP}/tapflow-smoke-root.txt"
+          api_body="${RUNNER_TEMP}/tapflow-smoke-auth-status.txt"
+
+          cleanup() {
+            status=$?
+            if [ "$status" -ne 0 ]; then
+              docker logs "$container" 2>/dev/null || true
+            fi
+            docker rm -f "$container" >/dev/null 2>&1 || true
+            docker volume rm "$volume" >/dev/null 2>&1 || true
+            exit "$status"
+          }
+          trap cleanup EXIT
+
+          docker volume create "$volume" >/dev/null
+          docker run -d \
+            --name "$container" \
+            -p 127.0.0.1:4000:4000 \
+            -v "$volume:/app/.tapflow/data" \
+            "$SMOKE_IMAGE" >/dev/null
+
+          for attempt in $(seq 1 60); do
+            if [ "$(docker inspect -f '{{.State.Running}}' "$container")" != "true" ]; then
+              echo "Container exited before readiness (attempt ${attempt})." >&2
+              exit 1
+            fi
+
+            root_status="$(curl -sS -o "$root_body" -w '%{http_code}' http://127.0.0.1:4000/ || true)"
+            api_status="$(curl -sS -o "$api_body" -w '%{http_code}' http://127.0.0.1:4000/api/v1/auth/status || true)"
+            if [ "$root_status" -ge 200 ] && [ "$root_status" -lt 400 ] \
+              && [ "$api_status" -ge 200 ] && [ "$api_status" -lt 400 ] \
+              && grep -Fq '<title>tapflow</title>' "$root_body" \
+              && grep -Eq '"initialized":(true|false)' "$api_body"; then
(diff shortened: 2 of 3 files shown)


jo-duchan:
Thanks — this closes the gap that mattered in #352. Before this, the image could build, but nothing actually proved it booted.

I read through the run on `a75d3d8`, and both legs exercised the full path on native runners (`/`=200, `/api/v1/auth/status`=200, and the JWT secret stayed stable across recreation). Putting the smoke behind `needs: build` is exactly the right place for this check.

A couple of things that are easy to miss:

- `**/*.tsbuildinfo`. I verified these live at `packages/*/tsconfig.tsbuildinfo`, at the package root rather than under `dist`, so `**/dist` never covered them. CI never runs into this, but anyone doing `docker build` after a local `pnpm build` does, which is exactly the self-build path #352 is meant to protect. Could you add that explanation as a comment next to the line? Right now it only exists in the PR description, which makes it easy for someone to remove later without realizing why it's there.
- Moving PR builds from `type=cacheonly` to `type=docker`, so the smoke exercises the actual artifact instead of rebuilding it.

Just two things I'd like to see before merging.

### 1. Restore container logs on failure

I disagree with the CodeRabbit suggestion here. I checked what the relay actually logs during startup (`packages/relay/src/server.ts`): a port message, a TLS notice, `Loaded credentials from <path>` (the path only), and `relay failed to start: <err>`. The JWT secret is read into memory but never logged, and the CI container is created fresh with no user data.

On the other hand, when the smoke test fails today all we get is `State.Status`, `State.ExitCode` and `State.OOMKilled`, which makes "the image doesn't boot" very hard to diagnose from CI alone.

I'd keep `docker logs --tail 50 "$container"` on the failure path only, alongside the existing `docker inspect` output.

### 2. Trim `scripts/__tests__/dockerPublishSmoke.test.mjs`

Could you take another look at [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md)? Sections 2 and 3 apply pretty directly here.

Most of this file mirrors a shell script that's already executed on every PR, so regressions there already show up as a failing workflow. Assertions like `expect(smoke.match(/start_container/g)).toHaveLength(3)` fail on a harmless rename, and `not.toContain('docker logs "$container"')` is exactly the absence-assertion pattern discussed in §2. I like the wording in §3: a spelling assertion is a floor, not a fence.

I'd keep only the assertions that protect regressions CI can't catch:

- **The platform matrix test.** If `linux/arm64` disappears, CI still passes while we silently publish a single-architecture image. You already mutation-tested both directions, which is exactly what §1 recommends.
- **The `.dockerignore` assertion.** It's inexpensive, and the regression is invisible in CI for the reason above.
- **The publish-path assertion** (`SMOKE_IMAGE=${IMAGE}@digest`), with a short header explaining that this branch can't be exercised in CI until registry credentials are available. With that explanation it becomes clear why the assertion exists, instead of looking like a mirror of the script.

Drop the rest.

Also, could you add a short header comment in the same style as `scripts/__tests__/ciGateCoverage.test.mjs`? I like how that file opens by naming the specific regression it protects against, and I think that's the right bar for workflow tests here too.

Once those two land I'll merge this.

Separately, and not for this PR: the smoke currently proves the process starts, serves the dashboard HTML and answers one JSON endpoint, while the relay's real job is the WebSocket hub. I'll open a follow-up issue for a handshake check.

jo-duchan:
Follow-up filed as #566 — the handshake check, kept separate from this PR.

jo-duchan:
Thanks — the trim is exactly what I was after. And the regex change is a real improvement I hadn't asked for: `(\S+)` catches a platform that gets *replaced*, whereas the old `(linux\/(?:amd64|arm64))` would have matched nothing and reported nothing.

I re-ran the arm64 mutation myself instead of taking the description at face value, and the guard still does its job:

```
expected [ 'linux/amd64' ] to deeply equal [ 'linux/amd64', 'linux/arm64' ]
```

Both build legs on `67b79ac` reach first boot and restart with `/=200` and `/api/v1/auth/status=200`, and the JWT secret stays stable across recreation. Putting the digest explanation in the file header rather than beside the assertion works fine for me.

One thing left before I merge.

### `2>/dev/null` throws away the very line we restored the logs for

```sh
docker logs --tail 50 "$container" 2>/dev/null || true
```

The placement is right: it's only on the failure path, uses `--tail 50`, and sits next to `docker inspect`. The problem is that the redirection drops the container's stderr, which is exactly where the boot failure is logged.

- `packages/relay/src/server.ts:53-56` — the top-level catch calls ``logger.error(`relay failed to start: ${String(err)}`)`` and exits with status 1.
- `packages/agent-core/src/logger.ts:51` — `logger.error` is `console.error`, so it writes to stderr.
- The container is started with `docker run -d` and no `-t`, so stdout and stderr stay separate, and `docker logs` writes the container's stderr to its own stderr.

So on a failed boot, we keep the `logger.info` lines on stdout (`tapflow relay running on port …`) but discard the one line explaining why the process died. That's worse than having no logs at all: it makes the startup look healthy.

`|| true` already absorbs the exit status, and the "container was never created" case is already covered by `diagnose_container`, which prints `Container exists: false` on the line above. There's nothing left to suppress here. I'd mirror the `diagnose_container >&2` call beside it:

```sh
docker logs --tail 50 "$container" >&2 || true
```

### Two small, non-blocking nits if you're pushing another commit anyway

- The `.dockerignore` test is still titled `does not copy stale TypeScript build metadata without matching dist output`, but the `**/dist` half of the assertion is gone. The title now promises more than the test actually checks.
- The header says which checks survived, but not why the platform check matters. Something like "CI stays green while we publish a single-architecture image" makes it clear that the assertion is protecting a real behavior rather than just mirroring the workflow, just like the digest note explains why it exists.

Once the redirection is fixed, I'll merge.

jo-duchan:
Merging this.

`60acb15` is what I asked for — the container's stderr is back in the log, and both build legs are green. I re-ran the arm64 mutation to be sure the rewritten header hadn't blunted the guard along the way; it still fails on the missing leg.

Third one from you, and the pattern holds: #520 fixed the bundle split, #525 guarded it, this one does both for the Docker image. Widening the regex while trimming the file was a nice call — I only asked for the trim.

Thanks.

Fhatu12:
Thanks Jo — really appreciate the careful review and the merge.

On the regex change, once I was trimming the structural tests I wanted the few guards that remained to protect the actual regression rather than the exact workflow spelling. The platform check felt especially worth strengthening because a single-architecture publish could otherwise stay green and be easy to miss. Good to hear the arm64 mutation still caught it independently.

I’ve enjoyed working through these three contributions with you. Thanks again for the detailed feedback along the way.

PR #580, opened 2026-08-17
test: smoke-check relay WebSocket upgrade

### LINKED ISSUE #566: Smoke-test the relay's WebSocket upgrade in the Docker image build
## Context

#565 added a runtime smoke test to `.github/workflows/docker-publish.yml`. It boots the built image on both architectures, checks that `/` serves the dashboard HTML and that `/api/v1/auth/status` answers, and confirms the JWT secret survives container recreation.

That proves the process starts and the HTTP surface works. It does not touch the relay's actual job. `RelayServer` attaches a `WebSocketServer` to the same HTTP server (`packages/relay/src/RelayServer.ts:249`), and every frame between an agent and a browser goes through it. An image whose upgrade path is broken while static files and REST still answer would publish green and be useless to whoever pulls it.

## Scope

Add one step to the smoke, after the existing readiness loop:

- [ ] Open a WebSocket to `ws://127.0.0.1:4000/` against the running container and assert the handshake completes (HTTP 101).
- [ ] Assert the relay then closes it with code `1008` (see below).
- [ ] Fail the step, and therefore the `merge` job, when either does not happen.

## The expected result is 101 followed by a 1008 close, not a connection that stays open

This is the part worth getting right before writing the check.

The relay puts no path restriction on the upgrade, so the root path is enough to reach `handleConnection` — no session or auth setup needed. But the container's port is published with `-p 127.0.0.1:4000:4000`, so the connection arrives at the container from the Docker bridge gateway, not from loopback. `resolveClientAddress` classifies it with `isLoopback` (`packages/relay/src/lib/clientAddress.ts:56`), so `isLocal` is false; with no cookie and no PAT, `classifyConnection` returns `reject` and the socket is closed with `1008` (`RelayServer.ts:483`).

That is correct behaviour, and it makes the check stronger rather than weaker: `101` proves the upgrade path is reachable in the packaged image, and `1008` proves the auth gate is live in it. Asserting both is a positive claim about a specific code, which is what we want.

**Do not "fix" the relay to keep the connection open so the check passes.** That would be a real auth regression, and it is the most likely wrong turn here.

## Out of scope

A full agent registration or a session join. Both need a real simulator or emulator, so both need macOS, and the smoke runs in a Linux container on a GitHub runner. The claim is narrow on purpose.

## Notes

- Keep it dependency-free if you can. The runner has Node, so a short `node -e` against the published port avoids adding a tool to the workflow.
- If this gets a companion check under `scripts/__tests__/`, read [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) first. The review on #565 asked for `dockerPublishSmoke.test.mjs` to be cut back to the assertions that guard a regression CI cannot show, and anything new should meet the same bar.
- Related: #352 (publishing the image), #565 (the smoke this extends).

jvo34:
## Summary

Adds the WebSocket handshake coverage requested in #566 to the existing Docker image runtime smoke test.

The new dependency-free probe:

- opens a WebSocket upgrade request against the running relay over the existing published-port path
- verifies the HTTP upgrade completes with `101`
- handles any WebSocket bytes delivered in the upgrade `head`
- verifies the relay immediately sends a close frame with code `1008` for the unauthenticated connection
- fails with a clear error if the upgrade fails, the close frame is missing, or the close code is unexpected
- uses a short timeout so the smoke step cannot hang

No relay behavior was changed.

I also added a small structural guard in `dockerPublishSmoke.test.mjs` to protect the upgrade/head handling and `1008` close assertion from being accidentally removed.

## Validation

- `pnpm exec vitest run --config scripts/vitest.config.mjs scripts/__tests__/dockerPublishSmoke.test.mjs`
  - 4/4 tests passed
- Local Docker image built successfully
- Existing HTTP smoke checks passed:
  - `/` → 200
  - `/api/v1/auth/status` → 200
- Local WebSocket probe passed:
  - HTTP upgrade → 101
  - unauthenticated close → 1008
- JWT secret file remained present and non-empty
- `git diff --check` passed

## Checklist

## Related `.work/` docs

N/A

Closes #566

## Summary by CodeRabbit

* **Tests**
  * Enhanced Docker runtime smoke testing to verify WebSocket connection upgrades.
  * Added validation for successful handshakes, valid server close frames, and the expected unauthenticated close code.
  * Improved detection and reporting of invalid responses, connection failures, premature closures, and timeouts.
  * WebSocket checks now run during container startup verification before persistence checks.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -39,6 +39,10 @@ jobs:
         with:
           fetch-depth: 0
 
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+
       - name: Warn when Docker Hub credentials are missing
         if: github.event_name != 'pull_request' && env.HAS_DOCKERHUB != 'true'
         run: echo "::warning::DOCKERHUB_USERNAME/DOCKERHUB_TOKEN are not set — building the image for validation but skipping push."
@@ -163,98 +167,7 @@ jobs:
           start_container
           wait_for_ready "first boot"
 
-          node <<'NODE'
-          const http = require('http')
-          const crypto = require('crypto')
-
-          const expectedCode = 1008
-          const timeoutMs = 5000
-          let done = false
-          let upgraded = false
-          let request = null
-          let socket = null
-          let buffer = Buffer.alloc(0)
-
-          const finish = (ok, message) => {
-            if (done) return
-            done = true
-            clearTimeout(timer)
-            if (request) request.destroy()
-            if (socket) socket.destroy()
-            if (ok) {
-              console.log('WebSocket smoke probe passed: upgrade status=101 and close code=1008.')
-              process.exit(0)
-            }
-            console.error(`WebSocket smoke probe failed: ${message}`)
-            process.exit(1)
-          }
-
-          const consume = (chunk) => {
-            if (done) return
-            if (chunk && chunk.length > 0) buffer = Buffer.concat([buffer, chunk])
-            if (buffer.length < 2) return
-
-            const opcode = buffer[0] & 0x0f
-            if (opcode !== 0x8) return finish(false, `expected first WebSocket frame opcode 0x8 (close) but got 0x${opcode.toString(16)}`)
-
-            const b1 = buffer[1]
-            const payloadLength = b1 & 0x7f
-            if ((b1 & 0x80) !== 0) return finish(false, 'unexpected masked server close frame')
-            if (payloadLength > 125) return finish(false, `close frame payload length ${payloadLength} is invalid for a control frame`)
-            if (buffer.length < 2 + payloadLength) return
-            if (payloadLength < 2) return finish(false, 'close frame missing close code payload')
-
-            const closeCode = buffer.readUInt16BE(2)
-            if (closeCode !== expectedCode) return finish(false, `expected close code ${expectedCode} but got ${closeCode}`)
-            finish(true)
-          }
-
-          const timer = setTimeout(() => {
-            if (!upgraded) return finish(false, `timed out after ${timeoutMs}ms waiting for HTTP 101 upgrade`)
-            finish(false, `timed out after ${timeoutMs}ms waiting for WebSocket close frame with code ${expectedCode}`)
-          }, timeoutMs)
-
-          request = http.request('http://127.0.0.1:4000/', {
-            headers: {
-              Connection: 'Upgrade',
-              Upgrade: 'websocket',
-              'Sec-WebSocket-Version': '13',
-              'Sec-WebSocket-Key': crypto.randomBytes(16).toString('base64'),
-            },
-          })
-
-          request.on('response', (res) => {
-            finish(false, `expected HTTP 101 upgrade but got HTTP ${res.statusCode ?? 'unknown'}`)
-          })
-
-          request.on('upgrade', (res, upgradedSocket, head) => {
-            if (res.statusCode !== 101) {
-              finish(false, `expected HTTP 101 upgrade but got HTTP ${res.statusCode ?? 'unknown'}`)
-              return
-            }
-            upgraded = true
-
-            socket = upgradedSocket
-            socket.on('data', consume)
-            socket.on('end', () => {
-              finish(false, 'connection ended before receiving the expected WebSocket close frame')
-            })
-            socket.on('close', () => {
-              finish(false, 'connection closed before receiving the expected WebSocket close frame')
-            })
-            socket.on('error', (error) => {
-              finish(false, `connection/upgrade failed: ${error.message}`)
-            })
-
-            consume(head)
-          })
-
-          request.on('error', (error) => {
-            finish(false, `connection/upgrade failed: ${error.message}`)
-          })
-
-          request.end()
-          NODE
+          node scripts/docker-publish-websocket-smoke.mjs
 
           first_jwt_secret="$(jwt_secret_fingerprint)"
 
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -162,6 +162,100 @@ jobs:
           docker volume create "$volume" >/dev/null
           start_container
           wait_for_ready "first boot"
+
+          node <<'NODE'
+          const http = require('http')
+          const crypto = require('crypto')
+
+          const expectedCode = 1008
+          const timeoutMs = 5000
+          let done = false
+          let upgraded = false
+          let request = null
+          let socket = null
+          let buffer = Buffer.alloc(0)
+
+          const finish = (ok, message) => {
+            if (done) return
+            done = true
+            clearTimeout(timer)
+            if (request) request.destroy()
+            if (socket) socket.destroy()
+            if (ok) {
+              console.log('WebSocket smoke probe passed: upgrade status=101 and close code=1008.')
+              process.exit(0)
+            }
+            console.error(`WebSocket smoke probe failed: ${message}`)
+            process.exit(1)
+          }
+
+          const consume = (chunk) => {
+            if (done) return
+            if (chunk && chunk.length > 0) buffer = Buffer.concat([buffer, chunk])
+            if (buffer.length < 2) return
+
+            const opcode = buffer[0] & 0x0f
+            if (opcode !== 0x8) return finish(false, `expected first WebSocket frame opcode 0x8 (close) but got 0x${opcode.toString(16)}`)
+
+            const b1 = buffer[1]
+            const payloadLength = b1 & 0x7f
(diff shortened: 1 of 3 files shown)


jo-duchan:
Thanks — this is the coverage #566 was asking for, and the probe itself is careful work.

You avoided the trap I mentioned on the issue, and the green build shows that. If the connection had arrived over loopback, the relay would have accepted it, no close frame would ever have arrived, and the probe would have timed out after 5s. Instead it passed on both build legs with `upgrade status=101` and `close code=1008`, so it really exercised the reject path at `RelayServer.ts:519`.

The frame parsing is solid too: waiting for `2 + payloadLength` before reading the code, rejecting a masked server frame, and treating any first opcode other than a close frame as a failure. `WS_REJECT_REASON` is already kept under 123 bytes by `connectionAuth.test.ts:58`, so the `payloadLength > 125` guard can't accidentally trip.

One thing on the test file before I merge.

### The structural guard mirrors the probe

```js
expect(smoke).toContain("request.on('upgrade', (res, upgradedSocket, head) => {")
expect(smoke).toContain('consume(head)')
expect(smoke).toContain('opcode 0x8 (close)')
expect(smoke).toContain('const expectedCode = 1008')
expect(smoke).toContain('closeCode !== expectedCode')
```

The probe runs on every PR, on both build legs. That changes what a guard over it is actually buying us, so I tried three mutations:

| Mutation | This test | CI | |
|---|---|---|---|
| rename `upgradedSocket` → `sock` (probe still works) | 🔴 | 🟢 | false alarm |
| `expectedCode = 1000` (a real bug) | 🔴 | 🔴 | redundant |
| delete the probe entirely | 🔴 | 🟢 | the one case that earns its place |

Only the third row catches something CI can't already catch. The other two are the cost: a harmless rename makes the suite fail, and a real bug already fails the build without any help from this test. `opcode 0x8 (close)` is probably the clearest example: it asserts on the wording of an error message, so it can still pass after the probe is broken, or fail after a harmless refactor.

Have a look at [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md), sections 2 and 3. This file went through the same exercise on #565 yesterday and ended up with three assertions for the same reason, so you're in good company. It isn't obvious at first.

I'd keep one assertion for the third row, using values fixed by the protocol rather than names chosen in the implementation:

```js
it('keeps the WebSocket upgrade probe in the smoke step', () => {
  const smoke = stepBlock('Smoke test image runtime')
  expect(smoke).toContain('Sec-WebSocket-Key')
  expect(smoke).toContain('1008')
})
```

`Sec-WebSocket-Key` can't disappear without breaking the handshake, and `1008` is the protocol close code. Both survive a refactor, and both disappear if the probe is removed.

While you're in there: the file header currently says it guards two things, a single-architecture publish and the credentialed digest path. If we're adding a third test, it'd be worth adding a third clause so the header still matches the file.

### One question and one nit, neither blocking

The probe is about 90 lines of JavaScript inside a YAML heredoc, and it's now the largest part of that step. Would you rather move it into `scripts/` and have the workflow call it? It would be covered by linting and the usual review tooling, and the guard above could simply check that the workflow invokes the script, which is much harder to break with a rename. I can also see the case for keeping it next to the container it probes, so I'm asking rather than suggesting a change. The heredoc itself looks correct.

The nit: the `build` job doesn't run `setup-node`, so it's using whatever Node version the runner image ships with. That's working today on both images. Adding the step costs a few seconds, but it makes the dependency explicit instead of implicit.

Once the guard is trimmed, I'll merge. Nice first contribution.

jvo34:
@jo-duchan Thanks so much! I really appreciate the detailed review and the explanation behind the guard coverage, the mutation examples made it really clear why most of those assertions are redundant with the probe already running in CI.

I’ll trim the guard down to the protocol-level assertions you suggested and update the header while I’m there.

I also like the idea of moving the probe into scripts/. I think that would be cleaner and easier to maintain than keeping ~90 lines of JS inside the YAML, so I’m happy to make that change too. And I can add setup-node while I’m in there to make the Node dependency explicit.

Thanks again for taking the time to walk through it so thoroughly, especially on my first contribution! Really appreciate it.

jvo34:
@jo-duchan Just pushed the follow-up changes, thanks again for the suggestions.

I moved the WebSocket smoke probe out of the workflow and into scripts/docker-publish-websocket-smoke.mjs, added setup-node using the repo’s .nvmrc, and trimmed the guard test down so it only verifies that the smoke step still invokes the standalone probe rather than duplicating the probe’s implementation details.

I also ran the guard tests, lint, syntax check, and the Docker smoke flow locally. The container came up successfully and the extracted probe completed with the expected HTTP 101 upgrade and 1008 close code.

Hopefully this is much closer to what you had in mind. Really appreciate the detailed feedback!

jo-duchan:
All three landed, and the probe body came across byte-for-byte. Nothing drifted in the move, so the behavior that was already green stayed green.

I re-ran the mutations against the new shape. Renaming `upgradedSocket` inside the script now leaves the suite green, and dropping the `node scripts/docker-publish-websocket-smoke.mjs` line from the step still turns it red. That's exactly the trade I was after. Asserting the invocation instead of the two protocol constants I originally suggested is the better choice now that those constants aren't in the step anymore.

One small thing before I merge.

```js
import { createRequire } from 'node:module'

const require = createRequire(import.meta.url)
const http = require('http')
const crypto = require('crypto')
```

That's left over from the heredoc version, where the script ran as CommonJS over stdin and `require` already existed. In an `.mjs` file it doesn't, so the shim is only there to keep the original two lines working instead of switching them to:

```js
import http from 'http'
import crypto from 'crypto'
```

I tried it that way to be sure. It reaches the connection attempt and fails with `ECONNREFUSED` when there's no relay running, which is exactly what I'd expect with no container up.

This isn't a written rule, but the other six files under `scripts/` all use normal ESM imports, and `createRequire` is really for cases where you need CommonJS resolution: `require.resolve`, for example, or a package without an ESM entry point. Neither applies to two core modules.

I'd rather this file not be the odd one out, since the next script someone writes will probably be copied from whichever file they open first.

Other than that, it's ready. Push that and I'll merge.

jo-duchan:
Hi @jvo34 — no rush on this, I just want to make the state explicit rather than let you guess.

One thing is still on your side before I merge, and it sits in the middle of my last comment where it's easy to miss. Swapping the `createRequire` shim for plain ESM imports:

```diff
-import { createRequire } from 'node:module'
-
-const require = createRequire(import.meta.url)
-const http = require('http')
-const crypto = require('crypto')
+import http from 'http'
+import crypto from 'crypto'
```

That is the whole change. Everything else is ready.

On CodeRabbit's FIN-bit comment: don't let it hold you up. RFC 6455 forbids fragmenting control frames, so a close frame without FIN would be the server violating the protocol — and this probe only ever talks to our own relay, which sets it. It's a smoke probe for one reject path, not a general WebSocket client. I'd leave it.

No deadline. And if you'd rather not carry it further, just say so — no hard feelings, and your commits keep your name on them either way.

jvo34:
@jo-duchan Sorry for the delayed response. I’ve been traveling and mostly offline the last few days, and I apologize for leaving you hanging.

I just pushed the final change swapping the createRequire shim for the plain ESM imports you suggested. I re-ran the syntax check, targeted smoke tests, and lint locally after the change, and everything is still green.

Also, thank you for re-running the mutations and confirming the new guard behaves the way we intended. I really appreciate you taking the time to verify the move and explain the tradeoff there. And thanks for clarifying the CodeRabbit FIN-bit comment as well, I’ve left that behavior unchanged as you recommended.

Thanks again for all the detailed feedback and patience throughout this! I’ve learned a lot from working through your review, especially around keeping the guard focused on the behavior CI actually needs to protect.

jo-duchan:
No apology needed at all, @jvo34 — being away for a few days is completely normal, and there was never any rush on this one. Thanks for coming back to it.

I checked d2b43af, and the imports look good. CI is green across both Node versions and both build architectures, so I'll merge it.

Nice work on this. Pulling the probe out of the heredoc without changing its behavior is the kind of refactor that's easy to get subtly wrong, and it made the guard discussion much simpler afterward. The final guard is better than the version I originally had in mind — asserting the invocation was the right choice.

Thanks for sticking with the review, too. This was a solid first contribution to tapflow, and I'd be happy to see you around again.

One last thing, completely optional: if you find the project useful, a ⭐ would really help. Either way, thanks again! 🙏

jvo34:
Thank you so much, @jo-duchan! I really appreciate all the time you put into reviewing this and explaining the reasoning behind your feedback. I learned a ton working through it, and I’m really happy with where the final implementation ended up.

I’d definitely love to contribute to tapflow again! Thanks for making my first contribution such a positive experience. And absolutely — ⭐ added! 😊

Thanks again!

Issue #581, opened 2026-08-17
Document the ?client= handshake parameter in the wire contract


jo-duchan:
The `?client=` handshake parameter is a wire contract and is documented only in the relay's own AGENTS.md.

A session's owner is now `<userId>:<clientId>`, where `clientId` comes from a query parameter on the WebSocket upgrade. Three in-repo clients send it and the relay reads it, but `packages/protocol/AGENTS.md` — the single source of truth for what travels between browser, relay and agent — does not mention it. A client written outside this repo has no way to learn that omitting it drops them to per-connection ownership, or that sending a stable value lets a reconnect resume its own sessions.

Deciding where it belongs is the work: that file currently owns *message* unions, and a handshake-level parameter is a different kind of contract. Either it grows a section for the connection itself, or the relay's file is declared the owner and the protocol file points at it. Found by the pre-PR review on the slice that added the parameter.

Issue #579, opened 2026-08-16
relay: a browser the heartbeat has already lost still holds its session, so the same tester's reconnect is refused


jo-duchan:
A browser that has vanished still holds its session for up to a heartbeat window, so the same tester reconnecting is refused for holding it themselves.

`SessionManager.join()` refuses when `session.browserSocket?.readyState === WebSocket.OPEN`. The relay's own liveness signal is separate: `HEARTBEAT_MS` is 30s and a socket is terminated only after it misses a pong, so a browser that went away — a sleeping laptop, a Wi-Fi blip, both documented as routine in `DeviceViewer`'s comments (#439) — leaves `readyState` OPEN on the relay for 30–60 seconds. `useRelay` retries after 2s, re-sends `session:start`, and is refused. Frames keep going to the dead socket meanwhile, and if the tester does not get back before `clearBrowser`'s idle timer fires, the relay asks the agent to shut the device down.

This is item 2 of #507. Items 1 and 3 shipped — the ownership check in #499's slice, and the one-slot reverse index in #571, which turned out to be a different defect from the one that issue described. #507 was closed with this half undone, so it is filed on its own rather than left in a closed issue's body.

**It is the same decision as #527**, which is why it is not simply "consult `wsAlive`". Liveness narrows the window; it cannot recognise the tester coming back, because a reconnect is a new socket. What does is a notion of owner that survives one — a tab-scoped client id — and that is also what unblocks `device:shutdown`'s missing ownership check, #567's "success answers the holder, failure answers the requester" split, and #574. The four close together or not at all.

jo-duchan:
Closed by #582. The fix is the one this issue predicted — not `consult wsAlive`, but an owner that survives a socket, plus liveness that a reconnect cannot be refused by.

- **Liveness is a pong timestamp**, not the swept boolean. The flag read every healthy socket as dead for the length of a ping round trip, and that window is longest for the holder — video goes out on that socket and the ping queues behind it. Occupancy and `busy` now take the same predicate, so a device can no longer show free while a join for it is refused.
- **A session is owned by a client, not a socket** (`<userId>:<clientId>` from the `?client=` handshake parameter), so the same tab reconnecting is recognised rather than told it holds the device itself.

`device:shutdown`'s missing ownership check went with it (#527). #567 and #574 were the other two named here and are still open — they are answerable now that the owner exists.

Issue #574, opened 2026-08-16
mcp-server: disconnectDevice cancels waiters for a session the relay never let it leave


jo-duchan:
`TapflowClient.disconnectDevice` cancels a session's waiters whether or not the relay accepted the leave, so a command the relay still runs is reported as ended by a leave that never happened.

The relay drops `session:leave` from a socket that does not own the session — `if (this.ownsSession(ws, …))` with no else and no reply. `send` succeeding says only that the frame left this process, so the client settles the waiters regardless.

Harmless for every ownership-gated command, because those are refused by `dispatchTarget` within a round trip and no waiter is left alive to cancel. **`device:shutdown` is the exception**, deliberately: it routes through `reachableTargetWithoutOwnership` and carries no ownership clause at all until #527. So: a dashboard tester holds `s1`; the MCP client never joined it but has the id from `list_devices`; `shutdown_device` is forwarded and the agent powers the device off; `disconnect_device` arrives before `device:shutdown-done`; the relay discards the leave, the client rejects the shutdown as ended by its own leave, and the real reply is then dropped for matching no waiter. The device is off and the model is told the shutdown failed, with a cause that did not occur and advice (`connect_device` again) that will be refused for as long as the tester holds the device.

The honest fix needs something the client does not have: a record of which sessions it actually holds, set on `session:joined` and cleared on leave or terminate — and "joined" is ambiguous while a join is in flight. #527 closing the shutdown ownership hole would also close this, so the two are worth deciding together. Found by the pre-PR review on the client-waiter slice.

PR #582, opened 2026-08-17
A session belongs to a client, and occupancy is judged by liveness

### LINKED ISSUE #527: relay, dashboard: device:shutdown is the one session command with no ownership check, and the blocker is that useAgentSession never joins

jo-duchan:
## Summary

Two defects met on the same two lines — `SessionManager.join()` and `handleSessionStart` — so this is one change rather than two. Closes #527, #579 and, as a consequence, #574.

- **#527** the relay decided who may command a session from a *socket*, but a browser tab holds four of them and the one holding the session is not the one that sends the teardown. A session is owned by `<userId>:<clientId>` now, from a `?client=` parameter at the handshake, minted per connection when absent — so there is no fallback branch and an unidentified client keeps the ownership it had.
- **#579** occupancy read `readyState`, which is not a liveness signal: a sleeping laptop holds a session for up to a minute. It reads the heartbeat's pong record instead, as a **timestamp** rather than the swept boolean — that flag called every healthy socket dead for a ping round trip, longest for the holder, whose socket also carries the video.

`device:shutdown` takes a weaker gate: refused when another client holds the session, permitted when nobody does. Not "owned by me", because the unmount teardown races its own viewer socket's close on a different connection.

Reasoning is beside the code; the release note is in `.changeset/session-owner-is-a-tab.md`.

## Checklist

- [x] Tests written and passing — 2679 across the monorepo, 24 mutations run
- [x] No `any`
- [x] Interface changes land in `agent-core` first — n/a; the handshake parameter is documented in `relay/AGENTS.md` and its home in the wire contract is #581

## Related `.work/` docs

- Plan: `.work/2026-08-17-session-owner-is-a-tab-plan.md`
- Review: `.work/reviews/feat__session-owner-is-a-tab.md` — five channels: two design lenses that **killed the first plan**, one on the four decisions that replaced it, and two on the diff

Worth reading. The pre-PR review found `crypto.randomUUID` in the dashboard — secure-context only, and a LAN deployment is plain HTTP. `lib/requestId.ts` records that exact lesson from the first time it happened and predicts the second in writing. Because the call sat at module scope in a file every route loads, it would have been a blank page rather than a broken socket, and it was green everywhere it could be run: dev is localhost, vitest is jsdom. A static check now makes a third impossible.

It also found that the new gate would have stopped a client that sends no id from shutting down its own device — deterministically, not as a race — which is the cost #527 was filed to avoid, reintroduced by #527's fix.

Two things in this slice came from review conversation rather than a lens: that a per-document identity is not the *root* fix for #579 (the root is that `readyState` is not liveness), and that splitting the two halves would have meant fixing the same two lines twice.

Split out: #581.

## Summary by CodeRabbit

* **Bug Fixes**
  * Sessions remain associated with the browser tab that opened them, preventing other connections from shutting them down.
  * Reconnecting tabs can resume sessions after network interruptions.
  * Disconnected sessions are released after approximately 45 seconds.
  * Device availability and “In use” indicators now consistently reflect active ownership.
  * Dashboard connections work reliably over non-secure HTTP contexts.

* **Documentation**
  * Updated guidance for session ownership, shutdown behavior, and connection liveness.

### DIFF
--- a/.changeset/relay-session-ownership-seam.md
+++ b/.changeset/relay-session-ownership-seam.md
@@ -7,8 +7,8 @@
 Answer a shutdown the relay cannot deliver, release every session a closing socket held, and let a viewer re-join a session it already holds
 
 Three defects that share a subject — who holds a session — and none of which needed the question that
-sounds like their root. That one is *who should be allowed to*, and it is answered in this same release — see the note titled "A
-session belongs to whoever opened it". It could not be answered in this slice because the dashboard's four
+sounds like their root. The root question — **who should be allowed to drive a session** — is answered in this same release; see
+the note titled "A session belongs to whoever opened it". It could not be answered in this slice because the dashboard's four
 senders per tab are four connections, and a socket-shaped owner refuses the tab's own teardown.
 
 What a user can observe:
--- a/.changeset/relay-session-ownership-seam.md
+++ b/.changeset/relay-session-ownership-seam.md
@@ -7,9 +7,9 @@
 Answer a shutdown the relay cannot deliver, release every session a closing socket held, and let a viewer re-join a session it already holds
 
 Three defects that share a subject — who holds a session — and none of which needed the question that
-sounds like their root. That one is *who should be allowed to*, and it is open: `device:shutdown` still
-has no ownership check, because three of the dashboard's four senders come from a socket that never
-joins the session and gating it there would break going back and the unmount teardown.
+sounds like their root. That one is *who should be allowed to*, and it is answered in this same release — see the note titled "A
+session belongs to whoever opened it". It could not be answered in this slice because the dashboard's four
+senders per tab are four connections, and a socket-shaped owner refuses the tab's own teardown.
 
 What a user can observe:
 
--- a/.changeset/session-owner-is-a-tab.md
+++ b/.changeset/session-owner-is-a-tab.md
@@ -19,8 +19,8 @@ What a user can observe:
 - **A Wi-Fi blip no longer costs you your session.** The relay treated a connection as present until TCP
   or a heartbeat noticed otherwise, up to a minute after a laptop went to sleep. Returning within that
   window meant being told the device was in use — by yourself. The tab is now recognised as the same tab, and the
-  session is simply resumed — unless you reload while the connection is down, which makes a new tab and
-  waits out the window.
+  session is simply resumed — unless you reload while the connection is down, which gives the page a new
+  identity even though it is the same tab, and waits out the window.
 - **A device whose tester's connection died frees up in at most 45 seconds** rather than up to a minute,
   and no longer appears free while it is still in use. (A tester who leaves a tab open is a different
   case — that is the idle timer's, not this.) Both questions — "can I take this?" and the "In use" badge — now read the same signal, and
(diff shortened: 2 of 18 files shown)


jo-duchan:
@coderabbitai review

jo-duchan:
Addressed in e4baf00. Thanks — the Major one was real, and worse than reported.

**`mayShutDown` recovered mintedness from a caller-supplied string.** Half of the owner key is the
`?client=` query parameter, so the `includes(':minted-')` test was reading a security decision out of
input. Both bypasses reproduced:

| holder sends | result |
|---|---|
| `?client=minted-1` | gate disabled entirely — back to the state #527 fixed |
| `?client=x:minted-y` | **a different user** can power off this device |

The second is the one that matters: the exemption was documented as never crossing a user boundary, and
the prefix test compared no users at all.

Fix: `ownerKeyFor` now returns `{ key, user, minted }`, the session carries `ownerMinted` and `ownerUser`
as fields, and the exemption is scoped to the same principal. Mintedness is granted by the relay and never
claimed. Four tests added, including two users authenticated with separate JWT cookies.

While mutating that line I found the PAT principal had no test at all — a local connection never reaches
`verifyPat` (it is consulted only when the caller is neither local nor cookie-authenticated), so the
obvious test would have compared `anon` to `anon` and passed. It now connects through a trusted loopback
proxy with `x-forwarded-for`, which is the state a real PAT client is in.

The three prose findings are fixed too: the incomplete clause in `relay-session-ownership-seam.md`, the
"makes a new tab" wording (same tab, new document), and both places that called the exemption
unconditional — no longer true now that it is user-scoped.

Full gate green: 2684 tests, `pnpm typecheck` 0 errors, `pnpm lint` 0 errors.

jo-duchan:
Fixed in a3e54ea — the sentinel was the part worth changing, and for the reason given: `handleConnection`
records every accepted socket, so nothing reaches that branch. The cost is not an exploit today but that the
invariant rested on that one call site staying total. A shared literal makes two unrecorded sockets *equal*,
and equal is what `ownsSession` tests for while equal principals are what the shutdown exemption tests for —
so a future path that accepted a socket without recording it would not fail closed, it would hand every such
socket ownership of every other's session.

The fallback now mints an identity for the socket and stores it: repeat calls agree, no two sockets ever do.
`minted` stays false — the exemption is for a client that identified itself with nothing, not for one the
relay failed to record. `ownerOf` and `userOf` delegate to `ownerRecord`, so there is one fallback rather
than three.

Tested by reaching past `private`, deliberately: there is no wire that produces this state, which is the
point — a test that could only drive it through a connection would be testing the seeding instead. Four
mutations, all caught: shared sentinel restored, unique key with shared user, not stored (fresh value per
call, which would stop a socket owning the session it just bound), and `minted: true`.

Issue #589, opened 2026-08-18
dashboard: Mac Resources and the comment panel carry a11y gaps the boot-fix pass did not close


jo-duchan:
## What

Two small UI fixes (#586 follow-up branch) put `MacResources.tsx` and `CommentPanel.tsx` through the a11y gate, which surfaced a backlog well beyond the diff. The blocking findings were fixed there — the chart's readings are keyboard-reachable through a `slider` cursor, the composer's controls are named, attachments announce their filename. These were reported alongside them and are not fixed:

- **`MacResources` has no unconditional `h1`.** The only one is the selected Mac's name, which renders conditionally, so the no-selection, loading and empty states have an `h2` (the chart titles, and the sidebar's "Macs") with no `h1` above them. The route needs a stable page heading.
- **The Mac list's state lives only in CSS.** Selection is a `bg-accent` class and online/offline is a coloured dot, so a screen-reader user can reach the region but cannot tell which Mac is selected or which are up. Wants `aria-current` and a visually-hidden status word.
- **The comment file-validation message is not announced.** Pick a file that is too large and the error appears as a plain `<p>`: no `role="alert"`, and the attach button does not point at it, so the picker's failure is silent to AT.

None of them blocks a commit today (they were warnings), and none is in the path either fix took. Filed rather than folded in because the branch was two UI corrections before a release, and this is a pass over a whole page.

Worth knowing for whoever takes it: the gate reviews the **files** a commit stages, not the hunks, so any change to these two files will surface all of this again.

Giannhskrg:
Hey, i'd like to be assigned for this. Let me know if anyone is already working on it.

jo-duchan:
Thanks for the interest! We don't assign issues ahead of time — see [CONTRIBUTING.md#claiming-an-issue](https://github.com/jo-duchan/tapflow/blob/main/CONTRIBUTING.md#claiming-an-issue). Open a draft PR or leave your findings on this issue, and it's yours. Nobody's working on it right now, so feel free to jump in.

A few pointers to get you started:

- The relevant files are `packages/dashboard/src/pages/MacResources.tsx` and `packages/dashboard/components/CommentPanel.tsx`.
- You don't need macOS for this one. Node.js 22+ and pnpm are enough. Run `pnpm install`, then `pnpm --filter @tapflowio/dashboard dev`. Without a relay running the Mac list will be empty, which is fine here — that's the no-selection state the first item is about.
[...]

PR #590, opened 2026-08-18
Keep the resource series inside its plot, and match the comment form's radius

jo-duchan:
## Summary

Two reported UI defects, plus the changeset backfill the release audit was blocking on.

**The resource chart painted through its own y-axis.** The window is anchored to a rounded-up tick so the
times stay round, which puts its left edge up to one step later than the oldest sample the relay returns —
ten minutes of points on the 1h range, mapped to a negative x because `scaleTime` does not clamp. Worst on
RAM, which sits where the "50%" and "25%" labels are. The series is clipped to the plot rather than
filtered, so an off-window sample still shapes the curve at the edge.

**The comment composer's radius** was `rounded-xl` (12px), which is not on this project's scale, against
the `rounded-md` (6px) list directly above it.

Touching those two files put both through the a11y gate, four rounds of it, and what it found was not
decoration: the chart's readings were mouse-only and the composer was unnamed. The chart's cursor is now a
`slider` over the sample index — the role whose native key model is the arrow keys — announcing through
`aria-valuetext`. Two of the gate's findings were bugs in the code I had just written rather than a11y
points: an unclamped cursor (7d → 1h left `aria-valuenow` past `aria-valuemax`) and a pointer path that
bypassed it, so the slider announced something other than what it showed.

One defect no gate caught and the reporter did: `outline-none` is a *transparent* 2px outline, so setting
`outlineColor` inline drew a black box around every plot at rest.

Also backfills a changeset for #537, which changed eight manifests with no release note — marked internal,
which is what `pnpm changeset:audit` was reporting before this.

## Checklist

- [x] Tests written and passing — 10 new, on a component that had none; 5 mutations run, 5 caught
- [x] No `any`
- [x] Interface changes land in `agent-core` first — n/a

## Related `.work/` docs

Review record is local (`.work/` is gitignored). The adversarial review was **skipped** deliberately — two
visual defects inside one package, judged against a screenshot — and the a11y gate served as the
independent context instead; its four rounds are recorded there. Worth one caveat: on the final commit
a11y-lens **timed out and skipped** rather than passing. The two errors it raised the round before are
fixed and only prose changed after, but the last state of this branch has not been through it.

Split out rather than fixed here: #589.

## Summary by CodeRabbit

- **New Features**
  - Mac resource charts now support keyboard navigation, screen-reader guidance, Escape-to-dismiss behavior, and accessible value readouts.
  - Mouse, touch, and keyboard interactions now share consistent chart cursor behavior.
- **Bug Fixes**
  - Chart data is clipped correctly within plot boundaries, and cursor positions remain valid when ranges change.
  - Comment attachment filenames are decoded for clearer accessible labels.
- **Accessibility**
  - Added accessible names and semantic headings for controls, forms, charts, and decorative icons.
- **Tests**
  - Expanded coverage for chart interaction, accessibility, clipping, and keyboard behavior.

### DIFF
--- a/.changeset/typecheck-and-lint-tests-backfill.md
+++ b/.changeset/typecheck-and-lint-tests-backfill.md
@@ -1,0 +1,24 @@
+---
+'@tapflowio/agent-core': patch
+'@tapflowio/android-agent': patch
+'@tapflowio/audiotap-helper': patch
+'@tapflowio/cli': patch
+'@tapflowio/flow-runner': patch
+'@tapflowio/ios-agent': patch
+'@tapflowio/mcp-server': patch
+'@tapflowio/relay': patch
+---
+
+Type-check and lint the test trees
+
+Backfills: #537
+
+<!-- changelog: internal — a `typecheck` script and a test-tree tsconfig per package; no runtime or
+     interface change a self-hoster can observe. What it found was inside the tests themselves: a double
+     declaring `implements DeviceAgent` while missing two members, five duplicate object keys, a call
+     passing one argument to a two-argument method, and a `test-utils` constraint no named message could
+     satisfy. -->
+
+Every package's build tsconfig excluded `src/__tests__` and eslint ignored it, so a test double could
+drift from the interface it doubled with both gates green. The manifests gained a `typecheck` script and
+the test trees a tsconfig of their own, which is the only reason this touches published files at all.
--- a/.changeset/typecheck-and-lint-tests-backfill.md
+++ b/.changeset/typecheck-and-lint-tests-backfill.md
@@ -2,7 +2,7 @@
 '@tapflowio/agent-core': patch
 '@tapflowio/android-agent': patch
 '@tapflowio/audiotap-helper': patch
-'@tapflowio/cli': patch
+'tapflow': patch
 '@tapflowio/flow-runner': patch
 '@tapflowio/ios-agent': patch
 '@tapflowio/mcp-server': patch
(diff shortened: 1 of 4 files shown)


jo-duchan:
Both findings acted on, and the CI failure underneath them turned out to be a third thing.

**The changeset named a package that does not exist — fixed, and it would have surfaced at release.** The manifest is `tapflow`, not `@tapflowio/cli`, and `changeset version` resolves the manifest. That is worth saying plainly: this changeset exists *because* `pnpm changeset:audit` flags merges that would reach release day with no note, so shipping it with a name that breaks `changeset version` would have moved the failure to exactly the moment the audit is meant to protect.

The other half of that finding I did not take: `@tapflowio/protocol` does not belong in it. The only protocol file #537 touched is `packages/protocol/AGENTS.md` — bumping and publishing a package to carry a doc edit is a release note for nothing.

**The cursor jumping to the end on refocus — fixed.** It starts unplaced (`null`) rather than at zero, so the first focus opens at the newest sample and a placed one is still where the reader left it after Escape or a blur. Regression test added, and the mutation that restores the old behaviour fails it.

**The failing `changeset` job was mine too, and separate from both.** The internal marker has to be on a line of its own — `/^<!--\s*changelog:\s*internal\b.*-->$/m` does not span lines — and I had wrapped it across five, so the gate never saw it and asked for a root changelog entry instead.

While this was open the a11y gate ran eight rounds over the two files and now reports no findings. Four of its later points were defects in the code I had just written rather than accessibility notes: an unclamped cursor, a pointer path that bypassed it, a reading built from the axis-label format (every sample in a day announced identically on the 7d range), and three separate renderings of the same number that disagreed on precision. One formatter now serves the tooltip, `aria-valuetext` and the chart's summary name.

Gate green: 2707 tests, typecheck and lint clean, `pnpm changeset:check` and `changeset:audit` both pass.

Issue #293, opened 2026-06-16
feat(relay): parse cert CN/SAN to show the domain in the import-cert start banner


jo-duchan:
## Problem

For `tls.mode: byo-api-token`, the relay start banner advertises the cert's domain (`config.tls.domain`). For `tls.mode: import-cert` there is no `domain` field, so the banner falls back to `localhost`:

```ts
// packages/cli/src/commands/relay-start.ts
// A domain-bound cert won't validate against localhost, so advertise the cert's domain instead.
const displayHost = config.tls?.mode === 'byo-api-token' ? config.tls.domain : 'localhost'
```

But an imported certificate is bound to its own domain. Telling the user to open `https://localhost:<port>` produces a name-mismatch warning — the exact problem the byo-api-token branch already avoids. The same fallback exists in `packages/relay/src/server.ts`.

## Proposal

When `mode === 'import-cert'`, read the certificate file and extract the host to advertise:

- Parse the leaf certificate's SAN (`subjectAltName`) DNS names, falling back to the subject CN.
- Prefer a SAN DNS entry over CN (CN is deprecated for hostname matching).
- For a wildcard SAN (`*.example.com`), there is no single concrete host — show the base domain or keep `localhost` with a note. Decide during implementation.
- Node's `tls`/`crypto` can parse a PEM via `new X509Certificate(pem)` (`.subjectAltName`, `.subject`); no new dependency.

## Definition of done

- [ ] import-cert start banner shows a domain parsed from the cert (SAN preferred, CN fallback) instead of `localhost`.
- [ ] Applied in both `relay-start.ts` and `server.ts` (shared helper to avoid duplication).
- [ ] Graceful fallback to `localhost` if the cert is unreadable/unparseable or only has a wildcard/IP SAN.
- [ ] Unit test for the parse helper: SAN DNS, CN-only, wildcard, malformed input.

## Context

Deferred review item from the LAN HTTPS work (#232). Low priority — cosmetic banner accuracy, no functional impact on TLS termination.

Issue #260, opened 2026-06-11
perf(relay): caching + compression headers for static assets (serveStatic)


jo-duchan:
## Problem
`serveStatic` (`packages/relay/src/RelayServer.ts:787-811`) sends no `Cache-Control` / `ETag` / `Last-Modified`, so hash-named Vite assets are re-downloaded on every load. No gzip/brotli compression either.

## Proposal
- Hashed assets (`/assets/*`): `Cache-Control: public, max-age=31536000, immutable`
- `index.html`: `no-cache`
- Prefer pre-compressed `.br` / `.gz` files when present
- Combine with the async-stat cleanup (see P2 housekeeping issue)

Noticeably improves dashboard initial load. (Review P1 — performance.)

PR #604, opened 2026-08-20
fix(scripts): ignore changelog markers in code fences

### LINKED ISSUE #560: changeset gate: the changelog: internal marker is honoured inside fenced code blocks

pttydou:
## Summary

- Ignore `changelog: internal` markers inside fenced code blocks.
- Track fence characters and lengths so nested examples do not close the outer fence early.
- Cover backtick, tilde, nested, mixed, and post-fence cases.
- The shared parser also prevents quoted markers in nested fences from bypassing the `no-changeset` and `Backfills:` gates.

Closes #560

## Validation

- Focused Vitest suites: 61 tests passed
- `pnpm lint`
- `pnpm typecheck`

## Checklist

- [x] No `any`

## Related `.work/` docs

- `.work/2026-08-20-changeset-fenced-marker-plan.md`
- `.work/reviews/feature__changeset-fenced-marker.md`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved changelog marker handling so markers inside Markdown code blocks are ignored.
  * Added support for backtick and tilde fences with varying lengths, indentation, and nested fence-like content.
  * Correctly handles closing fences with additional text and resumes scanning afterward.
  * Markers outside code blocks continue to release the required changelog entry.
  * Clarified that explanatory text following the internal marker is optional.

### DIFF
--- a/scripts/__tests__/auditBackfills.test.mjs
+++ b/scripts/__tests__/auditBackfills.test.mjs
@@ -62,4 +62,15 @@ describe('parseBackfills', () => {
   it('allows prose after the references on the same line', () => {
     expect(parseBackfills(body('Backfills: #413 — the shortcuts never worked before it'))).toEqual([413])
   })
+
+  it.each([
+    ['a shorter fence nested inside a longer one', '````md', '```md', 'Backfills: #413', '```', '````', 'Backfills: #414'],
+    ['a closing fence with an info string', '```md', '``` not-a-close', 'Backfills: #413', '```', 'Backfills: #414'],
+  ])('ignores a reference inside %s and resumes after it', (_name, ...lines) => {
+    expect(parseBackfills(body(...lines))).toEqual([414])
+  })
+
+  it('reads an unindented reference between four-space-indented fence-like lines', () => {
+    expect(parseBackfills(body('    ```md', 'Backfills: #413', '    ```'))).toEqual([413])
+  })
 })
--- a/scripts/__tests__/changesetGateAccuracy.test.mjs
+++ b/scripts/__tests__/changesetGateAccuracy.test.mjs
@@ -81,6 +81,11 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
     'internal.md': `${cs('"@tapflowio/protocol": patch')}\n<!-- changelog: internal — protocol typing, nothing observable -->\n`,
     'no-reason.md': `${cs('"@tapflowio/protocol": patch')}\n<!-- changelog: internal -->\n`,
     'mentions-it.md': `${cs('"@tapflowio/relay": patch')}\n\nSee the changelog: internal notes are not a marker.\n`,
+    'backtick-fence.md': cs('"@tapflowio/relay": patch') + '\n```md\n<!-- changelog: internal -->\n```\n',
+    'tilde-fence.md': `${cs('"@tapflowio/relay": patch')}\n~~~md\n<!-- changelog: internal -->\n~~~\n`,
+    'long-backtick-fence.md': cs('"@tapflowio/relay": patch') + '\n````md\n```md\n<!-- changelog: internal -->\n```\n````\n',
+    'mixed-fence.md': cs('"@tapflowio/relay": patch') + '\n```md\n~~~\n<!-- changelog: internal -->\n~~~\n```\n',
+    'after-fence.md': cs('"@tapflowio/protocol": patch') + '\n```md\nexample\n```\n<!-- changelog: internal -->\n',
   })[f]
 
   it('owes one for a plain changeset', () => {
@@ -98,6 +103,12 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
   // an opt-out -- the shape this repo has been bitten by twice, where a comment was read as a value.
   it('is not released by prose that merely says the words', () => {
     expect(changelogEntryOwed(['mentions-it.md'], read)).toEqual(['mentions-it.md'])
+  })
+  it.each(['backtick-fence.md', 'tilde-fence.md', 'long-backtick-fence.md', 'mixed-fence.md'])('is not released by a marker inside %s', (file) => {
+    expect(changelogEntryOwed([file], read)).toEqual([file])
+  })
+  it('is released by a marker after a closed fence', () => {
+    expect(changelogEntryOwed(['after-fence.md'], read)).toEqual([])
   })
 })
 
--- a/scripts/__tests__/changesetGateAccuracy.test.mjs
+++ b/scripts/__tests__/changesetGateAccuracy.test.mjs
@@ -86,6 +86,7 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
     'long-backtick-fence.md': cs('"@tapflowio/relay": patch') + '\n````md\n```md\n<!-- changelog: internal -->\n```\n````\n',
     'mixed-fence.md': cs('"@tapflowio/relay": patch') + '\n```md\n~~~\n<!-- changelog: internal -->\n~~~\n```\n',
     'after-fence.md': cs('"@tapflowio/protocol": patch') + '\n```md\nexample\n```\n<!-- changelog: internal -->\n',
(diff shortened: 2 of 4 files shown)


jo-duchan:
Hi @pttydou, and welcome — thanks for picking this one up.

The fence tracking is right, and it is right for the reason that matters: you implemented the CommonMark rules rather than a heuristic that happens to pass the examples in #560. Backtick fences reject an info string containing a backtick while tilde fences accept one, a closing fence has to match the character and be at least as long with no info string, and the three-space indent allowance is there. The nested case in the issue was exactly the length comparison, and that is what you fixed.

**Two things before this lands**, and both come from the same fact — one I don't think the PR realises about itself.

## `proseLines` feeds three gates, and this changes all of them

I ran your version against `main` on the same inputs:

```
── extractReason  (the no-changeset gate)
 != inside a nested fence         main: "x"    | PR: ""
 != closing fence has info string main: "x"    | PR: ""
 != fence indented four spaces    main: ""     | PR: "x"

── parseBackfills  (the Backfills: gate)
 != inside a nested fence         main: [413]  | PR: []
```

The first row is the finding. **The leak #560 describes was live in the `no-changeset` gate too** — a `<!-- no-changeset: ... -->` quoted inside a nested fence was read as a real marker on `main`, and that marker switches off a required check. A PR body that quoted the snippet from CONTRIBUTING.md inside a nested block could turn the gate off by accident. Your change closes that, and closes the same thing on `Backfills:`.

That is the right way to fix it — the function is shared precisely so this cannot drift, and its docstring says so:

> `extractReason` had this guard and `parseBackfills` was written without it, which is exactly the kind of drift a second copy invites.

But the tests only reach `changelogEntryOwed`. Landing one implementation while testing one caller puts the other two gates' fence behaviour on someone else's assertions, which recreates the asymmetry that docstring is warning about, one layer up. **Could you add the nested-fence case to `extractReason` and `parseBackfills`?** An `it.each` alongside the ones you wrote should do it.

## The four-space case is the one direction that loosens a gate

Row three above. `main` ignored a marker sitting between two four-space-indented fence lines; your version honours it.

I think you're right — four spaces makes those lines an indented code block rather than a fence, so an un-indented marker between them really is prose, and someone quoting the marker on purpose would have indented the marker too (that case still comes out ignored, I checked). So I'm not asking you to change the behaviour.

I would like it pinned by a test, though. It is the only input where this makes a gate easier to switch off than it was, and that is the kind of change that should fail loudly if a later edit moves it rather than being rediscovered by whoever is holding the diff.

## Checked and found sound

- Removing the "the reason is required" paragraph is correct, and I went looking for whether it was a behaviour change in disguise. It isn't: `accepts the marker with no reason after it` already exists on `main`, and its comment settles it — *"Requiring prose here would be a second, unstated rule."* The prose was stale and you aligned it. That belongs in this PR.
- An unclosed fence now runs to end of file, so the marker is ignored and the entry is owed. Fail-closed, which is the direction I want.
- Moving `changelogEntryOwed` off a whole-file regex also gains the frontmatter skip. Harmless, and slightly better.
- Marker anchoring is unchanged — you kept `raw` rather than the trimmed line, so an indented marker still does not match, same as before.
- Full `scripts` suite green on your branch: 28 files, 357 tests.

## One line for the PR body

The title and summary are all about `changelog: internal`. Whoever reviews this next should be able to see that two other gates stopped leaking out of nested fences, because that is the part with a security-ish flavour. A sentence is enough.

Nice work on this one. The issue asked for a single gate and you fixed three.

pttydou:
Added regression coverage for nested fences in `extractReason` and `parseBackfills`, including post-close sentinels, and pinned the four-space-indented fence-like case across all three gates. I also updated the PR summary to call out the `no-changeset` and `Backfills:` behavior. Focused tests: 60 passed.

jo-duchan:
Thanks — this covers it.

All three landed, and I re-ran the differential against `main` to check the new cases aren't vacuous: every assertion you added is one where `main` and this branch give different answers. The post-fence sentinels are what make "resumes after it" an actual claim rather than an assumption, and I'm glad you put them in.

**One thing left**, and it is the rule I singled out last time.

## The backtick info-string clause is the only one with no test

I mutated each rule in `proseLines` in turn and ran your 60:

```
baseline                                    60 passed
closing fence: length comparison removed     2 failed
closing fence: info string allowed           2 failed
closing fence: character need not match      1 failed
opening fence: any indentation accepted      3 failed
opening fence: backtick in info allowed     60 passed   ← nothing catches it
```

Deleting ``!marker[2].includes('`')`` outright leaves the whole suite green.

It is also the one clause where being wrong makes a gate **easier** to switch off, which is the direction #560 was about:

````
How to quote it:

```js`
```
<!-- no-changeset: reason -->
```
````

An info string on a backtick fence may not contain a backtick, so per CommonMark that first line is a paragraph and the block opens on the line below it — which leaves the marker quoted. Your version returns `''`. Without the clause it returns `'reason'`, so a PR body explaining how to quote the escape hatch would switch the gate off by explaining it. That is #560 again, one input over.

One `it` beside the `it.each` in `changesetReason.test.mjs` is enough. I don't think the other two gates need a copy — they share `proseLines`, and the remaining four rules are already pinned three ways over.

## Re-checked

- 60/60 on `cde98eb7`, and lint, typecheck and both Node lines green in CI.
- The `parseBackfills` four-space case reads `#413` here against `[]` on `main`, so that pin sits on the behaviour that changed rather than beside it.
- `changelogEntryOwed`'s four-space case answers the same on both — that gate had no fence guard at all before this PR, so there was nothing there to loosen. Worth keeping as a pin anyway.
- The PR body now names the other two gates. That was the part a later reader had no way to see from the title.

**Comment by pttydou** [comment]
Added the backtick-info-string regression in `changesetReason.test.mjs`. I mutation-checked it as well: removing that guard makes the new test fail with `reason` instead of an empty result. Focused tests: 61 passed; lint and typecheck passed.

**Comment by jo-duchan** [comment]
This is good to go — thanks for sticking with it through the back-and-forth.

One heads-up first: I pushed a commit to your branch (`eae0f873`) instead of sending you through another review round. It covers the tilde case symmetric to the one you just handled. A tilde fence's info string may contain a backtick, while a backtick fence's may not, so `marker[1][0] === '~' ||` is doing real work. Without it, all 61 tests still pass even though the check stops working for `` ~~~js` ``. The added test is the same shape as yours:

```js
const body = '~~~js`\n<!-- no-changeset: reason -->\n~~~\n'
expect(extractReason(body)).toBe('')
```

It didn't feel right to send you back for the other half of a case you'd effectively already covered. If you'd rather add it yourself, just let me know and I'll drop the commit — either way is fine.

For what it's worth, this was a genuinely strong first contribution. You implemented the CommonMark rules instead of just matching the examples in the issue, which is why the fix ended up closing leaks in two other gates nobody had asked about. And when I came back with a mutation table, you ran the mutation yourself before replying instead of taking my word for it. That's a great habit, and one I don't see often enough.

If you're up for another issue, the [good first issue](https://github.com/jo-duchan/tapflow/labels/good%20first%20issue) label is where I keep them. #293 and #260 are both self-contained if you're looking for something next. Happy to point you toward either one, or answer any questions before you get started.

And if you find tapflow useful, a ⭐ would mean a lot. It's still how most people discover the project. No pressure, of course.

Thanks again for the contribution. 🙏

PR #613, opened 2026-08-20
fix(relay): advertise imported certificate host

### LINKED ISSUE #293: feat(relay): parse cert CN/SAN to show the domain in the import-cert start banner

pttydou:
## Summary

- Read the first concrete DNS SAN from imported certificates, with CN fallback only when the SAN extension is absent.
- Use the resolved host in tapflow start, tapflow relay start, and the standalone relay server output.
- Keep localhost for wildcard, IP, malformed, ambiguous, or padded identities, including legacy IPv4 forms normalized by browsers.

Closes #293

## Validation

- Relay parser and server tests: 14 passed.
- CLI start command tests: 35 passed.
- Relay and CLI typechecks passed.
- Relay and CLI lint passed with no new errors.
- Full pre-commit typecheck and lint passed.
- Changeset check passed.

## Checklist

- [x] No `any`

## Related `.work/` docs

- `.work/2026-08-20-import-cert-display-host-plan.md`
- `.work/reviews/feature__import-cert-display-host.md`

## Summary by CodeRabbit

- **New Features**
  - Relay startup messages now display the complete HTTP or HTTPS URL.
  - Imported TLS certificates use the first valid, non-localhost DNS SAN, with the subject CN used only when no SAN extension exists.
  - Certificates with unusable DNS SANs fall back to `localhost` and show a warning; IP-only or malformed certificates fall back without a warning.
  - Managed TLS configurations continue using their configured domain.

- **Documentation**
  - Updated configuration guidance and changelog entries to explain hostname selection, warnings, and fallback behavior.


jo-duchan:
Thanks for this — the parser is solid, and a couple of the fixtures cover ground I did not expect anyone to reach on a first pass.

I ran the seven fixtures through Node 24 directly and then tried to break the `', '` split, because the comment above it makes a strong claim and I wanted to know whether it holds. It does. I built two certificates designed to smuggle a second entry past the split:

```
DirName SAN   → "DirName:\"CN=x\\\\\\u002c DNS:evil.example.com\", DNS:real.example.com"
otherName SAN → "othername:\"UPN:x\\u002c DNS:evil.example.com\", DNS:real.example.com"
```

Both resolved to `real.example.com`. I expected `DirName` to get through, since RFC 2253 escapes commas with a backslash rather than JSON-quoting, but Node quotes that case too. The comment is accurate.

Two things I want to call out as good, so they don't get lost below. Refusing to fall back to CN when a SAN extension is present but unusable is right (RFC 6125 §6.4.4), and the test for it is well built: every fixture carries a concrete `CN=cn.example.com`, so a wrong fallback shows up as `cn.example.com` instead of `localhost` rather than passing silently. That is an assertion of absence that can actually fail. The `ip-like` fixture is the other one — knowing that WHATWG URL parsing folds `2130706433` and `0x7f000001` into `127.0.0.1` is not obvious, and most implementations of this miss it.

## Three things before I merge

**1. A cert whose first DNS SAN is `localhost` gets no benefit from any of this.**

This is the shape `mkcert` produces, and I think it is the most common one for a self-hosted LAN setup:

```
SAN: DNS:localhost, DNS:tapflow.lan, IP Address:192.168.1.5
→ localhost
```

The comment the change replaces states the purpose as advertising the cert's domain *to teammates*, and this case misses exactly that. What makes it worth fixing rather than accepting is that the outcome depends on the order the operator happened to list names in when the cert was issued — which is not something anyone has a reason to think about.

Picking the first concrete DNS SAN that isn't `localhost`, and falling back to `localhost` when there is none, keeps every current test result identical. It is also the tradeoff `byo-api-token` already takes: it advertises `tls.domain` unconditionally, accepting that the advertised name has to resolve.

**2. A wildcard-only certificate says nothing about why it fell back.**

#293 asked for "show the base domain or keep `localhost` with a note", and the note is missing — the operator just sees `localhost` with no indication that a certificate was read at all.

This matters more than it looks, because `docs/reference/configuration.md` leads with that exact case:

> To use an internal PKI or **a wildcard certificate you already hold**, point to the files.

So the use case the documentation introduces first is the one that changes nothing here and explains nothing about why. A single `warn` when the certificate has DNS SANs but none usable would cover it.

**3. `docs/reference/configuration.md` doesn't say what import-cert advertises.**

The `tls.domain` row tells the reader "Teammates open `https://[domain]:[port]`". There's no equivalent for import-cert. Nothing in there is false, but the gap is one this change creates — a line or two describing the rule (first concrete DNS SAN, else `localhost`) would close it.

## Nits — take or leave

- `concreteDnsHost` opens with `const host = value` and never reassigns; the alias can go.
- `parseCertDisplayHost` is exported from the package index, but only `resolveRelayDisplayHost` has a consumer. A public export is hard to withdraw later — I'd keep it internal until something outside the package needs it.
- `server.test.ts` has no `afterEach(() => vi.restoreAllMocks())`, so the `process.on` spy is never restored. Harmless with one test in the file; less so once there are two.
- A trailing-dot FQDN (`example.com.`) and any host with an underscore are rejected and fall through to `localhost`. Both are rare enough that I'm fine leaving them.

## Split out

The agent-connect line under the banner still reads `--relay wss://<this-ip>:4000`, so copying it from a TLS banner produces the name mismatch this PR removes from the line above. I did not put it here because interpolating `displayHost` unconditionally would break the HTTP path, where it is `localhost` and the placeholder is doing real work. Filed as #627.

## One CI note

The `changeset` check is reporting `status: in_progress` with `conclusion: success`, so it may need a re-run before the merge button frees up. Nothing on your side.

pttydou:
Addressed the three merge blockers:

- SAN selection now skips `localhost` and continues to the first concrete DNS name; the `localhost, tapflow.lan, IP` fixture resolves to `tapflow.lan`.
- Import-cert emits one warning when DNS SANs are present but none is usable, while keeping the `localhost` fallback.
- The English and Korean configuration docs now describe SAN selection, CN fallback, and warning behavior.

I also removed the unused alias and public parser export, and restored the server test spies.

Validation: 16 relay tests, 35 CLI tests, both targeted mutations, the full pre-commit typecheck/lint, the docs build, and GitHub Actions all pass. Vercel is still the external authorization check.

jo-duchan:
Thanks for turning this around so quickly — and for going wider than I asked on the first one.

I checked the three against the parser rather than against the diff, running every fixture through Node 24's `X509Certificate` and recording both what it advertises and whether it warns:

| SAN | advertises | warns |
|---|---|---|
| `DNS:localhost, DNS:tapflow.lan, IP:192.168.1.5` | `tapflow.lan` | no |
| `DNS:LOCALHOST, DNS:tapflow.lan` | `tapflow.lan` | no |
| `email:a@b.com, DNS:localhost, DNS:relay.lan` | `relay.lan` | no |
| `DNS:*.example.com` | `localhost` | **yes** |
| `IP Address:127.0.0.1` | `localhost` | no |
| no SAN extension / malformed PEM | CN or `localhost` | no |

All three are closed. The first is closed wider than the report was: an uppercase `LOCALHOST` and a non-DNS entry sitting ahead of the list both survive it, and neither was something I asked for. The warning is wired at all three entry points — `logger.warn` in the relay, `print`'s `warn` in both CLI commands — and, more importantly, it stays quiet where it should: an IP-only SAN and a malformed certificate reach `localhost` without it, because neither has a DNS SAN to complain about. The docs hold up clause by clause against that table, including that distinction, which is the easiest half to get wrong.

Two things left. Neither is a correctness problem in the parser.

## 1. `CHANGELOG.md` didn't move with the last two commits

The changeset was updated for the localhost skip and the warning, but the `[Unreleased]` entry in `CHANGELOG.md` still describes the behaviour from before the review — so the two release notes now say different things about what this PR does. A sentence in `### Fixed` brings them back together. The entries around it link their issue, so a `([#293](https://github.com/jo-duchan/tapflow/issues/293))` would match the neighbours.

## 2. The CN guard has no test — I confirmed by mutating it

Replacing `concreteDnsHost(commonName) ?? FALLBACK_HOST` with `commonName ?? FALLBACK_HOST` leaves **all 16 assertions in `parseCert.test.ts` green**, while the parser begins advertising these as URL hosts:

```
CN=*.example.com  (no SAN) → *.example.com
CN=192.168.1.5    (no SAN) → 192.168.1.5
CN=0x7f000001     (no SAN) → 0x7f000001
```

The cause is that every no-SAN fixture carries a valid CN (`tap.example.com`), so nothing in the suite distinguishes "the CN was validated" from "the CN was passed straight through". That is the shape [`contributing/test-and-guard-coverage.md`](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) is about — a guard whose removal produces no visible difference in a fully green run. CodeRabbit reached the same line independently, which is usually a sign it is worth taking. One fixture with no SAN and a wildcard CN, plus an entry in the `it.each` that already exists, closes it.

One aside while you are generating that fixture: `tls-cert-san-ip-like.pem` and `tls-cert-san-whitespace.pem` were issued with a one-day validity and expired yesterday. Nothing fails and nothing will, since `parseCertDisplayHost` never looks at dates — but the other five are ten-year certificates, and matching them saves the next person from finding a fixture that looks broken for a reason unrelated to what it tests.

## On the fixtures, for the record

I looked at the seven new `.pem` files from a security angle, since committed certificate material is worth a second look. They are clean: certificate blocks only, no key material, each one self-signed under `example.com` — reserved by RFC 2606, so no CA anywhere trusts them — and each carries its own keypair whose private half was never committed. They cannot serve traffic or be presented as anything. Good instinct generating fresh keys per fixture rather than reusing the existing test key.

Once those two land I will approve — the merge is waiting on a review, not on CI. The Vercel check is deployment authorization for an outside contributor and has nothing to do with your code.

pttydou:
Addressed both remaining items:

- Updated `[Unreleased]` to match SAN selection, CN fallback, and warning behavior, with #293 linked.
- Added a no-SAN wildcard-CN fixture. Replacing the guarded CN fallback with direct `commonName` now fails: expected `localhost`, received `*.example.com`.

I also reissued the two one-day fixtures with the same CN/SAN values and ten-year validity; all touched fixtures are certificate-only.

Validation: parser tests (16/16), CN-guard mutation, repository typecheck/lint, docs build, and changeset check. GitHub Actions are running.

jo-duchan:
Thanks — clean turnaround. I checked both against the parser rather than the diff, running every fixture through Node 24's `X509Certificate` and re-running the mutation:

- `CHANGELOG.md` now matches the changeset and both docs, with #293 linked.
- `tls-cert-cn-wildcard.pem` kills the mutation: with `commonName ?? FALLBACK_HOST` it advertises `*.example.com` while all seven other fixtures are unchanged, so the new `it.each` entry fails exactly where it should.
- The two reissued fixtures keep their subject and SAN identical and moved only the validity to ten years. Certificate-only, fresh keypair each.

One small thing before I approve, and it is the same shape as the CN guard.

The warning is tested where it fires (wildcard SAN → once), but not where it has to stay quiet. `resolveRelayDisplayHost` is mocked in the relay and CLI suites, so `parseCert.test.ts` is the only place the real warn path runs — and only the wildcard case passes a spy. Both of these leave all 16 green:

- `warnAboutFallback: hasDnsSan` → `true`
- the `catch` branch's `warnAboutFallback: false` → `true`

So "IP-only and malformed certificates fall back without that warning", which the CHANGELOG and both docs now promise, is not held by anything yet. A spy on `tls-cert-san-ip.pem` and `'not a certificate'` in the existing `resolveRelayDisplayHost` block covers it. Since those are absence assertions, please run the two mutations above and confirm they go red before committing ([`contributing/test-and-guard-coverage.md`](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) §2) — the wildcard test already proves the spy is wired, so it should be a couple of lines.

Separately: a wildcard or IP CN also falls back silently, with no SAN to warn about. That is a design question rather than a defect here, so it is filed as #631 rather than held against this PR.

pttydou:
Added real warn-path coverage for both silent fallbacks: an IP-only SAN and malformed certificate material. Each still resolves to localhost, and the tests assert that the warning callback is not called. I ran both requested mutations; forcing either branch to warn makes its corresponding case fail. Parser tests pass (18/18), and the full pre-commit typecheck, lint, and a11y-lens checks pass.

jo-duchan:
That closes it — thank you.

I ran the two mutations crosswise rather than re-reading the diff: forcing the SAN branch to warn fails only the IP-only case, forcing the `catch` branch fails only the malformed one, and the wildcard test survives both. So the two new cases each carry their own weight, and the positive control did not quietly absorb them. Pairing each `not.toHaveBeenCalled()` with the host assertion is what makes them read as more than "nothing happened".

Worth saying too: this landed in the test file alone, eleven lines, parser untouched. Reaching for the implementation is the easy way out of a coverage gap, and you did not take it.

Four rounds, and each one came back a little wider than the report asked — the `localhost` skip catching uppercase and non-DNS entries, a fresh keypair per fixture, the expiry reissue nobody would have gone looking for. Thanks for your patience with a demanding review; it is a better change than the one I first read.

Merging from here. #631 tracks the silent CN-path fallback, and that one is not yours to carry.

PR #630, opened 2026-08-21
android-agent: scrcpy server process exits with no recorded cause

### LINKED ISSUE #481: android-agent: scrcpy server process exits with no recorded cause
Follow-up to #474, which added the missing `'error'` handler to the scrcpy server process. This is the remaining gap on the same object, scoped down from what I described in that review.

## Correcting the framing from #474

In the #474 review I said the process could die while the session still looked alive to the dashboard. That is **not** accurate, and I want it on the record before anyone builds on it.

`ScrcpySession` opens `videoSocket` against the forwarded port, and `ScrcpyVideo` closes its `ReadableStream` on the socket's `'end'`/`'close'`:

```ts
socket.on('end', () => this.finish())
socket.on('close', () => this.finish())
```
(`packages/android-agent/src/scrcpy/ScrcpyVideo.ts:55`)

When the server process dies, the `adb shell` child goes with it, the socket tears down, the reader sees `done`, `pumpVideo()` resolves, and `AndroidAgent` restarts the stream:

```ts
void this.pumpVideo(...).then(() => {
  if (state.scrcpySession === session && !state.restarting) {
    state.restarting = true
    void this.restartVideoStream(state)
  }
})
```
(`packages/android-agent/src/AndroidAgent.ts:589`)

So recovery already happens. The session does not silently rot.

## The actual gap

`serverProc` has no `'exit'` handler, so **when it dies, nothing records why.**

`EmulatorVideo` does the opposite on the same class of process:

```ts
this.encoder!.on('exit', (code) => onDead(`encoder exited (${code})`))
this.encoder!.on('error', (e) => onDead(`encoder process: ${e.message}`))
```
(`packages/android-agent/src/emulator/EmulatorVideo.ts:111`)

On the scrcpy path the exit code is dropped on the floor. What an operator sees in the log is a video restart with no stated cause — and if the underlying condition persists (device unplugged, `app_process` failing, the server OOM-killed on the device), a restart loop with no cause either. The `'error'` handler #474 added covers spawn-time failures; it says nothing about a server that started fine and exited later, which is the more likely production shape.

## Suggested change

Attach an `'exit'` handler next to the `'error'` one in `ScrcpySession.start()` and log the exit code and signal. Deliberately *not* proposing that it drive teardown — `pumpVideo` already owns that, and a second teardown trigger racing the socket path is a good way to turn a working restart into a double-restart. This is about diagnosis only.

Worth checking while in there whether the log should distinguish an expected exit (our own `stop()` / `kill()`) from an unexpected one, so routine session teardown doesn't print an error-shaped line every time.

## Severity

Low. Nothing is broken for users; this costs debugging time when a real-device session starts restarting and the log cannot say why. Reproducing it needs a physical Android device, so it's `requires: macOS` territory rather than something verifiable from a unit test alone — though the handler itself is unit-testable with the fake-proc pattern already in `ScrcpySession.test.ts`.

harshvardhan60792:
Fixes #481

Added an 'exit' handler on serverProc in ScrcpySession.start() so unexpected exits are logged with a cause (warn) while our own stop() stays quiet (debug). Removed a spurious package-lock.json the draft accidentally generated (this is a pnpm workspace, that file doesn't belong), reverted an unrelated await fix in RelayServer.ts that was bundled in by mistake, and fixed a real bug: the 'stopping' flag was set true in stop() but never reset, so after the first restart every later unexpected exit would silently log at debug forever, defeating the fix.

**Testing:** Added a regression test proving the reset: start, stop cleanly (debug only), restart, exit unexpectedly (must warn) -- this is exactly the case the original draft's tests didn't cover. Could not run the suite: @tapflowio/agent-core is a pnpm workspace sibling package that isn't linked without pnpm installed, and all 6 tests in the file fail identically on that import error, including the 5 pre-existing ones -- confirmed this is a tooling gap, not a regression, same conclusion the original draft reached.

---

## Checklist

- [ ] Tests written and passing
- [ ] No `any`
- [ ] Interface changes land in `agent-core` first
- [ ] No sensitive info (tokens, paths, credentials)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session monitoring to distinguish intentional server shutdowns from unexpected exits.
  - Unexpected server exits now generate warning logs with relevant exit codes and signals.
  - Intentional shutdowns are logged at debug level to reduce unnecessary warnings.
  - Sessions restarted after a clean stop correctly report subsequent unexpected exits.
  - Repeated stops preserve accurate exit classification for delayed process exits.
  - Added coverage for exit handling during stop, restart, and failed-start scenarios.


jo-duchan:
Welcome, and thanks for picking this one up — the change does what #481 asked for, including the part the issue only suggested checking.

Before the specifics: **CI is far less broken than it looks.** Both builds pass, and of 341 tests, 340 pass. One test fails — a new one — and the other three red checks follow from it or are unrelated:

| Check | What it actually is |
|---|---|
| `test (22)` / `test (24)` | One failing test: `logs at debug level (not warn) when serverProc exits after stop()` |
| `ci` | The aggregate gate. It is red because the tests are, not for a separate reason |
| `changeset` | No changeset for a published source change |
| `Vercel` | Deployment authorization for an outside contributor. Nothing to do with your code |

## Why that one test fails

`logger.debug()` does nothing at the default log level, so `console.debug` is never called — which is exactly what the failure says (`Number of calls: 0`).

```ts
// packages/agent-core/src/logger.ts:13
const raw = process.env.LOG_LEVEL ?? 'info'   // default: info === 1
...
debug(msg) {
  if (this.minLevel > LEVELS.debug) return    // 1 > 0 — returns here
  print(console.debug, ...)
}
```

Nothing sets `LOG_LEVEL` in CI, so the level is `info` and `debug` is filtered out. Your `warn` test passes for the same reason in reverse — `warn` is above the threshold.

Worth saying plainly: **you followed the house pattern correctly.** `vi.spyOn(console, ...)` is how this repo asserts on logs (`AndroidAgent.test.ts:2285`, `tunnel-runner.test.ts:59`). It just happens that `debug` is the one level no existing test had reached, so nothing here warned you.

The fix is two lines at the top of the describe block. `ScrcpySession` is imported dynamically inside each test, so setting the env before the first import is enough — `logger` is constructed once at module scope (`ScrcpySession.ts:13`):

```ts
beforeAll(() => vi.stubEnv('LOG_LEVEL', 'debug'))
afterAll(() => vi.unstubAllEnvs())
```

`vi.stubEnv` is already used in the repo (`doctor.test.ts:94`), so this needs nothing new.

One trap to save you an hour: **do not reach for `vi.resetModules()`.** It re-runs the `vi.mock('child_process')` factory, which builds fresh `vi.fn()`s, and the `spawn` your test file imported at the top then points at a mock the module under test no longer uses. Several tests start failing in confusing ways.

And please don't fix it by deleting the `debug` assertion. What remains would be `expect(warnSpy).not.toHaveBeenCalled(...)` alone, which passes just as happily if the exit handler was never attached at all — the shape [`contributing/test-and-guard-coverage.md`](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) exists to warn about. The assertion is worth keeping; it just needs the level raised so it can observe anything.

## On the note in the PR description

> Could not run the suite: `@tapflowio/agent-core` is a pnpm workspace sibling package that isn't linked without pnpm installed … confirmed this is a tooling gap, not a regression.

This one is worth correcting, because it cost you the signal. There is no tooling gap — `pnpm install` at the repo root links the workspace, and then:

```
pnpm --filter @tapflowio/android-agent test
```

runs the file. (Please use exactly that form rather than `npx vitest`; npm rewrites the root `package.json` on its way through and quietly flattens `pnpm.overrides`.) The file also has three pre-existing tests, not five — the other three are yours. In CI, five of the six pass. Treating the local failure as environmental is what hid the one real failure underneath it, and that is the only reason this arrived red.

A changeset is the other red check: `pnpm changeset`, written for someone reading release notes.

## A suggestion on how the expected exit is recognised

This is the one substantive design note, and it is worth doing here rather than later, because as written the change introduces a misleading log line on a path it created.

`start()` kills the server itself when startup fails:

```ts
} catch (e) {
  serverProc.kill()      // deliberate
  this.serverProc = null
  throw e
}
```

That path never sets `stopping`, so every failed `start()` now logs `server process exited unexpectedly (code=- signal=SIGTERM)` about a process **we** killed. #481 asked for the cause of a death to be recorded; inventing the wrong cause is worse than recording none, because an operator will go looking for whatever sent that SIGTERM.

The same flag has a second, quieter version of the problem: `stop()` sets it, `start()` clears it, and if a slow-to-die process emits `'exit'` after the next `start()` has already cleared the flag, the old process is reported as an unexpected death. In practice `start()` does an `adb push` first so the old exit almost always wins the race — but it is the same defect.

Comparing process identity closes all of it, and needs no reset at all:

```ts
/** The process we killed on purpose — compared by identity so a late exit from an old
 *  process is never blamed on the one running now. */
private stoppedProc: ChildProcess | null = null
```

```ts
serverProc.on('exit', (code, signal) => {
  const detail = `code=${code ?? '-'} signal=${signal ?? '-'}`
  if (this.stoppedProc === serverProc) logger.debug(`server process exited (${detail})`)
  else logger.warn(`server process exited unexpectedly (${detail})`)
})
```

with `this.stoppedProc = this.serverProc` in `stop()` before the kill, and `this.stoppedProc = serverProc` in that `catch` before its kill. The `this.stopping = false` line in `start()` then goes away — and with it the whole class of bug you found, since there is no longer a flag that can be left in the wrong position.

I checked this against your existing tests: the warn test emits `'exit'` before the 1500 ms timer resolves, so the `catch` has not run yet and it still sees an unexpected exit. It should stay green.

## What is good here

Three things, so they don't get buried under the above.

You kept to the issue's scope — #481 was explicit that the handler must not drive teardown, since `pumpVideo` already owns that, and you left it alone. You also acted on the last line of the issue, which only suggested *considering* whether expected and unexpected exits should be distinguished. And you found the unreset `stopping` flag on your own and wrote a regression test that genuinely fails without the fix, which is more than the change was asked to do.

Once the test and the changeset are sorted I will take another pass. The suggestion above is a suggestion — if you disagree with it, say so and we will talk it through rather than you just applying it.

harshvardhan60792:
Thanks for the detailed review! I've updated the PR with your suggestions:

- Replaced the boolean \stopping\ flag with process identity comparison (\stoppedProc === serverProc\) to avoid false alarms when startup fails.
- Stubbed \LOG_LEVEL\ to \debug\ during test execution so the debug assertions don't get filtered out.
- Added the changeset file for versioning.

Let me know if this looks good to you now!

packages/android-agent/src/scrcpy/ScrcpySession.ts
```diff
@@ -113,6 +116,16 @@ export class ScrcpySession {
     // Without a handler, an unhandled 'error' event (e.g. spawn() failing with ENOENT/EACCES,
     // or EPERM from kill()) throws and crashes the whole agent — taking down every device it manages.
     serverProc.on('error', (e) => logger.error(`server process: ${e.message}`))
+    // Log why the server exited so restart loops have a recorded cause. Expected exits (our own
+    // stop() sets this.stopping) are debug-level noise; unexpected ones are warnings.
```
jo-duchan: Small leftover from the rework: `this.stopping` was the first commit's field and doesn't exist any more, so this sentence now points a reader at something they won't find. The logic underneath it is right — it's only the description that got left behind.

```suggestion
    // Log why the server exited so restart loops have a recorded cause. An exit from the
    // process we killed on purpose is debug-level noise; anything else is a warning.
```

packages/android-agent/src/scrcpy/ScrcpySession.ts
```diff
@@ -133,13 +146,15 @@ export class ScrcpySession {
...
     } catch (e) {
+      this.stoppedProc = serverProc
       serverProc.kill()
       this.serverProc = null
       throw e
     }
   }
 
   stop(serial: string): void {
+    this.stoppedProc = this.serverProc
```
jo-duchan: Please read this one as a suggestion rather than a defect — I went looking for a way to trigger it and I don't think there is one today. `cleanupDeviceState` (`AndroidAgent.ts:533`) and `restartVideoStream` (`AndroidAgent.ts:858`) each set `state.scrcpySession = null` immediately after calling `stop()`, and `state.scrcpySession` is only assigned *after* `start()` resolves — so `stop()` never runs twice on the same session, and never runs on one whose `start()` threw.

What makes me want the guard anyway is that the class is currently relying on its callers to keep that true. If `stop()` is ever reached with `serverProc` already `null`, this line overwrites the marker with `null` — and then the exit event from the process we *did* just kill arrives, finds no match, and gets logged as `unexpectedly`. That's precisely the false positive the identity check exists to prevent, and it would show up on the failure path, which is the one where a misleading log costs the most.

One line, and the invariant lives in the class that owns it rather than in two call sites that happen to be written correctly.

```suggestion
    // Keep the previous marker when there is nothing to mark — a stop() that finds serverProc
    // already null would erase it, and the exit we caused would come back as unexpected.
    if (this.serverProc) this.stoppedProc = this.serverProc
```

jo-duchan (commented):
Thanks for the rework — this is a better design than the first pass and I want to say why, so it doesn't get lost in the diff.

Swapping the `stopping` boolean for a process-identity check removes the reset entirely rather than fixing it. A boolean makes "clear the flag" something a future caller has to remember; `this.stoppedProc === serverProc` has no state to reset, and as a bonus it attributes a late exit to the process it actually came from instead of blaming whichever one is running now. Marking the process in `start()`'s `catch` before killing it is the same instinct applied to the failure path — that's the case I'd have expected a first pass to miss, and you didn't.

The tests hold up too, and not by accident. Both of the "did **not** warn" assertions are paired with a live positive assertion in the same test — `debugSpy` in the clean-stop case, the second `warnSpy` in the restart case — so if the logger ever went silent the positive half would fail first. That's exactly the shape [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) asks for, and it's the part most PRs get wrong.

Two small inline notes below — one is a leftover sentence, the other is defensive rather than a live bug. Neither is a blocker.

---

### The `changeset` check needs a root `CHANGELOG.md` entry

This is the one thing genuinely blocking the merge, and it's on us for not making it obvious — the per-package changeset you added is correct and needed, but there's a second, separate file. The root `CHANGELOG.md` is what a self-hoster reads to decide whether to upgrade, and nothing generates it.

It applies here because the warning fires at the default `LOG_LEVEL=info`: someone running an Android agent will now see a line in their logs that wasn't there before. That's observable, so it earns an entry. Please add this under the existing `### Fixed` in `## [Unreleased]`:

```markdown
- **A restarting Android video stream now says what happened to it** ([#481](https://github.com/jo-duchan/tapflow/issues/481)). When the scrcpy server process on a device exits, tapflow already recovers — the socket tears down, the reader finishes, the stream restarts. What it did not do was record why the process went away, so a device that restarted once and a device restarting in a loop left the same log behind. The exit code and signal are now written down, and a teardown tapflow asked for stays at debug, so the line appears only when something happened that nobody asked for.
```

Reword it however you like — that's the house voice, not a requirement.

### The `build (linux/amd64)` failure is not yours

Worth saying plainly so you don't go looking. The log reads:

```
prebuild-install warn install read ECONNRESET
→ node-gyp rebuild → gyp ERR! find Python ... Could not find any Python installation
```

The prebuilt `better-sqlite3` binary download was cut off mid-transfer, so the install fell back to compiling from source in an image that has no Python. Nothing to do with this diff — the same job passed on your first commit, and arm64 passed on this one. I'll re-run it. (It's a real fragility in our Docker build, and it's ours to fix, not yours.)

### One last nit

The PR description still explains the first commit's `stopping` flag that "was set true in `stop()` but never reset" — the rework replaced that mechanism, so the body now describes code that isn't here. Worth a refresh so a future reader of the merge commit isn't sent looking for a field that never shipped.

jo-duchan:
@harshvardhan60792 A nudge on this one — it's been sitting a few days, and I think my last comment is why: I said the rework was a better design, which reads like it's done. It isn't quite.

**The `changeset` check is red.** Your changeset file is fine — the gate is asking for the *root* `CHANGELOG.md`. Add an entry under `## [Unreleased]` → `### Fixed`; the job output spells out the alternative if you'd rather mark the change internal instead.

**And one stale comment.** `ScrcpySession.ts` still says `our own stop() sets this.stopping` above the exit handler, where the field is now `stoppedProc`. CodeRabbit flagged it and it's the one note that didn't get picked up — worth closing, since killing that boolean was the point of the rework.

Neither needs a rethink. Ping me once they're in and I'll take another look.

jo-duchan:
@harshvardhan60792 Checking in again, and this time with an offer rather than another nudge.

Both of your PRs are finished work sitting on trivia. Here it is the root `CHANGELOG.md` entry and
the comment above the exit handler that still says `this.stopping` where the field is now
`stoppedProc`. On #687 it is the same CHANGELOG entry plus a Summary in the PR body. That is four
lines across two branches, and none of it is the part that took you any thought.

So: if it is easier for me to add those lines than for you to come back to it, say the word — or just
say nothing by the end of this week and I will take that as a yes. I would push them to your branches
and merge, and both stay your commits and your contribution. The changelog line is a sentence I can
write from your own diff; I would rather do that than leave two good fixes unmerged over punctuation.

If you would rather finish them yourself, that is better and I am in no rush beyond this week — reply
here and I will hold off entirely.

One thing worth saying plainly, since two rounds of my comments have been about process: the actual
engineering in both of these was good. Killing the `stopping` boolean in favour of comparing process
identity is a better design than what the issue asked for, and you found the unreset-flag bug on your
own and wrote a test that fails without the fix. That is why I keep coming back to these rather than
closing them.

harshvardhan60792:
Added the CHANGELOG.md entry under Unreleased and updated the comment in ScrcpySession.ts to reference stoppedProc.

jo-duchan:
Picking this back up. Both notes from my last review are addressed — the `this.stopping` sentence at the top of the exit handler now describes `stoppedProc`, which is the mechanism that actually shipped. I merged your branch onto `main` and ran it: `ScrcpySession` is 6/6 green.

Two things left, both mechanical.

### The changelog entry is in the wrong section

It landed under `### Added`, and this is a fix — an exit that already happened is now recorded, rather than a new capability. Please move it under `### Fixed` in `## [Unreleased]`.

The wording is yours to keep; I only asked for a specific section, and that part I do mean. `Added` and `Fixed` are what a self-hoster scans to decide whether an upgrade is worth doing, and a fix filed as a feature reads as noise in both directions.

### Rebase onto `main`

`CHANGELOG.md` conflicts — only that file, the code merges clean. It is the ordinary collision between two PRs both writing to `## [Unreleased]`, not anything you did.

---

Nothing else. Once those two are in this is ready from my side.

jo-duchan:
Picked this up rather than leave it waiting any longer — you'd done the work and it was sitting on a changelog heading. Three commits on your branch:

- merged `main` in (that was the CHANGELOG conflict — your branch predates the `### Fixed` section the entry belongs under)
- moved the entry to `### Fixed` and dropped the now-empty `### Added`
- applied the `stop()` guard from the review, with a test

On the guard: I went looking for a caller that reaches it and agree there isn't one — `cleanupDeviceState` and `restartVideoStream` both null the session right after stopping it, and the session is only assigned after `start()` resolves. It's in because the invariant belongs in the class, and it has a test because a guard nothing can fail is decoration; removing the `if` turns it red.

Before pushing I mutated the exit handler five ways and all five die, including replacing `stoppedProc === serverProc` with a truthiness check — the exact bug the identity comparison is there to stop. 360 android-agent tests green, typecheck and lint clean.

Thanks for the patience on this one, and sorry it sat.

PR #632, opened 2026-08-22
chore: add scripts test command

### LINKED ISSUE #616: No approved way to run the scripts suite alone, and the workaround corrupts the root manifest
The root `test` script is `vitest run --config scripts/vitest.config.mjs && pnpm -r test --if-present`, so the cross-package static checks under `scripts/__tests__/` run first and every package suite runs after. There is no script that runs only the first half.

That matters more than it sounds. Those checks are what fail when a protocol message is added, a count pin drifts, or a guard stops guarding — exactly the checks you iterate against while making such a change. Today the only approved way to see them is `pnpm test`, which takes about five minutes because it runs all 2900-odd tests behind them.

The unapproved way is `npx vitest run --config scripts/vitest.config.mjs`, and the root `AGENTS.md` forbids it for a good reason: npm rewrites the root `package.json` on its way through, collapsing `pnpm.overrides` to `pnpm: {}` and rewriting `pnpm-lock.yaml` beside it, reporting none of it. `git checkout HEAD -- package.json pnpm-lock.yaml` puts both back, but only if you notice.

**It has been reached for twice in one program of work**, by someone who had read the rule and written it into a review record the first time. A prohibition that is reached for repeatedly by someone who knows it is usually missing an affordance rather than a reminder.

Proposal: add a root script — `"test:scripts": "vitest run --config scripts/vitest.config.mjs"` — and have `test` call it, so the two cannot drift. Then the AGENTS.md rule can name the alternative instead of only the prohibition, which is the difference between a rule you follow and a rule you work around.

Worth checking at the same time whether the per-package equivalent has the same gap: `pnpm --filter <pkg> test <path>` does work for a single file, so this may be the only hole.

ML642:
## Summary

Adds `pnpm test:scripts` as the documented entry point for the root scripts suite.
The root `test` command now delegates to it, preserving the existing command and test order.

I also checked the per-package targeted test command mentioned in #616. `pnpm --filter <pkg> test <path>` forwards a single test-file path to Vitest for package-level test scripts, so this PR only addresses the missing root scripts-suite affordance.

## Checklist

- [ ] Tests written and passing
- [x] No `any`

## Related `.work/` docs
- `.work/reviews/feature__scripts-test-suite-616.md` — independent adversarial review, no findings.

Closes #616

## Summary by CodeRabbit

* **Documentation**
  * Updated testing guidance with instructions for running the repository’s scripts suite.
  * Clarified the recommended commands for package-level and cross-package tests.

* **Chores**
  * Improved the default test workflow to run scripts checks before package tests.
  * Updated test configuration guidance to reflect the standard scripts-suite command.

Closes `#616`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -201,7 +201,7 @@ gates went on: `AgentRegistry.test.ts` declared `implements DeviceAgent` while m
 sites naming a type violated.
 
 ### Test Hygiene
-Tests run through `pnpm --filter <pkg> test`, never `npx vitest` — not even from inside the package directory, and not for a single file. npm rewrites the root `package.json` on its way through and collapses `pnpm.overrides` to `pnpm: {}`, leaving `pnpm-lock.yaml` rewritten beside it, and it reports none of that. Reviewing #474 cost exactly this: one `npx vitest` on one test file, and the entire override block was gone with only `git status` to say so. `git checkout HEAD -- package.json pnpm-lock.yaml` puts both back — check `git diff` on them first if you were editing either on purpose, since that discards everything uncommitted in both.
+Tests run through `pnpm --filter <pkg> test`; use `pnpm test:scripts` for the root scripts suite. Never use `npx vitest` — not even from inside the package directory, and not for a single file. npm rewrites the root `package.json` on its way through and collapses `pnpm.overrides` to `pnpm: {}`, leaving `pnpm-lock.yaml` rewritten beside it, and it reports none of that. Reviewing #474 cost exactly this: one `npx vitest` on one test file, and the entire override block was gone with only `git status` to say so. `git checkout HEAD -- package.json pnpm-lock.yaml` puts both back — check `git diff` on them first if you were editing either on purpose, since that discards everything uncommitted in both.
 
 After running tests (especially repeated or looped runs), always check for zombie vitest processes and kill them:
 ```bash
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -136,6 +136,12 @@ All packages:
 pnpm test
 ```
 
+Repository scripts suite (cross-package static checks):
+
+```sh
+pnpm test:scripts
+```
+
 A specific package:
 
 ```sh
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
   },
   "scripts": {
     "build": "pnpm --filter @tapflowio/protocol build && pnpm --filter @tapflowio/agent-core build && pnpm --filter @tapflowio/dashboard build && pnpm --filter @tapflowio/flow-runner build && pnpm --filter @tapflowio/relay build && pnpm --filter @tapflowio/ios-agent build && pnpm --filter @tapflowio/android-agent build && pnpm --filter tapflow build && pnpm --filter @tapflowio/mcp-server build",
-    "test": "vitest run --config scripts/vitest.config.mjs && pnpm -r test --if-present",
+    "test": "pnpm test:scripts && pnpm -r test --if-present",
+    "test:scripts": "vitest run --config scripts/vitest.config.mjs",
     "lint": "eslint scripts && pnpm -r --if-present lint",
     "typecheck": "tsc -b && pnpm -r --if-present typecheck",
     "prepare": "lefthook install || true",
--- a/scripts/vitest.config.mjs
+++ b/scripts/vitest.config.mjs
@@ -4,7 +4,7 @@ import { dirname, resolve } from 'path'
 
 // Deliberately NOT at the repo root. A config file there is inherited by every package that has
 // none of its own — all of them do — and they then run with this `include` and find no tests.
-// Invoked explicitly: `vitest run --config scripts/vitest.config.mjs`.
+// Invoked explicitly: `pnpm test:scripts`.
 //
 // `root` is resolved from this file, not from the cwd: vitest resolves a relative `root` against
 // the working directory, so '..' pointed one level above the repo.


jo-duchan (commented):
Welcome, and thank you — this is a notably clean first contribution, so let me be specific about what it gets right rather than waving at it.

You built the mechanism the issue asked for rather than a second copy of it. `test` calling `test:scripts` means the two **cannot** drift, which is the whole point of #616's proposal and the part that is easy to miss: pasting the vitest invocation into a new script would have satisfied the letter of the request and left the exact hazard it exists to remove. The script name matches the repo's other colon-scripts (`dev:down`, `changeset:check`, `overrides:audit`), and the AGENTS.md line now names the alternative instead of only the prohibition — which is the sentence #616 was really asking for.

The PR itself is put together properly too, which I mention because it is rarer than it should be: the template is filled in rather than pasted over, the `no-changeset` marker carries a reason on its own line, and the review record is named. Leaving **Tests written and passing** unchecked was the right call, not an omission — the delegation *is* the anti-drift guarantee here, and a test asserting that one script calls another would only restate the diff.

Three things below. One is a small fix I'd like in this PR, one is a suggestion, and one is good news.

### The config's own comment now points at the old way

`scripts/vitest.config.mjs:7` still says:

```js
// Invoked explicitly: `vitest run --config scripts/vitest.config.mjs`.
```

That line sits inside the very file this PR gives a name to, and it is the first thing anyone opening the config reads. AGENTS.md puts "prose the change just made false" squarely in the fix-it-here column, and it is one line:

```js
// Invoked explicitly: `pnpm test:scripts`.
```

(It is outside the current diff, so I could not attach this as a suggestion — sorry for the copy-paste.)

### CONTRIBUTING.md's `## Tests` section has the same gap

`CONTRIBUTING.md` lists **All packages** (`pnpm test`) and **A specific package** (`pnpm --filter …`), and the scripts suite is neither. AGENTS.md is where agents and maintainers look; CONTRIBUTING.md is where a human opening the repo for the first time looks — and #616 is a discoverability issue at heart, about an affordance people reach past because they cannot find it. Four lines there would finish the job:

````markdown
The root scripts suite (cross-package static checks):

```sh
pnpm test:scripts
```
````

Your call, and I would not block on it — but the affordance only works where someone looks for it.

### The pre-existing failures aren't real — good news

The body says `pnpm test:scripts` reproduces scripts-suite failures also present on upstream `main`. I ran the suite on a built checkout of `main` just now:

```
Test Files  29 passed (29)
     Tests  387 passed (387)
```

What you almost certainly hit is the build prerequisite: `testsReadSource.test.mjs` plants a marker in `packages/relay/dist/index.js` to prove test imports resolve to source rather than `dist`, and it throws `packages/relay/dist/index.js is missing — run \`pnpm build\` first` on a checkout that has not been built. `pnpm build` once, and the suite goes green. So that caveat can come out of the body — your change is cleaner than you gave it credit for.

There is a real observation hiding in it, though, and it is worth more than the caveat was: the affordance this PR adds still has an unstated `pnpm build` prerequisite, and #616 wanted a fast iteration loop. Not this PR's job to solve, but if you noticed it the hard way, a follow-up issue saying so would be a genuinely useful second contribution.

### One question

#616 closes with "worth checking at the same time whether the per-package equivalent has the same gap: `pnpm --filter <pkg> test <path>` does work for a single file, so this may be the only hole." Did you get a chance to look? A sentence either way in the body would close the issue cleanly rather than leaving the question open behind it.

ML642:
Thanks for the detailed review.

I pushed the requested follow-up changes:

- Updated the comment in `scripts/vitest.config.mjs` to point to `pnpm test:scripts`.
- Added `pnpm test:scripts` to the `CONTRIBUTING.md` Tests section as the root scripts suite / cross-package static checks command.

I also checked the per-package targeted test command mentioned in #616. `pnpm --filter <pkg> test <path>` forwards a single test-file path to Vitest for package-level test scripts, so this PR only addresses the missing root scripts-suite affordance.

jo-duchan (approved):
All four landed, and cleanly. Approving — thank you.

The config comment now points at the name the config was given, `CONTRIBUTING.md` carries the affordance where a first-time contributor will actually go looking for it, and the caveat about upstream failures is out of the body. Putting the scripts suite between **All packages** and **A specific package** was a better call than the wording I suggested: the section now reads broadest to narrowest.

Thank you especially for chasing #616's closing question instead of letting it merge open behind you. I checked your answer independently and it holds — every package's `test` script is a bare `vitest run`, so a trailing path is forwarded straight through to vitest, and the root scripts suite really was the only hole. That is the kind of loose end that normally survives until someone reopens the issue six months later to ask.

The part worth being most pleased about was right in your first commit, though: `test` delegating to `test:scripts` means the two **cannot** drift. Pasting the vitest invocation into a second script would have satisfied the request word for word and quietly preserved the exact hazard it was filed to remove.

One housekeeping note, and nothing for you to do about it: fork PRs need a maintainer to approve each CI run, so the checks on your latest commit are waiting on us rather than on you. The merge follows once they are green.

If you want a second one to pick up: the follow-up I floated earlier is sharper now than when I raised it, precisely because of this PR. `CONTRIBUTING.md` now advertises `pnpm test:scripts` to newcomers, and on a fresh clone it still fails until `pnpm build` has been run once — `testsReadSource.test.mjs` needs a built `packages/relay/dist/index.js` to plant its marker in. You are the person who hit that, so you would write that issue better than anyone else could. Entirely optional.

And if you have enjoyed working on tapflow, a star on the repo would genuinely mean a lot. No obligation at all — this contribution stands on its own either way.

PR #633, opened 2026-08-23
fix(scripts): read the changelog:internal marker the way the other two markers are read

harshvardhan60792:
This finishes #604.

`proseLines` already filters out lines where a marker could be quoted (such as fenced code blocks, four-space-indented blocks, and tab-indented blocks) before yielding lines. Consequently, parsing `raw` inside `changelogEntryOwed` was redundant, and differed from the sibling markers (`no-changeset` and `Backfills:`) which already read the trimmed `line`.

This change updates the `changelog:internal` marker to align with the other two markers by matching on the trimmed `line`.

Note: This trim deliberately widens the opt-out behavior to include markers indented by 1 to 3 spaces (e.g. inside a list-item continuation line), bringing parity with the other two parser consumers. Added unit tests to pin the behavior (confirming that 1-3 spaces are accepted, but four spaces representing an indented code block are not released).

## Summary by CodeRabbit

* **Bug Fixes**
  * Changelog opt-out markers are now recognized when surrounded by permitted whitespace.
  * Markers inside indented code blocks no longer incorrectly bypass the changelog-entry requirement.

* **Tests**
  * Added coverage for whitespace-tolerant markers.
  * Added coverage confirming that markers in four-space-indented code blocks are treated as code rather than opt-out instructions.

### DIFF
--- a/scripts/__tests__/changesetGateAccuracy.test.mjs
+++ b/scripts/__tests__/changesetGateAccuracy.test.mjs
@@ -87,6 +87,7 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
     'mixed-fence.md': cs('"@tapflowio/relay": patch') + '\n```md\n~~~\n<!-- changelog: internal -->\n~~~\n```\n',
     'after-fence.md': cs('"@tapflowio/protocol": patch') + '\n```md\nexample\n```\n<!-- changelog: internal -->\n',
     'indented-fence.md': cs('"@tapflowio/protocol": patch') + '\n    ```md\n<!-- changelog: internal -->\n    ```\n',
+    'indented-marker.md': `${cs('"@tapflowio/protocol": patch')}\n  <!-- changelog: internal -->  \n`,
   })[f]
 
   it('owes one for a plain changeset', () => {
@@ -113,6 +114,9 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
   })
   it('is released by an unindented marker between four-space-indented fence-like lines', () => {
     expect(changelogEntryOwed(['indented-fence.md'], read)).toEqual([])
+  })
+  it('is released by the marker even with leading/trailing spaces', () => {
+    expect(changelogEntryOwed(['indented-marker.md'], read)).toEqual([])
   })
 })
 
--- a/scripts/__tests__/changesetGateAccuracy.test.mjs
+++ b/scripts/__tests__/changesetGateAccuracy.test.mjs
@@ -88,6 +88,7 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
     'after-fence.md': cs('"@tapflowio/protocol": patch') + '\n```md\nexample\n```\n<!-- changelog: internal -->\n',
     'indented-fence.md': cs('"@tapflowio/protocol": patch') + '\n    ```md\n<!-- changelog: internal -->\n    ```\n',
     'indented-marker.md': `${cs('"@tapflowio/protocol": patch')}\n  <!-- changelog: internal -->  \n`,
+    'indented-code-marker.md': `${cs('"@tapflowio/protocol": patch')}\nThe gate takes an opt-out:\n\n    <!-- changelog: internal -->\n`,
   })[f]
 
   it('owes one for a plain changeset', () => {
@@ -117,6 +118,13 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
   })
   it('is released by the marker even with leading/trailing spaces', () => {
     expect(changelogEntryOwed(['indented-marker.md'], read)).toEqual([])
+  })
+  // The trim widened what counts as a marker, so what still must NOT count is worth pinning:
+  // four spaces makes an indented code block -- someone quoting the syntax -- and `proseLines`
+  // is now the only thing between that and a silent opt-out, since the match itself no longer
+  // sees the indentation.
+  it('is not released by a marker inside a four-space-indented code block', () => {
+    expect(changelogEntryOwed(['indented-code-marker.md'], read)).toEqual(['indented-code-marker.md'])
   })
 })
 
--- a/scripts/check-changeset.mjs
+++ b/scripts/check-changeset.mjs
@@ -403,8 +403,8 @@ export function ignoredOnlyChangesets(files, ignored, read) {
  */
 export function changelogEntryOwed(files, read) {
   return files.filter((f) => {
-    for (const { raw } of proseLines(read(f), { skipFrontmatter: true })) {
-      if (/^<!--\s*changelog:\s*internal\b.*-->$/.test(raw)) return false
+    for (const { line } of proseLines(read(f), { skipFrontmatter: true })) {
+      if (/^<!--\s*changelog:\s*internal\b.*-->$/.test(line)) return false
     }
     return true
   })


scripts/check-changeset.mjs
```diff
@@ -403,8 +403,8 @@ export function ignoredOnlyChangesets(files, ignored, read) {
  */
 export function changelogEntryOwed(files, read) {
   return files.filter((f) => {
-    for (const { raw } of proseLines(read(f), { skipFrontmatter: true })) {
-      if (/^<!--\s*changelog:\s*internal\b.*-->$/.test(raw)) return false
+    for (const { line } of proseLines(read(f), { skipFrontmatter: true })) {
+      if (/^<!--\s*changelog:\s*internal\b.*-->$/.test(line)) return false
```
jo-duchan, line 407: No change requested here — I just want the reasoning on the record next to the line, since a two-word diff is the easiest kind to misread later.

This is the whole of it: `proseLines` has already thrown away every line that a marker could be *quoted* in — fenced blocks, four-space-indented blocks, tab-indented blocks — before it yields anything. So by the time this test runs, indentation carries no meaning any more, and matching on `raw` was asking a question that the generator had already answered. The two sibling markers concluded the same thing; this one just hadn't caught up.

Worth noting for whoever reads this next: with this change, nothing in the file destructures `raw` from `proseLines` any longer. I'd leave the generator yielding it — a future caller may well want it — but it's a fair signal that `line` is the contract all three consumers were reaching for.

scripts/__tests__/changesetGateAccuracy.test.mjs
```diff
@@ -87,6 +87,7 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
     'mixed-fence.md': cs('"@tapflowio/relay": patch') + '\n```md\n~~~\n<!-- changelog: internal -->\n~~~\n```\n',
     'after-fence.md': cs('"@tapflowio/protocol": patch') + '\n```md\nexample\n```\n<!-- changelog: internal -->\n',
     'indented-fence.md': cs('"@tapflowio/protocol": patch') + '\n    ```md\n<!-- changelog: internal -->\n    ```\n',
+    'indented-marker.md': `${cs('"@tapflowio/protocol": patch')}\n  <!-- changelog: internal -->  \n`,
```
jo-duchan, line 90: Good fixture — one string covering both the leading and the trailing whitespace is the right economy.

Could you add a second one next to it? Suggestion below. It pairs with the test further down; see that comment for why.

```suggestion
    'indented-marker.md': `${cs('"@tapflowio/protocol": patch')}\n  <!-- changelog: internal -->  \n`,
    'indented-code-marker.md': `${cs('"@tapflowio/protocol": patch')}\nThe gate takes an opt-out:\n\n    <!-- changelog: internal -->\n`,
```

scripts/__tests__/changesetGateAccuracy.test.mjs
```diff
@@ -114,6 +115,9 @@ describe('a changeset that owes the root CHANGELOG an entry', () => {
   it('is released by an unindented marker between four-space-indented fence-like lines', () => {
     expect(changelogEntryOwed(['indented-fence.md'], read)).toEqual([])
   })
+  it('is released by the marker even with leading/trailing spaces', () => {
+    expect(changelogEntryOwed(['indented-marker.md'], read)).toEqual([])
+  })
```
jo-duchan, line 121: This is the one thing I'd like added before merge, and it's about what the change quietly takes away rather than what it adds.

Before this PR there were two independent things stopping a marker inside a four-space-indented code block from silently opting a change out: `proseLines` skipping the line, **and** the `raw` match refusing to see past the indentation. Your change removes the second one — correctly, that's the point — which leaves `proseLines` load-bearing and alone. Nothing in the suite pins it at this call site today: `indented-fence.md` asserts the opposite direction (an *unindented* marker between indented fence-like lines, which should release).

That's the shape [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) warns about — a guard whose remaining protection is untested reads exactly like one that is covered. It's cheap to close, and this PR is the right place because this PR is what opens it.

I ran both fixtures against your branch to make sure I'm not sending you after a red test: `indented-marker.md` returns `[]` and `indented-code-marker.md` returns `['indented-code-marker.md']`, as written.

```suggestion
  it('is released by the marker even with leading/trailing spaces', () => {
    expect(changelogEntryOwed(['indented-marker.md'], read)).toEqual([])
  })
  // The trim widened what counts as a marker, so what still must NOT count is worth pinning:
  // four spaces makes an indented code block -- someone quoting the syntax -- and `proseLines`
  // is now the only thing between that and a silent opt-out, since the match itself no longer
  // sees the indentation.
  it('is not released by a marker inside a four-space-indented code block', () => {
    expect(changelogEntryOwed(['indented-code-marker.md'], read)).toEqual(['indented-code-marker.md'])
  })
```

jo-duchan (commented):
Thanks for picking up another one, and for going after a real inconsistency rather than an easy one. Up front, so it isn't buried: **the change itself is correct.** Almost everything below is about the framing rather than the code.

I checked the three callers that share `proseLines`, and you found the odd one out. Measured against your branch:

```
extractReason,  marker indented 2 spaces in a list : "because"
extractReason,  marker with a trailing space       : "because"
parseBackfills, indented 2 spaces                  : [413]
parseBackfills, trailing space                     : [413]
```

`extractReason` and `parseBackfills` were both already reading the trimmed `line`. `changelogEntryOwed` was alone on `raw`, so someone who learned the rule from one marker got a different answer from the third — for three markers that all do the same job, which is switching a gate off. `line` is the right side to converge on.

I also ran the dangerous cases to be sure nothing reopened:

```
DEMANDS entry  marker inside a ``` fence
DEMANDS entry  marker in a four-space-indented code block
DEMANDS entry  marker in a tab-indented block
DEMANDS entry  marker in a blockquote
RELEASES       marker indented 2 spaces          (new)
RELEASES       marker with a trailing space      (new)
RELEASES       plain marker
```

The fence stays shut, which is the one that matters. All 32 tests in `changesetGateAccuracy.test.mjs` pass on your head, and the tidiest evidence that you're right is this: after your change **no caller destructures `raw` at all** any more. `proseLines` yields it and nothing reads it. That is what drift looks like after it has been cleaned up.

### The framing does need to change

`Fixes #560` doesn't hold. **#560 was closed on 2026-08-20 by #604** (`fix(scripts): ignore changelog markers in code fences`, `7361266`) — which is the PR that introduced `proseLines` in the first place. The title here is #560's own title, truncated, so as written this PR announces a fix for a bug that was already fixed three days before it was opened. Someone reading the merge commit in a month gets sent looking for a hole that isn't there.

The good news is that the real justification is stronger than the one on the tin, and it is sitting inside #604. That PR's summary says:

> The **shared** parser also prevents quoted markers in nested fences from bypassing the `no-changeset` and `Backfills:` gates.

…and then left one of the three consumers on `raw` — which reads like the old `/m` regex's semantics surviving the move intact, rather than a decision anyone made. **This finishes #604; it does not fix #560.** Retitling along those lines — something like `scripts: read the changelog:internal marker the way the other two markers are read` — and rewriting the body to match would make it land properly. `Fixes #560` should come out. There's no open issue to close here and that's fine; a PR is allowed to just be a follow-up.

### One thing worth saying out loud

The trim genuinely widens what counts as an opt-out: a marker indented one to three spaces — a list-item continuation line, say — now releases a change where it previously didn't. I think that's the right call, precisely because the two sibling markers already behave that way. But #560's own text listed "an indented marker in a list or blockquote" among the near-misses that fail in the **safe** direction, so removing that margin shouldn't be something a reviewer discovers by testing. Please name it in the body as a deliberate widening rather than leaving it implicit.

Two inline notes below — one is a test I'd like added before this merges.

### Two things that are on us, not you

CI hasn't run here at all: both workflows are sitting at `action_required`, waiting for a maintainer to approve a run from a fork. Nothing you can do from your side, and we'll sort it out.

And no changeset is needed for this one — `shipsToUsers('scripts/check-changeset.mjs')` is `false`, so the gate won't ask. (#604 added a `no-changeset` marker anyway, which is belt and braces rather than a requirement.)

harshvardhan60792:
Done! I've made the requested updates:
- Re-titled the PR and updated the description to frame it as finishing #604 (removing the `Fixes #560` text) and explicitly documenting the 1-3 space indentation widening behavior.
- Added the `indented-marker.md` and `indented-code-marker.md` mock fixtures.
- Added the regression test to pin that a marker inside a four-space-indented block is not released.

Let me know if there's anything else needed!

Issue #639, opened 2026-08-23
The agent cannot see whether layer 1 (the content filter) is actually enforcing


jo-duchan:
`SimulatorNetwork.state()` decides `available` from the dylib's verdict, which is evidence about **layer 2** — the injected library inside the app. Nothing in the agent can see whether **layer 1**, the host content filter, is working.

#607 closed the half that was reachable from an exit code: the container app now exits with a distinct non-zero code for each failure (activation, load, save, an approval that never came, an install that needs a reboot) instead of exiting 0 through all of them, and the README carries the table. That is genuinely useful — a user who declines the filter in System Settings now produces a failure the agent can see.

What it cannot cover is the state after a successful run. Zero means the save was accepted; the framework hands `vendorConfiguration` to the running provider afterwards, on its own schedule, with no acknowledgement coming back. A provider that was killed, never approved, or is running an older bundle leaves the agent reporting `available: true` over a kernel that is dropping nothing — a device drawn as offline that is not, which is the sign-off failure this feature exists to avoid.

Needs an artefact the agent can read, and that is a design decision rather than a fix: the provider could write a heartbeat with the rule it currently holds, or the host could re-read the configuration back and compare, or the agent could probe. Each has a different failure mode when the provider is merely slow.

Touching: `packages/ios-agent/ios-netfilter/Extension/Provider.swift`, `packages/ios-agent/src/SimulatorNetwork.ts`.

jo-duchan:
Done in #666 (merged). The provider vends a read-only XPC `ping` answering `{enforcing, rule, pid}`, the host binary exposes it as `--confirm`, and `setOffline` refuses unless the answer says `enforcing` and names this device. `enforcing` is a field of its own because `rule: []` cannot tell an idle provider from a stopped one — measured on a `--off` provider, which stays alive and answers in 16ms.

Enforcement that stops *after* the fact is covered too: the provider pulses once a second while enforcing, and the agent reports `enforcement-lost` when that stops. Measured: a killed provider leaves the kernel passing traffic for about 5.8s, 23–27 requests per occurrence.

Issue #607, opened 2026-08-20
Feature to simulate no cellular or internet connectivity


keshav-cilio:
### What problem does this solve?

Feature to simulate no cellular or internet connectivity

### What would you like tapflow to do?

Feature to simulate no cellular or internet connectivity

### Which part does this touch?

iOS (simulator)

### Contribution

- [ ] I would be willing to work on this

---

## What is left, and what closes this

Every issue below carries `Parent: #607`, so this list is the whole remaining surface and nothing
else counts as part of it. Shipped so far: the three-layer offline path (#618, #638, #639, #644),
the reason set and XPC confirmation (#666), layer-3 swallow and verdict classification (#669), and
distribution plus the CLI that installs it (#675).

**Closing this issue needs these four, in order:**

- [ ] #653 — the injected library writes its verdict non-atomically
- [ ] #658 — the filter rule has no owner
- [ ] One PR for the surface a tester actually touches: #617, #620, #625, #628, #634
- [ ] #654 (the state file proves the rule arrived, not that flows drop) and #629 (probe whether the
      hooks still install on this Xcode)

**Adjacent, and deliberately not blocking this** — each is real, none of them is part of "an iOS
simulator can be taken off the network":

- #662 `agent:register` carries no version · #667 `network:state` cannot say "unknown"
- #668 a swallowed status-bar failure on the online path
- #670 what happens if the Developer ID is revoked · #671 two checkouts, one `/Applications`
- #672, #674 (artifact guard, `Parent: #647`) · #676 the XPC listener does not validate its peer

jo-duchan:
Thanks for filing this. I agree it matters, and it's going into an upcoming update.

Scoping note: the first version will be a plain on/off toggle. Throttling presets like 3G or slow 4G are a follow-up.

jo-duchan:
Reopening — this was closed prematurely. The network on/off toggle has shipped for Android, and on iOS the app already sees itself as offline, but actually cutting iOS traffic (the host-side piece) is still in progress. Keeping this open until the iOS side is complete.

jo-duchan:
Both platforms have landed.

- **Android** — airplane mode, so the device is genuinely offline rather than being told a story (#619).
[...]

PR #651, opened 2026-08-23
Let the provider say what it is actually enforcing

### LINKED ISSUE #641: Provider.handleNewFlow pays a full parent walk for every host flow
`Provider.handleNewFlow` resolves every flow's owning simulator by walking the process's parent chain to the top with one `sysctl(KERN_PROC_PID)` per level, and only *positive* results are cached — a flow that turns out to belong to the Mac itself pays the whole walk, every time, and caches nothing.

Host flows are the overwhelming majority. The user's browser, mail, every background daemon on the machine goes through this before being allowed.

Not obviously fixable by caching the negative: the miss is rejected by a `launchd_sim` path check, and a pid that is *not* a simulator process today can be reused by one tomorrow, so a negative cache needs the same `(pid, start-time)` identity the positive one now uses (added in #607) plus a bound on its size.

Worth measuring before choosing: the walk is 2–4 `sysctl` calls on a host flow, which may be small enough that this is not worth the cache at all. No measurement exists — that is the first step, not the fix.

Touching: `packages/ios-agent/ios-netfilter/Extension/Provider.swift`.

### LINKED ISSUE #642: Decide whether a flow whose simulator cannot be resolved should fail open or closed
`Provider.handleNewFlow` allows any flow whose owning simulator it could not resolve, and logs it identically to a genuine host flow (`udid=-`). Two different things are being reported as one:

- a flow that really belongs to the Mac
- a flow whose lookup **failed** — no audit token, an unreadable `KERN_PROCARGS2`, a process that exited mid-walk

Today both are allowed. That is the right default for the first and a silent hole for the second: a simulator that is supposed to be offline reaching the network because a `sysctl` returned an error, with the log saying the flow was a host flow.

The decision is which way an unresolvable flow should fail, and it is genuinely two-sided. Failing closed on a lookup error means a transient `sysctl` failure can cut the user's own browser, which is worse than the hole. Failing open, as now, means the feature has a silent bypass. A third option is to keep failing open but **distinguish the two in the log**, so the hole is at least observable — which is cheap and may be the whole answer.

Touching: `packages/ios-agent/ios-netfilter/Extension/Provider.swift`.

jo-duchan:
## Summary

Three follow-ups from #607 that turned out to be one thing seen from three sides: **nobody could see
what the running filter was doing.** One file answers all three, which is cheaper than three
mechanisms and is why they land together.

`/Library/Application Support/tapflow/tapflow-netfilter-state.json` — root-owned, world-readable,
refreshed on a 5s pulse and immediately when the rule changes:

- **#639** — `rule` is what the *running* provider holds, not what was saved. The agent decided
  "steerable" entirely from the injected dylib's verdict, which is evidence about layer 2, so a
  filter that was killed or never approved left the control claiming a kernel was dropping traffic
  it was not.
- **#641** — the attribution walk costs **~190–290µs**, measured over hundreds of flows. That is the
  number the issue asked for before deciding about a cache, and it argues against one.
- **#642** — an unresolved flow is no longer indistinguishable from a host flow. It is still
  allowed, deliberately; what was wrong is that the hole was invisible.

Closes #641, closes #642. Refs #639 — the agent-side consumption needs a `NetworkUnavailableReason`
member that does not exist yet, and lands with it.

## What the review and the verification changed

The second commit is most of the work, and two of its findings are worth calling out:

**The heartbeat had no heart.** It wrote only from `handleNewFlow`, so a dead provider left its last
rule on disk forever — the exact lie #639 exists to catch, written to a file. Now it pulses, and
`SIGKILL` was timed to set the reader's threshold: the file freezes and launchd restarts the
provider about **7 seconds** later.

**The agent was asking macOS to install or replace a system extension on every network toggle**,
because activation and configuration were the same code path. Separated into `--install` (once per
release), the default (rule write), and `--off`.

Verification also found that an `OSSystemExtensionRequest` can go **entirely unanswered** — no
error, no refusal, no delegate callback. The approval deadline added in #645 did not cover it, since
that one arms inside `requestNeedsUserApproval`. There is now an overall 45s deadline, exit 6,
verified by firing reproducibly.

## Checklist

- [x] Tests written and passing — 442 in `ios-agent`; the changed code is a system extension, verified against two real simulators
- [x] No `any`
- [x] Interface changes land in `agent-core` first — n/a, none changed

## Related `.work/` docs

`.work/reviews/fix__639-641-642-provider-visibility.md`. Four findings, all real, plus two the
verification turned up on its own. It also records a correction worth reading: **I asserted a cause
for the stalled activation in three places and my own next measurement disproved it** — a reboot
cleared the extension backlog and the very next replacement stalled identically. All three now say
the cause is unknown and that the deadline exists because the failure is silent.

The one thing not verified: the full three-layer end-to-end. The reboot cleared `/tmp`, taking the
probe app with it. This branch touches only the provider and the host binary, so layer-1 isolation
and the heartbeat — what it actually changes — are covered by the measurements in the record.

## Summary by CodeRabbit

* **New Features**
  * Added separate controls for installing, enabling, disabling, and updating iOS network-filter rules.
  * Added live filter status reporting with active rules, flow counts, attribution results, and timing metrics.
  * Improved handling and reporting of host, simulator, and unresolved traffic attribution.

* **Bug Fixes**
  * Added a 45-second timeout for stalled filter activation requests.
  * Improved detection of inactive or stalled filters and reduced unnecessary extension reinstalls.

* **Documentation**
  * Updated usage, troubleshooting, and build guidance for the iOS network filter.


jo-duchan:
@coderabbitai — the three outside-diff findings on `Provider.swift` had no thread to answer in, so they are answered here. All three taken, in `7e07cbb`. Nothing declined.

They turned out to be one failure with three entrances, which is why they are in one commit: **every one of them ends with the state file telling its only reader something untrue about what the filter is enforcing.**

**`probed` cached the failure (125-146).** Set before either candidate was tried, so one transient refusal silenced the file for the provider's whole life. The direction that lands in is the bad one: the agent reads permanent silence as "not enforcing" and reports the device as online while the filter goes on dropping its traffic. Only a resolved path is remembered now. The re-probe costs four syscalls, on the `io` queue, at most once a pulse, and only while there is no path — the repeated log lines were the real cost, so they got the guard instead of the work.

**Render order and enqueue order could disagree (149-178).** Taken as proposed: the enqueue moved inside the lock. Worth stating why that is safe rather than just that it is — `io.async` copies a block and returns, and nothing in this class ever waits on `io` (`remove` is `async` too), so holding the lock across it cannot deadlock. The window was small; the consequence was not, since an older `rule` landing last is the same lie as the one above, told the other way round.

**The rule was hand-quoted (195-204).** It arrives through `vendorConfiguration`, which this provider does not write and cannot constrain, so one quote or backslash made the whole file unparseable — which reads as "not enforcing" too, and with nothing in the log to say why. `JSONSerialization` quotes it now. Checked against the documented sample before committing: ordinary output is byte-identical, so the README needed no change, and the hostile cases parse.

```
PARSES   {"at":…,"pulseSeconds":5,"rule":[],…}
PARSES   {"at":…,"pulseSeconds":5,"rule":["A-1","B-2"],…}
PARSES   {"at":…,"pulseSeconds":5,"rule":["say \"hi\""],…}
PARSES   {"at":…,"pulseSeconds":5,"rule":["back\\slash"],…}
PARSES   {"at":…,"pulseSeconds":5,"rule":["nl\nhere"],…}
```

The extension builds clean with no warnings. `CHANGELOG.md` also picked up a conflict with `main` after #649 merged; both entries kept.

PR #656, opened 2026-08-24
fix(relay): cancel an overdue trailing request instead of firing both edges

### LINKED ISSUE #621: Make the network-request-state coalescing tests deterministic instead of wall-clock
Found by the pre-PR review of #614's wire slice. Split out because a deterministic fix needs a design decision, not a small edit.

Three tests in `packages/relay/src/__tests__/sessionOwnershipSeam.test.ts` — the two coalescing ones and the disposal one — use `vi.useFakeTimers({ toFake: ['Date'] })`, freezing `Date` while leaving `setTimeout` real. That combination is deliberate: faking `setTimeout` would also freeze the WebSocket library's timers and the relay's own heartbeat, which the existing IDR test says in as many words.

The cost is that their intermediate "still one request" assertions depend on 3–4 real WebSocket round trips finishing inside a real ~400 ms window, against a relay that is genuinely listening with sqlite behind it. On a loaded CI runner that window can be missed, and the assertion reads 2.

The two ways out are a design choice:

- Fake `setTimeout` and deal with the ws and heartbeat timers it also captures — possibly by faking only within a scoped window, or by injecting a clock into `networkStateRequester`.
- Expose a test seam on the requester so the trailing edge can be driven directly, at the cost of the production path and the tested path diverging slightly.

Not urgent: no flake has been observed, and the assertions are correct when they run. This is about whether they stay correct on someone else's machine.

ML642:
## Summary

Replaces the network-state requester tests’ real ~400 ms timing window with deterministic scheduler coverage.

The relay still uses real `Date` and `setTimeout` in production. The extracted internal requester accepts a scheduler only at its timing boundary, so unit tests can drive leading-edge, trailing-edge, exact-boundary, disposal, and overdue-timer behaviour without faking WebSocket or heartbeat timers.

Also fixes an overdue pending timer edge case that could otherwise send a duplicate `network:request-state` after a new window began. The lifecycle integration tests continue to cover requester reuse, disposal on session removal, and retention across a dirty browser disconnect.

Closes #621
## Testing

- `corepack pnpm --filter @tapflowio/relay test src/__tests__/trailingRequester.test.ts src/__tests__/sessionOwnershipSeam.test.ts` — 45 passed
- Relay `typecheck` and `lint` passed.
- `pnpm changeset:check` passed.
- Mutation checks confirmed the new tests fail when coalescing, disposal cancellation, exact-boundary handling, or overdue-timer cancellation is removed.
- Full relay suite on Windows: 699 passed; 5 unrelated platform-specific failures require POSIX file modes or the `zip` executable.

## Checklist

- [x] No `any`

## Related `.work/` docs
- `.work/archive/2026-08-23-deterministic-network-requester-621-plan.md`
- `.work/reviews/test__deterministic-network-requester-621.md`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved relay network-state request handling by coalescing bursts and preventing duplicate delayed requests.
  * Ensured pending requests are canceled when sessions are removed or the relay shuts down.
  * Preserved network-state request handling across temporary browser disconnects and delivered responses to replacement connections.

* **Tests**
  * Added deterministic coverage for request coalescing, timer boundaries, cleanup, session rejoining, and disconnect scenarios.


jo-duchan:
Thanks for this one — the shape is right, and I want to say why before the two small things, because neither of them is about the design.

Injecting the scheduler at the timing boundary, rather than faking timers wholesale, is the choice that makes this work. `toFake: ['Date']` was always a compromise: it left the trailing edge as a real ~400 ms wait, and `vi.waitFor` resolving on the first poll that sees the count is a real flake surface — the old block had to add a 150 ms settle to paper over exactly that. Putting the policy behind `TimerScheduler` gets the whole window under a clock you control, and the four helper cases — coalescing, exact boundary, disposal, overdue — read as a specification instead of as timing trivia.

I'm also glad you left `idrRequester` alone. It drops where this one coalesces, and the comment beside `NETWORK_STATE_REQUEST_THROTTLE_MS` says why the two constants aren't shared; folding both onto one helper would have made tuning one tune the other.

And the overdue branch is a real fix, not a defensive one. Node runs due timers in the timers phase, *before* poll — so a trailing timer that comes due **during** the poll phase is still armed when a WebSocket message callback in that same phase calls the requester, `wait` is already `<= 0`, the leading edge fires, and the timer fires again on the next iteration. Two sends for one window. Narrow, but nothing about it is theoretical.

## Two things before it goes in

### The title

`test(relay):` doesn't cover what's in the diff. Cancelling the overdue timer is a production behaviour change in `packages/relay/src/lib/trailingRequester.ts`, and we squash-merge, so the title is the line someone reads in `git log` a year from now. Per CONTRIBUTING the type should follow the riskiest part of the change, not the largest:

```
fix(relay): cancel an overdue trailing request instead of firing both edges
```

You did document it — the changeset says it plainly, so nothing was hidden here and this is only about the commit line. And retitling doesn't drag a root `CHANGELOG.md` entry along with it: a duplicate `network:request-state` between the relay and the agent still isn't something a user can observe, so your `<!-- changelog: internal -->` marker stays correct as written.

### One assertion in `keeps one requester across repeated re-joins`

The comment above it says the seam "proves each re-join **reaches** the same per-session requester". The identity half is asserted; the reaching half isn't. I measured it rather than guessing — this at `RelayServer.ts:1665` is a later re-join that no longer reaches the requester:

```ts
if (session.agentCapabilities?.includes('network-control')) {
  const fresh = !this.networkStateRequesters.has(session.id)
  const r = this.networkStateRequester(session.id)
  if (fresh) r()
}
```

and on your head it reads:

```
keeps one requester across repeated re-joins            PASS
serves a viewer that comes back … dirty disconnect      FAIL
```

So the suite does catch it — but only in the test about the *dirty-disconnect* path, and only because that one swaps a `vi.fn()` into the map. The test actually named for repeated re-joins walks straight through it. That's the kind of gap that survives a refactor: someone tightens the re-join path later, sees the seam test named for it go green, and the thing that objects is a test about laptop lids.

A delegating wrapper closes it, and keeps the real requester running rather than standing in for it:

```ts
const reached = vi.fn()
const requesters = (server as unknown as {
  networkStateRequesters: Map<string, NetworkRequester>
}).networkStateRequesters
requesters.set(sessionId, Object.assign(
  () => { reached(); requester!() },
  { dispose: () => requester!.dispose() },
))

await rejoin(sessionId, browser)
await rejoin(sessionId, browser)
expect(reached).toHaveBeenCalledTimes(2)
```

Measured both ways: it fails under the mutation above, and 45/45 pass on your head unmutated. Take the shape or leave it — the assertion is what I'm asking for.

## What I checked and cleared

So you know where the rest landed, and so nobody re-does it:

- **The leading edge is still proven on the wire.** `asks the agent what the device network is doing when a viewer re-joins` (`:380`) still counts a real `network:request-state` on the agent socket, so replacing the `fire` closure's `sendTo` still dies. The trailing edge now runs the same closure, so moving its coverage into the helper costs less than it looks like.
- **`disposes a requester when the session is forgotten`** splitting into "the relay calls `dispose`" plus "`dispose` cancels the timer" is the right seam, and both halves are asserted.
- **`elapseWithoutRunning`** is a good primitive. It's the only way to reach the overdue case at all, and it's named for what it does.

Nothing else from me. Retitle, add the assertion, and this is good to go — thanks for the care that went into the mutation notes in the description, they made this much faster to check.

ML642:
Thanks for the detailed review. I’ve updated the title to:

fix(relay): cancel an overdue trailing request instead of firing both edges

I also added the assertion to keeps one requester across repeated re-joins so it now verifies that each re-join actually reaches the same requester, while keeping the real requester behavior intact.
All tests are passing on my side.

jo-duchan:
Both landed, and thanks for turning it around so quickly.

I re-ran it here before merging: 45/45 on the two files, 704 on the full relay suite, typecheck clean. And with the `fresh` guard mutation applied, `keeps one requester across repeated re-joins` now fails on its own (`expected 2 times, got 0`) instead of leaving the dirty-disconnect test to do the objecting. That was the whole point of the ask, so this is exactly it.

Nice touch keeping the identity assertion and pointing it at the wrapper. That pins that the relay does not swap the map entry either, which the snippet I pasted had quietly dropped.

Merging it. Thanks for the care on this one, the mutation notes in the description included — they made both rounds much faster to check. Hope you pick up another.

PR #678, opened 2026-08-25
fix(ios-agent): write the verdict atomically, and stop the injected library failing silently

### LINKED ISSUE #653: The injected library writes its verdict non-atomically, so a concurrent read reports the hooks as failed
Parent: #607

Found by the design review of #618 · #638 · #639 · #644, deferred because the fix is in the injected library's write path and does not belong in that protocol change.

`SimulatorNetwork.readVerdict` treats an unparseable verdict file as `installed: false` (`packages/ios-agent/src/SimulatorNetwork.ts:312-314`), which `state()` turns into `hooks-not-installed`. Today every unsteerable reason renders the same sentence, so this is invisible. The reason set work about to land gives `hooks-not-installed` its own permanent, red rendering — "tapflow couldn't reach inside this app", a reason whose only remedy is to give up.

The dylib writes that file with `fopen` / `fputs` / `fclose` (`packages/ios-agent/src/network-hook.m`, `tf_write_verdict`), so a reader can see a truncated file. The agent reads it on every `state()` call, which the relay triggers on `device:ready`, on a viewer's re-join, and after every toggle — so a read concurrent with the write is reachable on a completely healthy session.

**A parse failure is not evidence that the hooks failed.** It is the codebase's own rule — *require the specific failure, not any failure* — pointed at the one place that currently breaks it.

Write to a temp path in the same directory and `rename()` onto the target, so a reader sees either the whole old file or the whole new one. `rename` within a filesystem is atomic; the condition file next to it is empty and needs no such care.

Touching: `packages/ios-agent/src/network-hook.m`.

jo-duchan:
## Summary

The injected library wrote its verdict with `fopen(path, "w")`, which truncates the file the agent
may be reading at that instant. The agent reads it on every `state()` — `device:ready`, a viewer's
re-join, after every toggle — so a read landing inside the write was reachable on a session where
nothing was wrong, and what it got was half a file. It now writes beside the target and renames onto
it.

Two things this could not be landed without, both about the same committed binary:

- **`bin/libtapflow-nethook.dylib` had no build recipe.** The flags lived in whichever shell last
  produced it. They were recovered from the binary rather than remembered — `otool -l` for the
  platform and deployment target, `size -m` to confirm — and a rebuild matched the committed one
  section for section, `__text` included at 9040. `build-nethook.sh` holds them now.
- **Nothing hashed it**, while every test that exercises the hook injects a fake dylib path, so
  editing the source and shipping the old binary was silent. It is recorded against its sources now,
  the way the network filter is, sharing that guard's hashing rather than copying it. The difference
  is stated where it matters: a failure here is the contributor's to fix, because no signing key is
  involved.

The library also had no diagnosis at all. dyld ignores a `DYLD_INSERT_LIBRARIES` path that does not
exist without a word, so a damaged install launched the app unhooked and wrote no verdict — leaving
the control asking the tester to launch an app through tapflow, for the whole session, while the app
they launched was running in front of them. `tapflow doctor ios` reports it now, and `state()` says
`hooks-not-installed`.

Three fixtures named `/fake/libtapflow-nethook.dylib`. Adding the check broke 17 tests, which is what
it looks like when nothing ever asked whether the thing being injected was there.

Closes #653.

## Checklist

- [x] No `any`

`@tapflowio/protocol` is in the changeset: `hooks-not-installed`'s one-line contract said the agent
"proved this by trying", which a `stat` does not.

## Related `.work/` docs

`.work/reviews/fix__nethook-atomic-verdict.md`. Two independent channels, ten findings, all ten
fixed. The one worth reading is that a guard **this branch added** claimed to be structural and was
not: it proved the shipped dylib renames by searching the whole file for the bytes `_rename`, and a
reviewer disproved that in one compile — a dylib with zero rename imports, a `const char *note =
"atomic_rename"` and a plain `fopen(path, "w")` carries those bytes in `__cstring`. The whole
sequence went green with the pre-#653 binary shipped. It now reads `LC_SYMTAB`.

The record also states what this PR did **not** get: the design-level review AGENTS.md requires for a
change spanning packages. It grew from a one-package fix into `ios-agent` + `cli` + `protocol` during
implementation, and two of the ten findings are what that pass would have caught before the code.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved iOS diagnostics to detect missing or unreadable network hooks and provide repair guidance.
  - Prevented transient false failures by writing iOS network status updates atomically.
  - Clarified network status reporting when hooks are unavailable or app launch is still pending.

- **Documentation**
  - Updated CLI documentation with network-hook checks, installation issues, and recovery steps.

- **Quality Improvements**
  - Added verification that the shipped network-hook binary matches its recorded source and build metadata.


jo-duchan:
@coderabbitai review

Issue #634, opened 2026-08-23
The device toolbar's dividers group by an unstated criterion


jo-duchan:
Parent: #607

The device toolbar draws three groups separated by two dividers, and nothing records what the groups
are. Reading the buttons, the criterion does not survive the middle group.

```
[launch] [platform buttons] [deeplink]
─────────
[screenshot] [record]
─────────
[rotate] [network]
```

**There are three groups but only two kinds of thing.** The middle one is coherent — it takes
something out of the session. The first and third are both "act on the device", split by something
that is never stated, and because the capture group sits *between* them they read as one category
arbitrarily cut in half.

A criterion is almost there, and it is worth naming before deciding anything: the first group is
**momentary** (press and it is over) and the third is **sticky** (the device stays that way until
someone changes it back). If that is the intended split then the order is wrong — capture belongs at
the end, so the toolbar reads *act → what condition you leave it in → what you take away* — and the
keyboard toggle is in the wrong group, since a software keyboard left up stays up.

Grouping the sticky ones has support in the toolbar's own reasoning. `networkLook` gives the offline
position the only colour in the toolbar, and says why: *"It is a state a tester deliberately put the
device into and will forget about, and forgetting is what makes the next hour of testing
confusing."* Rotate and the keyboard are forgotten the same way.

Three ways to settle it, and the choice is a product one rather than a cleanup:

| | What it is | Cost |
|---|---|---|
| **momentary / sticky / capture** | Capture moves to the end, keyboard moves down | The keyboard button moves, and people have muscle memory for it |
| **act / capture** (two groups) | One divider goes away | Nothing groups the states that are easy to forget |
| **by frequency** | Screenshot rises toward the top | Abandons the kind axis; the frequencies differ per platform |

Whatever is picked, **write the criterion down next to the dividers** — the eight commits that
touched `SimulatorToolbar.tsx` are all about the network control's accessibility, and none of them
says why the lines are where they are. That is why this is being asked again.

`packages/dashboard/components/device/shared/SimulatorToolbar.tsx`

jo-duchan:
Shipped in #682, and the rule it asked for is written down rather than implied.

The toolbar now has four groups, ordered by what the tester is doing to the device rather than by how
the feature is built: **Navigation → Device → Capture → Environment**, and inside each group the
controls reached for most come first. They are `role="group"` with names, so the structure exists for a
screen reader too — before this they were `<div>`s with a background colour, and to assistive tech the
[...]

Issue #620, opened 2026-08-21
NetworkControlCapability writes change device state without reporting it on the wire


jo-duchan:
Parent: #607

Found by the design review of #607's viewer-path slice (step 4a). Deferred because the fix presumes an interface decision that is #617's.

`AndroidAgent.setNetworkOffline` returns the resulting `NetworkStatePayload` to its caller and sends nothing on the WS, while `network:set` — the viewer path into the same device — always answers with a `network:state`. The two paths change the same state and only one reports it.

Today nothing breaks, because no consumer exists (#617 confirms that). The moment one does — `mcp-server` exposing the tool — an agent-driven flip leaves every viewer on that session rendering the previous position, with no producer to correct it until the browser rejoins. Step 4a does not help here: it re-asks the agent *on join*, and no join happens.

That is the exact failure the unsolicited-report design exists to prevent, reached through the one write path that has no report.

Decide alongside #617: the fix is to report on the session the write landed on, which presumes the interface knows which session that is — and today `setNetworkOffline` picks `deviceStates.keys().next().value`, an arbitrary first session.

`AudioStreamCapability` is worth looking at in the same pass, since it is the other capability interface and may have the same shape.

jo-duchan:
Closing: the observation is real, the harm it predicts cannot occur.

The asymmetry described here is exactly right — `setNetworkOffline` returns its payload to the caller
and sends nothing, while `network:set` into the same device always answers with a `network:state`. Two
paths change the same state and only one reports it.

What does not hold is the trigger. This issue says the moment a consumer exists — "`mcp-server`
exposing the tool" — an agent-driven flip leaves every viewer rendering the previous position. **MCP
does not call this method and will not.** `mcp-server` and `flow-runner` each hold a WebSocket client
to the relay and address devices by session over the wire, so the network tool they expose goes through
`network:set`, which already names its session and already answers with a correlated `network:state`.
It never reaches `setNetworkOffline`. That is written up in `NetworkControlCapability.ts`'s doc block,
and was verified again while working this: the two methods have no callers anywhere, and agents are
constructed only inside their own process.

A design review then refuted the obvious fallback too — "report to every session holding the device".
There is no such thing: the relay builds `registeredSessions` from a `deviceId → sessionId` map
(`RelayServer.ts`), and a session holds one `browserSocket`. Two sessions on one device, and two
browsers on one session, are both unproducible. `reportEnforcementLost`'s loop over sessions is
[...]

Issue #617, opened 2026-08-20
NetworkControlCapability's session pick is arbitrary, and the interface cannot express which one


jo-duchan:
Parent: #607

Found by adversarial review of the Android network work (#607). Deferred because the fix is an interface decision, not a line.

`AndroidAgent.setNetworkOffline` and `networkState` both resolve their device with `this.deviceStates.keys().next().value` — insertion order, not a caller-chosen session. On an agent holding several sessions, `setNetworkOffline` is a **write** that lands on whichever session registered first, potentially another tester's device, and the caller has no way to say otherwise because `NetworkControlCapability` takes no session id.

`get sessionId()` on the same class has the same shape and has been fine, but reading and writing are not the same risk: the worst case there is answering about the wrong device, and here it is taking someone else's device off the network while they are using it.

No consumer exists yet — `hasNetworkControl` is called nowhere outside `agent-core` and its own tests — so nothing is broken today. That is also why now is the moment: the first consumer will be written against whatever this interface says.

The decision is whether `NetworkControlCapability` should be session-scoped, which is not just this method:

- `AudioStreamCapability` has the same shape and the same question.
- A session id in the signature makes every implementer carry it, including third-party platforms registered through `AgentRegistry.register()`.
- The alternative is that these methods are documented as "the agent's current session" and the WS path (`network:set`, which does carry a session) stays the only session-accurate route — in which case the doc has to say so where an implementer will read it.

jo-duchan:
#645 (the iOS half of #607) changed half of this, and it is worth narrowing rather than leaving the issue describing a state that no longer exists.

**iOS is resolved.** `IOSAgent.setNetworkOffline` and `networkState` no longer take `deviceStates.keys().next()`. They go through `soleLiveDeviceId`, which applies the refusal this class had already settled on and written down for its other entry points — one booted device or an error, never a silent pick — and confirms the boot state with simctl rather than trusting a flag that reconnects clear. So the write cannot land on another tester's device there.

**Android still has exactly the shape described above.** `AndroidAgent.setNetworkOffline` and `networkState` both still read `this.deviceStates.keys().next().value`, and `serialFor` returns `undefined` for a session whose device is not booted — so on a host with several sessions it either writes to whichever registered first or refuses while a booted device sits beside it.

[...]

PR #682, opened 2026-08-26
fix(dashboard): forget the network position when the device does, and group the toolbar by intent

### LINKED ISSUE #625: useNetworkControl keeps the pre-reboot position across a device:boot on the same session
Parent: #607

Found by the pre-PR review of #607's dashboard slice. Split out because what a reboot should reset is a design call, not a one-liner.

`useNetworkControl` resets its position only when `sessionId` changes. `DeviceViewer` drops `deviceReady` to false on `device:booting` and on agent-away, and back to true on `device:ready`, but the position survives all of it — and the report deadline does not re-arm, because it is gated on `position === 'waiting'`.

So during an Android emulator reboot the toolbar keeps showing the position from before it, for the 30–60s the boot takes. That is worse than merely stale: the agent's boot path clears airplane mode and reports online, so an amber "Device is offline" sits over a device being reset to the opposite. And unlike the first boot, nothing catches the loss of that report — the same silence that yields `unknown` before boot #1 leaves a stale position after boot #2, permanently.

What needs deciding:

- Should `deviceReady` going false return the position to `waiting`? That re-arms the report deadline, which is the mechanism that already handles "no report arrived".
- What happens to an in-flight `network:set` across a reboot — the request deadline added in the same slice will clear `pending`, but should the reboot clear it sooner and say why?
- `device:booting` and agent-away both drop `deviceReady`, and they are not the same event: one is a device replacing itself, the other is the same device becoming unreachable. They may want different answers.

jo-duchan:
## Summary

Three things the network control and the toolbar were getting wrong on screen.

**A device that restarts keeps its session**, and the control only forgot what it knew when the
*session* changed — so for the 30–60 seconds an emulator takes to come back it showed the position
from before it. The agent's boot path turns airplane mode off and reports the device online, so an
amber "offline" sat over a device being reset to the opposite, and the report deadline never re-armed
because it is gated on `position === 'waiting'`. Nothing replaced it. It forgets the moment the device
stops being ready.

**The toolbar's buttons were grouped by a criterion nobody had written down.** They are grouped now by
what the tester is doing to the device — Navigation → Device → Capture → Environment, frequent first
inside each — and the rule with its worked examples is in `packages/dashboard/AGENTS.md`, where
somebody adding a button reads it before opening the component. GPS goes in Environment, Shake in
Device, and the deeplink is Navigation because from the tester's side it is "go to this screen".

**Where a button sits is decided in one place now.** Android's order came from the *agent*, because
its buttons arrive as a capability list and the dashboard rendered that array in array order — so a
reorder in `android-agent` moved buttons in the browser with nothing on either side to notice. The
toolbar takes `navigationSlot` and `deviceSlot` instead of one `platformSlot`; the agent still owns
which buttons exist.

Also recorded rather than changed: `NetworkControlCapability` is an in-process API. `mcp-server` and
`flow-runner` hold a relay client and address devices by session over the wire, so the network tool
they would expose goes through `network:set`, which already names its session and answers with a
correlated report. Two issues had asked this interface for a session id and a wire report on the
premise that MCP calls it; five comments repeating that premise are corrected.

Closes #625, #634, #617, #620.

## Checklist

- [x] No `any`

## Related `.work/` docs

`.work/reviews/fix__network-viewer-surface.md`.

**The gate before the reviewers found six things, and every one was mine.** Three rounds of the
pre-commit a11y pass: the four groups existed only as `<div>`s with a background colour, invisible to
assistive tech while the test asserted "order in the accessibility tree"; `display: contents` can drop
an element's role with its box, and **jsdom evaluates no CSS**, so that test passed either way; a
named, empty Environment group when the agent has no network control; and a keyboard toggle that
announced "unavailable" with no reason, then with a name that changes under the focus that just
clicked it — which screen readers do not re-announce. The network control ten lines away carries a
live region for exactly that reason and says so; I had not read it.

Then two review channels found the seam between the two new guards: **nothing renders `AndroidViewer`**,
so dropping `deviceSlot={deviceSlot}` or swapping the two lists left dashboard 435 and scripts 445
green while half the Android toolbar vanished. And they found that a claim in the changelog — that the
platforms had drifted and testers found the same button in different places — **did not happen**;
`git show main:` shows the pre-change order was identical on both. Corrected in all four places.

One finding was a phantom, and the record says why: the two channels shared a checkout, one of them
planted a measurement probe, and the other read the tree while it was there.

## Summary by CodeRabbit

* **New Features**
  * Standardized device toolbar layout into Navigation, Device, Capture, and Environment sections across iOS and Android.
  * Improved keyboard-toggle accessibility with status updates, clearer labels, and transition feedback.

* **Bug Fixes**
  * Network controls now reset correctly when devices restart or become unavailable.
  * Keyboard-toggle requests time out gracefully, preventing stuck loading states and duplicate actions.

* **Documentation**
  * Clarified network-control behavior and toolbar organization guidelines.


jo-duchan:
@coderabbitai review

PR #694, opened 2026-08-27
test(dashboard): cover reset mode across agent changes

### LINKED ISSUE #608: test(dashboard): QASession Android reset test's second assertion cannot fail
Split out of #447 item 3 (PR pending). Found by adversarial review; not a hole, but a line that reads as covered and is not.

`packages/dashboard/src/__tests__/QASession.reset.test.tsx:171`

```ts
expect(screen.queryByRole("switch")).not.toBeInTheDocument()
await user.click(screen.getByText("Pixel 7"))
expect(viewerMounts).toEqual([{ deviceId: "dev-a", resetMode: "app-only" }])
```

The switch is not rendered, so `resetMode` is still its initial `app-only`. Deleting the guard in `useDeviceSelector.ts` — `setAppliedResetMode(fullResetSupported ? resetMode : "app-only")` → `setAppliedResetMode(resetMode)` — leaves this assertion green. The first assertion is sound (it fails if the capability check is loosened); the second one exercises nothing.

The hook-level test added alongside it does cover the guard, so nothing is unprotected. What is missing is the integration-level path: arm the toggle on a capable agent, go back to the Mac list without picking a device, select an agent that does not advertise `full-reset`, then pick — which is the same journey the hook test walks with `rerender`.

Background: `contributing/test-and-guard-coverage.md` §2.

ML642:
## Summary

Adds dashboard integration coverage for Full reset across an agent change.

The test arms Full reset on a capable Mac, returns to the Mac list without selecting a device, then selects an agent without `full-reset`. It verifies the viewer receives `resetMode: 'app-only'`, preventing an armed destructive reset from carrying across agents.

closes #608 
## Checklist

- [x] No `any`
- [x] Interface changes land in `agent-core` first — not applicable

## Related `.work/` docs

- `.work/2026-08-27-dashboard-reset-integration-608-plan.md`

## Summary by CodeRabbit

* **Bug Fixes**
  * Selecting a device without full-reset support now automatically limits an armed reset to an app-only reset.
  * Prevents unsupported full resets from being initiated.
  * Improves reset handling across devices with different reset capabilities, helping ensure that only supported reset options are applied.

### DIFF
--- a/packages/dashboard/src/__tests__/QASession.reset.test.tsx
+++ b/packages/dashboard/src/__tests__/QASession.reset.test.tsx
@@ -69,6 +69,15 @@ const ANDROID_AGENTS: SessionInfo[] = [{
   capabilities: ['clipboard'],
   devices: [device('dev-a', 'Pixel 7', 'android')],
 }]
+const MIXED_CAPABILITY_AGENTS: SessionInfo[] = [
+  ...AGENTS,
+  {
+    agentName: 'legacy-mac',
+    platform: 'ios',
+    capabilities: ['clipboard'],
+    devices: [device('dev-c', 'iPhone 14')],
+  },
+]
 
 /** The breadcrumb is rendered by the layout, outside QASession — and once a session is open it is
  *  the only way back to the Mac list, which is the path this issue is about. */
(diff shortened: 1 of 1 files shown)


jo-duchan:
This is the integration path #608 was missing, and it walks exactly the journey the issue described — arm on a capable Mac, back to the list without picking, then an agent that does not advertise `full-reset`.

I checked it the way the issue framed it. With the guard in `useDeviceSelector.ts:35` reduced to `setAppliedResetMode(resetMode)`, your test is the only one that fails (1 failed, 6 passed), and the old Android assertion stays green under that same mutation — which is the thing the issue was pointing at. I also ran your file against current main: 7/7, and none of the three files it leans on have moved since your base.

One small thing before it goes in. The line #608 is named after is still there, untouched:

```ts
await user.click(screen.getByText('Pixel 7'))
expect(viewerMounts).toEqual([{ deviceId: 'dev-a', resetMode: 'app-only' }])
```

It is not worthless — it pins that the Android path mounts with the default — but it still reads as proof of the capability guard, and once this PR closes #608 there is no note anywhere saying otherwise. A one-line comment above it, saying what it does and does not prove and that the guard is covered by your new test, is all I am after. No need to delete it.

Thanks for this one.

ML642:
I added a comment above the existing Android assertion clarifying that it only pins Android’s default app-only mount. The armed capability guard is covered by the new mixed-agent integration case below it.
Verified: QASession.reset.test.tsx passes 7/7.

jo-duchan:
The comment is what I asked for, and its placement is right — directly above the two lines it describes.

I re-ran it against current `main` rather than your base, since the branch is 152 commits behind now: 7/7, and the mutation still isolates the same way.

Good to merge. Thanks for seeing this one through.

PR #695, opened 2026-08-27
test(scripts): guard dashboard capability consumers

### LINKED ISSUE #609: Guard the agent capability strings against producer/consumer drift
Split out of #447 item 3 (PR pending). Found by adversarial review. Pre-dates that change — it adds the second instance rather than creating the exposure.

`AgentCapability` (`packages/agent-core/src/types.ts`) is the producer side, but both consumers hard-code the string against a `string[]` that closes nothing:

- `packages/dashboard/hooks/useDeviceSelector.ts` — `capabilities.includes("full-reset")`
- `packages/dashboard/components/DeviceViewer.tsx:503` — the same for `"clipboard"`

Rename a union member and the agents are forced to update by `tsc`, because they type their `AGENT_CAPABILITIES` constant on it. The dashboard compiles unchanged and gates on a value no agent sends, so the control disappears for good — and the dashboard tests hard-code the same literals, so the whole suite stays green. `full-reset` has a live temptation behind it: the wire spells the adjacent concept `resetMode: "full-erase"`, one letter-group away.

The agent-core comment already notes the trade-off ("The dashboard cannot import from here"), but noting a gap is not guarding it.

Shape to copy: `scripts/__tests__/protocolMessageNames.test.mjs`, which does exactly this for message names. Read the `AgentCapability` union from agent-core source text and assert every capability literal the dashboard tests membership on is a member of it.

Read `contributing/test-and-guard-coverage.md` before writing it — a static check whose header cites a lesson usually reads as having applied it.

ML642:
## Summary

Adds a source-based guard for dashboard capability checks.

The guard reads `AgentCapability` from agent-core and verifies that every dashboard `capabilities.includes(...)` / `agentCapabilities.includes(...)` literal is declared by that producer contract. This catches a renamed capability such as `full-reset` drifting to `full-erase`.

closes #609 
## Checklist

- [x] No `any`
- [x] Interface changes land in `agent-core` first — not applicable

## Related `.work/` docs

- `.work/2026-08-27-agent-capability-consumers-609-plan.md`

## Summary by CodeRabbit

* **Tests**
  * Enhanced automated validation for dashboard capability checks using both single- and double-quoted values.
  * Added coverage to detect undeclared capabilities consistently across supported formats.
  * Continued verifying that dashboard consumers reference only declared agent capabilities and that the expected number of consumers remains unchanged.

### DIFF
--- a/scripts/__tests__/agentCapabilityConsumers.test.mjs
+++ b/scripts/__tests__/agentCapabilityConsumers.test.mjs
@@ -14,7 +14,7 @@ function withoutComments(text) {
 }
 
 function stringLiterals(text) {
-  return [...text.matchAll(/'([^']+)'/g)].map((match) => match[1])
+  return [...text.matchAll(/(['"])([^\r\n]*?)\1/g)].map((match) => match[2])
 }
 
 function agentCapabilities(text) {
@@ -23,14 +23,18 @@ function agentCapabilities(text) {
   return new Set(stringLiterals(match[1]))
 }
 
+const capabilityCheck = /\b(?:agentCapabilities|(?:\w+\??\.)?capabilities)\??\.includes\(\s*(['"])([^\r\n]*?)\1\s*\)/g
+
+function capabilityChecks(text) {
+  return [...withoutComments(text).matchAll(capabilityCheck)].map((match) => match[2])
+}
+
 function dashboardCapabilityConsumers() {
   const consumers = []
-  const capabilityCheck = /\b(?:agentCapabilities|(?:\w+\??\.)?capabilities)\??\.includes\(\s*'([^']+)'\s*\)/g
 
   for (const file of sources('packages/dashboard')) {
-    const source = withoutComments(readFileSync(join(root, file), 'utf8'))
-    for (const match of source.matchAll(capabilityCheck)) {
-      consumers.push({ file, capability: match[1] })
+    for (const capability of capabilityChecks(readFileSync(join(root, file), 'utf8'))) {
+      consumers.push({ file, capability })
     }
   }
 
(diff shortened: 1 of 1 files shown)


ML642:
CI failure appears unrelated to this change: SimulatorNetwork.test.ts failed during temporary-directory cleanup with ENOTEMPTY at rmSync. The dashboard capability-consumer guard tests passed. This looks like an iOS test cleanup flake; rerunning the failed job would be appropriate.

jo-duchan:
This is the shape #609 asked for, and the anti-vacuity floor is the part I was most glad to see: `toHaveLength(3)` set from the measured count rather than a round number, which is exactly what `contributing/test-and-guard-coverage.md` §3 asks for.

Verified here. Renaming the union member in agent-core to `full-erase` makes the guard fail and name `useDeviceSelector.ts` as the undeclared consumer, which is the drift the issue exists for. Running the file against current main passes too — main still has the same three call sites, the same union, and the same `include` in `scripts/vitest.config.mjs`.

One thing to write down before it goes in. The matcher sees `*.capabilities.includes(` and `agentCapabilities.includes(`, so a call that reaches the array through an alias is invisible to it. I measured it: planting this in `useDeviceSelector.ts` leaves the guard green, and the cardinality assertion does not object either, because the three it counts are all still there.

```ts
const caps = selectedSession?.capabilities ?? []
const bogusSupported = caps.includes('totally-bogus')
```

That is not a flaw in the implementation — it is what a source-text guard is. But §3's prescription for this exact case is to say so in the file: a spelling assertion is a floor, not a fence. A line naming what the matcher cannot see (an alias, a constant, a template literal) is all this needs.

And the red `test (22)` is not yours. It is `SimulatorNetwork.test.ts > arm > still recovers a leftover when the device that owns it comes back` in ios-agent, which this PR does not touch — `test (24)` passed, #694 passed on the same base half an hour earlier, and that file passes 79/79 on current main here. I will re-run the job.

ML642:
I documented the matcher boundary beside the regex: it only catches direct literal checks, while aliases, constants, and template literals remain outside the scope of this source-text guard.
I also corrected the double-quoted mutation fixture to use an actually undeclared capability. The targeted guard test passes 2/2.

jo-duchan:
The boundary comment is what I asked for, and I checked that it is honest rather than decorative: planting the alias I described leaves the scripts suite at 470/470, which is exactly what that line now says it cannot see.

I re-ran the rest against current `main` too, since the branch is 152 commits behind: 39 files / 470 tests pass, and the rename still fails the guard and names `useDeviceSelector.ts`.

One thing I found while checking this is filed as #700 rather than piled onto your PR — the relay has a capability gate of its own that this scan does not reach, and closing it is a typing decision rather than a wider scan. Nothing for you to do here.

Good to merge. Thanks for this one.

Issue #661, opened 2026-08-24
test(relay): make the full relay test suite Windows-compatible


ML642:
### What happened?

On Windows, the full relay suite is not portable:

powershell
corepack pnpm --filter @tapflowio/relay test

### Steps to reproduce

1. Use Windows with Node.js and Corepack available.
2. Clone the repository and install dependencies.
3. From the repository root, run:

   corepack pnpm --filter @tapflowio/relay test

### Expected behavior

The relay test suite should complete successfully on Windows, or explicitly skip only assertions that cannot be meaningful on Windows.

ZIP fixtures should not require a system `zip` executable, and temporary test resources should close cleanly before cleanup.

### Platform

Neither (relay / dashboard / CLI)

### tapflow version

0.1.0 (source checkout; @tapflowio/relay 0.19.0)

### Node.js version

v24.13.0

### Mac architecture (the machine running the agent)

arm64 (Apple Silicon)

### Relevant logs

```shell
Test Files  4 failed | 57 passed | 1 skipped (62)
Tests       5 failed | 699 passed | 1 skipped (705)

accountKey.test.ts / diskCertStore.test.ts:
expected file mode 0600, received 0666 on Windows.

builds.test.ts / upload-limits.test.ts:
ZIP fixtures were not created because the system `zip` executable is unavailable.

builds.test.ts cleanup:
EPERM while removing temporary test directories.
```

PR #698, opened 2026-08-28
fix(cli): use resolved TLS host in agent instructions

### LINKED ISSUE #627: Agent-connect line still says <host> while the banner names the certificate's host
## Problem

Once #613 lands, the start banner names the certificate's host, but the agent-connect line directly underneath it still carries a placeholder:

```
Relay  : https://relay.example.com:4000
Connect Mac agents:  tapflow agent start --relay wss://<host>:4000 --token <agent-PAT>
```

`packages/cli/src/commands/start.ts` has the same shape with `<this-ip>`. A teammate who substitutes an IP address there gets a certificate name mismatch — the exact failure #293 set out to remove from the line above it.

## Why this needs a decision rather than a one-line substitution

`displayHost` is `localhost` whenever TLS is off, and that is the majority path. Interpolating it unconditionally would print:

```
tapflow agent start --relay ws://localhost:4000 --token <agent-PAT>
```

which is wrong for the case the placeholder exists to serve: an agent running on a *different* Mac, where `localhost` resolves to that Mac rather than the relay. So the placeholder is correct today for HTTP and wrong only for TLS.

## Options

1. Interpolate only when TLS is on and the resolved host is not `localhost`; keep the placeholder otherwise.
2. Always keep the placeholder, and add a line under the TLS banner naming the host agents must use to satisfy the certificate.
3. Print both — the concrete host as the example, with a note that a remote agent needs a name the certificate covers.

Option 1 is the smallest, but it makes the banner's shape depend on configuration, which is worth weighing against option 2's constant shape.

## Definition of done

- [ ] A teammate can copy the agent-connect line from a TLS banner without producing a name mismatch.
- [ ] The HTTP path still directs a remote agent away from `localhost`.
- [ ] Both `start.ts` and `relay-start.ts` agree.

## Context

Split out of the review on #613 (see #293 for the banner work itself). Cosmetic — no functional impact on TLS termination or agent connectivity, since the operator can already work out the right host.

bferanmi806-sketch:
## Summary

Use the resolved certificate hostname in remote-agent connection instructions when TLS has a concrete non-localhost identity. Both `tapflow start` and `tapflow relay start` keep their existing placeholders for HTTP and TLS localhost-fallback cases, so a remote Mac is never instructed to connect to `localhost`.

Focused start/relay command tests pass, along with typecheck, lint, changeset checks, and pre-commit. The full CLI suite on Windows reports 277 passed with 25 pre-existing platform-specific failures.

Closes #627

## Checklist

* [x] Tests written and passing
* [x] No `any`
* [x] Interface changes land in `agent-core` first — not applicable; no interface changes
* [x] No sensitive info (tokens, paths, credentials)

## Related `.work/` docs

* `.work/2026-08-28-agent-connect-host-plan.md`
* `.work/reviews/fix__627-agent-connect-host.md`

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated TLS startup instructions to show the certificate-resolved hostname for remote agent connections.
  - Preserved placeholder host guidance for HTTP connections and local TLS setups.
  - Corrected connection hints in both standard and relay-only startup flows.
  - Improved hostname handling regardless of capitalization.

- **Documentation**
  - Added the fix to the unreleased changelog.

### DIFF
--- a/.changeset/brisk-agents-connect.md
+++ b/.changeset/brisk-agents-connect.md
@@ -1,0 +1,5 @@
+---
+"tapflow": patch
+---
+
+TLS startup instructions now use the certificate-resolved hostname for remote agent connections instead of leaving a misleading host placeholder.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **TLS startup instructions now use the certificate-resolved hostname for remote agent connections** ([#627](https://github.com/jo-duchan/tapflow/issues/627)).
+- **TLS startup instructions now use the certificate-resolved hostname for remote agent connections** ([#627](https://github.com/jo-duchan/tapflow/issues/627)). HTTP and local fallback paths intentionally retain the host placeholder because an agent running on another Mac must not be directed to `localhost`.
 
 ## [0.20.0] - 2026-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **TLS startup instructions now use the certificate-resolved hostname for remote agent connections** ([#627](https://github.com/jo-duchan/tapflow/issues/627)).
+
 ## [0.20.0] - 2026-08-28
 
 ### Breaking Changes
--- a/packages/cli/src/__tests__/commands/relay-start.test.ts
+++ b/packages/cli/src/__tests__/commands/relay-start.test.ts
@@ -27,6 +27,14 @@ import { RatholeTunnel } from '../../lib/rathole-tunnel.js'
 import { TailscaleTunnel } from '../../lib/tailscale-tunnel.js'
 import { cmdRelayStart } from '../../commands/relay-start.js'
 
+function agentConnectLine(output: string[]): string {
+  const start = output.findIndex((entry) => entry.includes('tapflow agent start --relay'))
+  expect(start).toBeGreaterThanOrEqual(0)
+  const end = output.findIndex((entry, index) => index >= start && entry.includes('--token <agent-PAT>'))
+  expect(end).toBeGreaterThanOrEqual(start)
+  return output.slice(start, end + 1).join(' ')
+}
+
 describe('cmdRelayStart', () => {
   let output: string[]
   let exitSpy: MockInstance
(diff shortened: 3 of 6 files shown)


bferanmi806-sketch:
@jo-duchan I enjoyed working through this one. If there are any other issues or roadmap items where an extra contributor would be useful, feel free to point me toward them. I’d be happy to keep contributing to Tapflow.

jo-duchan:
Welcome, and thank you — this is a well-put-together first PR.

Two things I want to name specifically, because they are the parts that are easy to get wrong and you did not. The `agentConnectLine` helper is what I noticed first: `banner()` wraps at 72 columns and emits each chunk as its own `console.log`, so an assertion against a single output entry would have been quietly checking half a line. Reassembling between the two markers is the right shape. You also got both halves of the release paperwork right — the changeset names `tapflow`, which is the CLI package, and the root `CHANGELOG.md` is hand-written and separately gated, so both were genuinely required rather than one being redundant.

What I verified locally:

- `pnpm --filter tapflow test` on macOS: 302 passed, 0 failed. That matches your 277 + 25 exactly, so those 25 are the platform-specific ones and none of them is yours.
- `typecheck`, `lint` and `pnpm changeset:check` all clean.
- Reverting the fix to a bare placeholder fails your two TLS tests; dropping only the `!== 'localhost'` half fails your two fallback tests. Both halves are load-bearing, which is the part I care about most — a test that passes either way would have looked identical in the diff.
- No `<host>` / `<this-ip>` site left unhandled, and the two commands agree. #627's definition of done is met.

One change before this goes in.

`displayHost !== 'localhost'` is case-sensitive, and the producer can hand you a spelling that is not lowercase. A certificate with no SAN at all falls through to the CN branch of `parseCertDisplayHostResult`, and that branch — unlike the SAN branch directly above it, which compares `host.toLowerCase() !== FALLBACK_HOST` — does no case folding at all. I built one rather than reasoning about it:

```sh
openssl req -x509 -newkey rsa:2048 -keyout k.pem -out c.pem -days 1 -nodes -subj "/CN=Localhost"
```

```
displayHost = "Localhost"
tls && displayHost !== 'localhost'  →  true

Connect Mac agents:  tapflow agent start --relay wss://Localhost:4000 --token <agent-PAT>
```

Which is the failure the placeholder exists to prevent, arriving through the door the new guard opens.

Either fix is one line, and I would take either:

- `displayHost.toLowerCase() !== 'localhost'` in both commands. Smaller, and stays inside the two files you are already touching.
- Fold case in relay's CN branch the way its SAN branch already does. Deeper, and it also repairs the `Relay :` line above, which advertises `https://Localhost:4000` today.

Pick whichever you would rather defend, and add the case to the tests you already have — mocking `resolveRelayDisplayHost` to return `'Localhost'` is all it takes.

One smaller thing while you are in there. The CHANGELOG entry is a headline with no body, and its neighbours in that file tend to say what the reader will see and why it is that way. A sentence noting that the HTTP path deliberately keeps the placeholder — because an agent on another Mac must not be sent to `localhost` — would carry the decision rather than only its outcome.

The red Vercel check is the fork-deployment authorization; it needs a team member to act, is not your diff, and does not gate this.

And yes — I would be glad to point you at more. Let's land this one first.

bferanmi806-sketch:
@jo-duchan updated this. The localhost fallback check is now case-insensitive in both paths, with regression coverage for both `localhost` and `Localhost`, and I clarified the fallback reasoning in the changelog.

Focused tests are 20/20 and 23/23, and typecheck, lint, changeset check, and pre-commit all pass. Full CLI on Windows is now 279 passed with the same 25 pre-existing Windows-specific failures.

Should be ready for another look when you get a chance.

jo-duchan:
The case fold is in both commands and the parameterised pair covers it. I ran your branch merged onto current main, which is 20 commits ahead of your base: 304 pass, and reverting `.toLowerCase()` fails your two `Localhost` cases and nothing else. The changelog sentence carries the decision, which the outcome alone did not.

Good to merge.

On more work: #661 asks for the relay suite to be made Windows-compatible, and you are the only person here running Windows — you have reported the same 25 failures on two PRs now. That is the one I would take, and it is `help wanted` already.

Extending the existing case with `it.each` instead of copying it is the right instinct, and you picked the changelog note up without being asked twice. A clean first contribution, and thank you for it.

bferanmi806-sketch:
@jo-duchan Thanks, I’m on it!

PR #708, opened 2026-08-30
Fix/mac resources a11y

### LINKED ISSUE #589: dashboard: Mac Resources and the comment panel carry a11y gaps the boot-fix pass did not close

Giannhskrg:
Fixes #589

Addresses the remaining accessibility findings in Mac Resources:
- Stable page-level h1 heading
- Selected Mac and online/offline state exposed to assistive technology via aria-current and a visually-hidden status word
- File-attachment validation errors announced (role="alert") and tied to the attach button via aria-describedby

Changeset added targeting @tapflowio/relay (ships via relay).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved screen reader support throughout Mac Resources, including clearer navigation state, online/offline status labels, and improved heading structure.
  * File attachment errors in the comment composer are now announced more clearly and associated with the upload control.
  * Attachment filenames are announced to assistive technologies.
* **Accessibility**
  * Mac Resources chart readings are now reachable using the keyboard.
  * Comment composer controls now have accessible names.

## Summary by CodeRabbit

- **Accessibility**
  - Improved keyboard and screen-reader access to Mac Resources navigation and agent selection.
  - Added a clear page heading and improved page structure for assistive technologies.
  - Added online/offline status announcements for selected Macs.
  - Improved comment attachment error announcements and associated errors with the attach button.

- **Bug Fixes**
  - Fixed accessibility issues in Mac Resources and the comment composer.


jo-duchan:
Thanks for picking this one up. All three findings in #589 are covered and the code itself is sound — I merged current `main` into your branch and ran it there: dashboard typecheck clean, lint 0 errors, 477 tests pass. Four things before it goes in.

**`aria-pressed` should be `aria-current`.** #589 asked for that and your description says it, but the code sets `aria-pressed`, which declares a toggle button, and clicking an already-selected Mac does not unselect it. The repo settled this once already: `components/device/shared/SimulatorToolbar.tsx:257` records why `aria-pressed` was tried and dropped there, and `IOSViewer.tsx:594` keeps it for a control that genuinely is one. So `aria-current={isSelected ? 'true' : undefined}`.

The changeset credits work that is not in this diff. It says chart readings are keyboard reachable, composer controls are named, and attachments announce their filename; those landed in the #586 follow-up, which is why #589 lists them as already fixed. That file becomes the published `@tapflowio/relay` changelog, so it should name what you did — a stable page-level `h1`, the Mac list's selection and online state exposed to assistive technology, and the attachment error announced and tied to the attach button. Your PR description needs the same trim.

The `changeset` job is red, and it reproduces locally:

```
Changeset added with no entry in the root CHANGELOG.md:

  .changeset/twelve-books-appear.md
```

Add a line under `## [Unreleased]` / `### Fixed` in the root `CHANGELOG.md`; CONTRIBUTING's "Changelog" section has the shape. The `<!-- changelog: internal -->` opt-out does not fit, since this is something a user can observe.

Also the branch is 230 commits behind `main`. It merges clean, so `git merge origin/main` is enough.

Three smaller ones, your call on each. There is no test, and #589 exists because the a11y gate reads whole files instead of the hunks a commit stages, so nothing stops the next edit to these two files from undoing this; `src/__tests__/` has the pattern and about 30 lines covers it — `getByRole('heading', { level: 1 })` with no Mac selected, `aria-current` on the selected entry, and `role="alert"` plus the attach button's `aria-describedby` in `CommentPanel`. One thing that will trip you up there: `MacResources.chart.test.tsx` renders `AreaChartInner` directly because `ParentSize` measures 0 in jsdom, but the no-selection state draws no chart, so rendering the page itself works fine. Second, the `sr-only` `h1` sits after the sidebar's `h2`, so a heading list reads "Macs" before the page title; moving it to the first child of the outer flex div fixes that, and `sr-only` is `position: absolute` so the layout does not move. Third, `CommentPanel.tsx:52` is indented three spaces where the file uses two, which ESLint does not catch.

Your PR body also dropped the repo template, the Checklist and the "Related `.work/` docs" section. Nothing to redo for this one, worth knowing for the next.

Naming `@tapflowio/relay` in the changeset instead of the private `@tapflowio/dashboard` is what first PRs here usually get wrong, and you had it without being told. You also took all three of CodeRabbit's notes and put the filename display back instead of leaving it dropped. Good first contribution.

Giannhskrg:
Hey thanks for the thorough review, it was really helpful. My commit should now be cleaned up.

jo-duchan:
Three of the four are in, and both of the optional ones as well. Moving the `sr-only` `h1` to the
first child of the outer flex div is the one worth calling out, because that is what puts the page
title ahead of the sidebar's `h2` in a heading list rather than merely present somewhere on the page.

The fourth is still open and it is now the only thing standing between this and a merge. The branch
is 248 commits behind and GitHub reports the PR as conflicting, so no check has run against your last
commit — the results showing on the PR are from the commit before it.

The conflict is a single file, and it is the one you just edited: `main` added a TLS entry under the
same `### Fixed` heading you added yours to. Both lines belong there, so it resolves by keeping both:

```
git fetch origin
git merge origin/main
```

I ran that locally to see what is underneath the conflict. Dashboard typecheck is clean, lint reports
0 errors, 477 tests pass, and `changeset:check` passes. `MacResources.tsx` and `CommentPanel.tsx` have
not been touched on `main` across those 248 commits, so none of your work has gone stale — the
changelog really is all of it.

`aria-current={isSelected ? 'true' : undefined}` is the exact form, and you rewrote the changeset to
describe this PR instead of leaving half of the old sentence in place. Both of those are easy to get
approximately right and you got them right.

Giannhskrg:
Heads up: merging in upstream/main surfaces a pre-existing typecheck failure,
unrelated to this PR:

  components/device/shared/SimulatorToolbar.tsx:3:41 - error TS2305:
  Module '"lucide-react"' has no exported member 'RadioOff'.

Confirmed by checking out bare upstream/main and running
`pnpm --filter @tapflowio/dashboard typecheck` directly — fails identically
with no changes from this branch involved. Likely from the lucide-react 1.x
migration in 0.20.0's CHANGELOG. Flagging since it'll block my CI even though
it's not something I touched.

jo-duchan:
That one is your `node_modules`, not `main`. Run `pnpm install` after the merge and it goes away.

`RadioOff` was added to lucide in 1.6.0. Your branch points at `^0.577.0`, `main` moved to `^1.33.0`
in `11df45be`, and `RadioOff` appears 5 times in 1.33.0's `.d.ts` and 0 times in 0.577.0's. Merging
brings the new `package.json` and the source that imports the icon, but installed packages do not
move with a merge, so TypeScript reads the old typings against the new source. I ran
`pnpm --filter @tapflowio/dashboard typecheck` on your branch merged onto current `main` and it is
clean, along with lint and 477 tests.

Checking out bare `upstream/main` reproduced it for the same reason — `node_modules` does not follow
a branch switch either, so both checkouts were reading the same 0.577.0 typings.

Your CI is unaffected whichever way this goes: the workflow installs from the lockfile on every run,
so it never sees a stale tree. `main`'s own CI is green on `e4f53342` from earlier today.

Going and reproducing it against a clean checkout before reporting is the right instinct, and it is
what makes this quick to answer — the conclusion was off but the method was not, and the trap it
walked into catches people who have been here much longer.

PR #709, opened 2026-08-30
test(relay): make suite Windows-compatible

### LINKED ISSUE #661: test(relay): make the full relay test suite Windows-compatible

bferanmi806-sketch:
## Summary

Make the relay test suite Windows-compatible without changing production key-storage behavior or ZIP extraction.

- Gate only the POSIX `0600` permission assertions on Windows while keeping account-key creation, persistence, reuse, and certificate save/load coverage everywhere.
- Replace system `zip` fixture creation with a self-contained Node ZIP helper that emits explicit root `.app/` directory entries, plist files, placeholder binaries, and no-app archives.
- Close each `builds.test.ts` SQLite handle before removing its temporary directory.

Closes #661

## Validation

Before, on unmodified upstream `main` at `aec46dbc174b6398686eba3b0de5b528535dc167`:

```text
Test Files  4 failed | 57 passed | 1 skipped (62)
Tests       5 failed | 699 passed | 1 skipped (705)
```

The five failed tests were the two POSIX mode assertions, two app ZIP metadata tests, and one no-app ZIP upload test; three `builds.test.ts` cleanup hooks also reported `EPERM`.

After, on final commit `a5dd2a12ed9777c85b150a8de360df3a84b3a101`:

```text
Test Files  61 passed | 1 skipped (62)
Tests       704 passed | 1 skipped (705)
```

The unchanged production `unzip` executable was already installed under Git for Windows but not on this PowerShell session's `PATH`, so it was appended only for the test process.

- Relay typecheck passed.
- Relay lint passed with 5 pre-existing warnings and 0 errors.
- Repository pre-commit checks passed.
- Independent adversarial review found no actionable issues.

## Checklist

- [x] No `any`
- [x] Interface changes land in `agent-core` first (not applicable)

## Related `.work/` docs

- `.work/reviews/test__661-windows-relay-suite.md`

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform test reliability, including Windows-compatible permission checks.
  * Standardized ZIP fixture creation across application, build, and upload-limit tests.
  * Added support for testing macOS application bundle ZIP fixtures.
  * Ensured temporary test databases are closed during cleanup.

### DIFF
--- a/packages/relay/src/__tests__/accountKey.test.ts
+++ b/packages/relay/src/__tests__/accountKey.test.ts
@@ -24,7 +24,10 @@ describe('loadOrCreateAccountKey', () => {
     const k1 = await loadOrCreateAccountKey(fp, create)
     expect(k1).toBe('ACCOUNT-KEY-1')
     expect(creates).toBe(1)
-    expect(fs.statSync(fp).mode & 0o777).toBe(0o600)
+    // Windows ACLs do not map to POSIX permission bits.
+    if (process.platform !== 'win32') {
+      expect(fs.statSync(fp).mode & 0o777).toBe(0o600)
+    }
 
     const k2 = await loadOrCreateAccountKey(fp, create)
     expect(k2).toBe('ACCOUNT-KEY-1') // 디스크에서 재사용
--- a/packages/relay/src/__tests__/builds.test.ts
+++ b/packages/relay/src/__tests__/builds.test.ts
@@ -2,21 +2,8 @@ import { describe, it, expect, beforeAll, afterAll, beforeEach, afterEach } from
 import fs from 'fs'
 import path from 'path'
 import os from 'os'
-import { spawnSync } from 'child_process'
 import { initDb, getDb, closeDb } from '../db'
-
-// ── fixture helpers ────────────────────────────────────────────────────────
-
-function makeTestZip(tmpDir: string, appName: string, plistXml: string): string {
-  const appDir = path.join(tmpDir, `${appName}.app`)
-  fs.mkdirSync(appDir, { recursive: true })
-  fs.writeFileSync(path.join(appDir, 'Info.plist'), plistXml)
-  // minimal binary placeholder (Mach-O magic is not needed for zip structure test)
-  fs.writeFileSync(path.join(appDir, appName), Buffer.from([0xcf, 0xfa, 0xed, 0xfe]))
-  const zipPath = path.join(tmpDir, `${appName}.app.zip`)
-  spawnSync('zip', ['-r', zipPath, `${appName}.app`], { cwd: tmpDir })
-  return zipPath
-}
+import { makeAppZip, writeZipFixture } from './helpers/zipFixture'
 
 const XML_PLIST = `<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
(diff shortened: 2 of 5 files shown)


bferanmi806-sketch:
@jo-duchan thanks again for pointing me to this one. Happy to keep helping out with Tapflow wherever useful.

jo-duchan:
Two small things, then this is ready.

`diskCertStore.test.ts`'s `key 파일은 0600 권한으로 저장한다` has no assertions left on Windows. Its only `expect` is inside the guard, so a Windows run reports a green test whose name still promises a 0600 check. Keep the guard — `jwtSecret.test.ts:44` does exactly this — but that one also puts `(POSIX 한정)` in the test name so the report says what it skipped, and that half is worth copying. `accountKey.test.ts` keeps other assertions in the same test, so it is less pressing.

`zipFixture.ts` should say why the `Foo.app/` entry in `makeAppZip` is there. `findAppDirInZip` matches a line of `unzip -l` output ending in `.app/`, so an empty directory entry that looks removable is load-bearing, and nobody editing the helper later can see that from the helper. A line on why this is hand-written instead of a dependency would help too, since the repo carries no ZIP library and that choice is invisible now.

I did not treat the green suite as proof the writer is correct, so I generated all three fixtures and ran `unzip -t` over them: no CRC or central-directory errors, and `unzip -l` output matches the regex including the `Food Truck.app` case. Re-ran the four files on current `main`, 58 pass.

Two things I found that are not yours to fix: there is no `windows-latest` job, so nothing stops the next PR undoing this with every check green, and production still shells out to `unzip` and `tar`, which is why you needed it on `PATH` at all. Those go in an issue under #661 rather than here.

Emitting the explicit directory entry means you went and read `findAppDirInZip` before writing the fixture, instead of producing a zip that merely opens. That is the reason this one verified in minutes.

bferanmi806-sketch:
@jo-duchan updated both points. I renamed the DiskCertStore permission test to make the POSIX-only scope explicit, and added comments documenting why the .app/ directory entry is required and why the ZIP fixture stays in-process.

Relay typecheck and lint are clean, and the focused tests that can run in this Windows environment pass. Thanks for the careful review.

jo-duchan:
The rename keeps the guard exactly where it was and changes only the name, so a Windows run now
reports what it skipped instead of a green test promising a 0600 check it never made. Renaming rather
than restructuring is the right call here: the guard was already correct and only the report was
misleading.

Both comments in `zipFixture.ts` are in, and they answer different questions — one says why the
directory entry cannot be deleted, the other why the writer is hand-rolled at all. The first names
`findAppDirInZip`, which is what gives the next person editing the helper somewhere to go.

Nothing else moved in this commit: two comments and one test name, so the fixtures still produce the
same bytes I checked with `unzip -t` last time. Checks are green and the branch is level with `main`.

Good to merge, and thank you for turning both of these around so quickly.

PR #720, opened 2026-09-01
fix(dashboard): keep the record button focusable while a recording is processed

### LINKED ISSUE #624: Two toolbar buttons drop keyboard focus by disabling themselves mid-request
Found by the a11y gate while adding the network control (#607). Split out because the fix changes how two existing controls behave, not just what they announce.

Two toolbar buttons disable themselves while their own request is in flight:

- `IOSViewer.tsx` — the software-keyboard toggle, `disabled={swKeyboardPending}`
- `SimulatorToolbar.tsx` — the record button, `disabled={recordState === 'uploading' || recordState === 'done'}`

A `disabled` element cannot hold focus, so the browser blurs the button the user is standing on and focus falls to `<body>`. A keyboard user loses their place mid-request and has to Tab back through the toolbar to get to it; on the record button that happens exactly when an upload starts, which is the longest of these waits.

The network control added alongside this stays enabled for the same in-flight state and guards inside its handler instead, so the two shapes now sit next to each other in one toolbar.

The fix is `aria-disabled` plus an early return in the handler, which keeps the element focusable and still refuses the click. What makes it a decision rather than a one-liner is that `aria-disabled` does not stop pointer events either — so each button needs its own answer to what a second click during the wait should do, and the record button's `done` state may want to stay genuinely disabled rather than merely announced as such.

daniarjabagin:
## Summary

Fixes #624

The record button used `disabled` while a recording was being processed (`uploading`) and once it was saved (`done`). A disabled element cannot hold focus, so activating "Stop recording" from the keyboard dropped focus to `<body>`, and the name change to "Processing the recording" / "Recording saved" was announced to nobody.

- `aria-disabled` + `aria-busy` instead of `disabled`, with the guard in the handler — the same shape as the restart and network controls in the toolbar
- A sibling `role="status"` region, mounted unconditionally, that says "Recording.", "Processing the recording." and "Recording saved."
- The two network tests that looked up the only `status` on the page now read the region through the button's `aria-describedby`

The software-keyboard half of #624 already landed in f2b9c4f7.

Changeset targets `@tapflowio/relay`, and there is a CHANGELOG entry under Unreleased / Fixed.

## Checklist

- [x] Tests written and passing (`SimulatorToolbar.record.test.tsx`; dashboard: 485 tests, typecheck and lint clean)
- [x] No `any`
- [x] Interface changes land in `agent-core` first — none

## Related `.work/` docs

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recording controls for keyboard and assistive technology users.
  * The record button now remains focusable while processing or saving.
  * Added status announcements for recording, processing, and successful saving.
  * Prevented accidental clicks while a recording is being processed or saved.
  * Improved visibility of the button’s busy state during recording processing.
  * Updated status messaging to clearly indicate when a recording is being processed.


daniarjabagin:
Left this one out on purpose: jsdom does not blur a focused element when it becomes `disabled`, so a focus-retention test passes against the old code too and cannot fail. The `hasAttribute('disabled')` assertion is what actually pins the fix.

jo-duchan:
Maintainer review. Merged onto `main` and ran it — `SimulatorToolbar` is 59/59 across three files, and it merges clean with no conflict.

**This is right, and it is right in the way the file already does it.** I checked the claim in your comment rather than taking it: the reboot control and the network control in the same component both use `aria-disabled` with the guard in `onClick`, and the network one carries the sentence *"Enforced here, not only announced here"* for the same reason yours does. You matched an existing pattern instead of inventing one, which is the part I most often have to ask for.

Three things I verified because they are the ways this shape usually goes wrong:

- **The live region is mounted unconditionally and only its text toggles.** A region that appears at the same moment as its content is frequently not announced at all. Your test pins this the hard way — `expect(statusOf(record())).toBe(region)` across every transition asserts it is the *same node*, not merely that some region has the right text. Most people check `textContent` and stop.
- **The absence assertion has a positive control.** `expect(onRecordToggle).not.toHaveBeenCalled()` for the inert states sits in the same block as `toHaveBeenCalledTimes(1)` for the live one, so a handler that stopped firing entirely fails the second half first. That is what [contributing/test-and-guard-coverage.md](https://github.com/jo-duchan/tapflow/blob/main/contributing/test-and-guard-coverage.md) is about, and it is the most common thing to get wrong here.
- **`done` is transient.** `useClientRecording` returns to `idle` after 2s, so this does not leave a permanently focusable button that does nothing.

`aria-disabled:opacity-50` is a Tailwind 3.4 built-in and `components/ui/sidebar.tsx` already uses it, so no config needed. It also goes slightly further than the two controls you followed — they show a spinner while pending and skip the dimming — and I think going further is right, because `done` has no spinner and would otherwise look fully enabled for two seconds.

---

### One suggestion

The bug was that focus fell to `<body>`. The test asserts that through a proxy:

```js
expect(button.hasAttribute('disabled'), 'a disabled button cannot hold focus').toBe(false)
```

That is true and it is one step removed. A direct assertion would survive a future change that keeps `disabled` off but drops focus some other way — a re-key, a conditional remount, a wrapper that unmounts the button between states:

```js
record().focus()
rerender(/* uploading */)
expect(record()).toHaveFocus()
```

Not blocking. It is the difference between testing the fix and testing the bug.

### One question

The button carries both `aria-describedby={recordStatusId}` and `role="status"` on the target. While the button is focused and the state changes, several screen readers will announce the live region *and* re-read the description. Was that deliberate?

I can see the argument for keeping both — the live region fires once, so the description is what a user who tabs to the button afterwards gets. If that was the reasoning it is worth a line in the comment, because the next person will read it as redundancy and remove one.

---

Nothing blocking. Say the word on the focus assertion and I will merge either way.

daniarjabagin:
Thanks for the careful read.

**Focus assertion:** added. The test now focuses the button while `recording`, moves it through `uploading` and `done`, and asserts `toHaveFocus()` after each. Kept the `disabled` check beside it, since that is the one that names the mechanism.

**`aria-describedby` + `role="status"`:** deliberate, for the reason you guessed. The region fires once at the transition; the description is what someone who tabs to the button afterwards gets, otherwise a `done` button with an idle-looking icon says only "Recording saved" as its name and nothing about what it would do. A double read while focused seemed the cheaper failure. Added a comment saying so.

60/60 locally, no lint changes.

jo-duchan:
Both done, and done better than I asked. Merged onto `main` and ran it: `SimulatorToolbar` is 60/60 green, and it merges clean.

**The focus test is now about the bug rather than about the fix.**

```js
record().focus()
rerender('uploading');  expect(record()).toHaveFocus()
rerender('done');       expect(record()).toHaveFocus()
```

Checking `disabled` was absent would have kept passing through a remount or a re-key that dropped focus some other way. This would not. That is the whole difference, and you took it across both transitions rather than just the first — `done` arrives asynchronously too, and it is the one I would have forgotten.

**And the `aria-describedby` answer is the part I want to point at.**

> Both on purpose: the live region fires once, when the state changes; the description is what a user who tabs to the button afterwards gets. Some screen readers read both while the button is focused, which is the cheaper failure.

That last clause is what makes it a decision instead of an oversight. You did not claim the duplicate announcement away — you said it happens and why you accepted it, so the next person reading this has to argue with a reason rather than delete something that looks redundant. Comments that survive contact with a future maintainer look exactly like that.

---

Nothing left from me. Approving; the maintainer will merge.

This is a good first contribution to land — you found the real cause (a disabled element cannot hold focus, so the accessible-name change was announced to nobody), you followed the pattern the file already used for its reboot and network controls instead of inventing one, and both rounds of review came back tighter than they went out. Thanks for the care.

Issue #734, opened 2026-09-02
The agent's heartbeat fallback trusts a world-writable /tmp path


jo-duchan:
Parent: #723

`SimulatorNetwork.readFilterState()` reads the provider's heartbeat from the first of two paths that
parses:

```
/Library/Application Support/tapflow/tapflow-netfilter-state.json
/tmp/tapflow-netfilter-state.json
```

The second is world-writable. Any local process can create it with a current timestamp and a rule
naming a device, and the agent will read it as the provider's own publication.

That path has existed since #639 and until now it only fed the liveness watcher, where a forged file
delays an `enforcement-lost` report. **#733 made it load-bearing for a decision**: the confirmation
that follows a rule write now falls back to this file when the XPC channel is gone, and a confirmed
write is what lets `setOfflineLocked` apply layers 2 and 3. So a forged file can put an "offline"
control over a simulator whose traffic is flowing — the sign-off failure the whole feature exists to
prevent.

It is only reachable while the protected file is **absent**, since the first path that parses wins.
That is the state of a Mac whose filter is stopped or was never installed, which is also the state
where the lie is most useful to tell.

### Why this is not a one-line fix

Three answers, and they are not equivalent:

- **Require root ownership of the fallback.** The provider is a system extension and runs as root, so
  `statSync(p).uid === 0` discriminates. But it applies to `readFilterState`, which the liveness path
  shares, and the tests inject their own paths — so the rule has to be scoped to the default fallback
  rather than to any configured path, which is a distinction worth stating deliberately.
- **Drop the fallback from the confirmation path only.** Three lines, and it leaves liveness as it is.
  But it removes the fallback exactly where the protected directory is unwritable, which is the case
  the fallback was added for.
- **Stop writing `/tmp` at all** and let an unwritable protected directory be a hard failure the
  provider reports. `Provider.swift` already logs `heartbeat path refused` for it, so the state is
  observable; the question is whether it should be fatal.

Choosing needs a view on how often the protected path is genuinely unwritable, which nothing here
measures.

### Acceptance

- A forged `/tmp` state file cannot make `applyAndConfirm` succeed, with a regression test.
- Whatever is decided applies to the liveness watcher as well, or the record says why it does not.

Raised by CodeRabbit on #733.

Issue #731, opened 2026-09-02
Confirmation deadline is sized from n=5, and migrate now exits non-zero on it


jo-duchan:
Parent: #725

`CONFIRM_DEADLINE_MS` is 30 seconds, sized from five observations recorded in `SimulatorNetwork.ts`: about 5.8s typically, with one run of five taking 21.3s. The 21.3s sample is a *fresh launch*, which is exactly the condition right after a replace. Thirty seconds is 1.4× the observed maximum from a sample of five, plus up to one polling interval of overshoot.

On a Mac with a heavier tail, a perfectly good install is reported as `installed-unconfirmed`, and `tapflow migrate net-filter` then exits non-zero on a success path — the first time that command does so. A provisioning script would fail there while passing on a machine stuck waiting for a System Settings toggle nobody will click.

Two decisions, both needing more than this PR had:

- **Resize the deadline from a real sample.** More than five runs, on a replace rather than a same-version reinstall.
- **Decide whether an unconfirmed result should be a non-zero exit at all.** It is one today because the state it names — configuration switched on, nothing answering — is the shape that took a Mac's network down on 2026-09-02. But `needs-approval`, where the filter is definitely off and a human must act, exits 0. The two are not obviously consistent, and the answer is a CLI contract judgement rather than a correctness one.

PR #743, opened 2026-09-03
fix(android-agent): record why the scrcpy server process exited

### LINKED ISSUE #481: android-agent: scrcpy server process exits with no recorded cause

daniarjabagin:
## Summary

Closes #481.

Picked this up from the good-first-issue list. The scrcpy server process only had an `error` handler, so if it died after a clean start the log just showed the video stream restarting and nothing about why. I added an `exit` handler next to it that logs the exit code (or the signal, when there is no code).

I kept it to logging only, as the issue suggests. `pumpVideo` already handles the restart through the socket closing, and I did not want a second trigger racing it.

For telling our own kill apart from a real crash I used the process reference: both `stop()` and the catch in `start()` set `this.serverProc = null` before the exit event can fire, so if the reference is already gone the exit is one we asked for and it goes to debug instead of warn. That way routine teardown does not print a warning every time.

Tests follow the fake-proc pattern already in `ScrcpySession.test.ts`: one for the exit code, one for the signal, and one for the exit after our own kill. The last one asserts `kill` was called first so the "no warning" check has something to fail against.

## Checklist

- [x] No `any`
- [ ] Interface changes land in `agent-core` first (no interface change here)

## Related `.work/` docs

None.

## Summary by CodeRabbit

- **Bug Fixes**
  - Android agent logs now identify why the scrcpy server stopped, including its exit code or signal.
  - Expected server shutdowns during session cleanup are logged at debug level instead of as warnings.
  - Unexpected server exits are clearly reported, helping distinguish device disconnects or terminated processes from unexplained restarts.


jo-duchan:
Thank you for this. **#630 has been open against #481 since 21 August**, and under [CONTRIBUTING](https://github.com/jo-duchan/tapflow/blob/main/CONTRIBUTING.md#claiming-an-issue) opening a PR is how an issue is claimed — so that one holds it and this will be closed.

Nothing on #481 said so in words. The claim was there only as a linked pull request in the sidebar, and the good-first-issue list that surfaces these shows a title and nothing about open work. **A workflow is going in that comments on an issue automatically the moment a PR links it**, so the next person reads it as a sentence instead of inferring it from the page furniture. That is the fix; this thread is what prompted it.

---

**I read your implementation before deciding, not just the timestamps.** Both PRs land on process identity to tell a kill we asked for from a server that died on its own, which is the right shape and not the obvious one — a boolean flag is what most first passes reach for. The difference is narrow: yours compares against the live reference (`this.serverProc === serverProc`), #630 records the process it killed (`this.stoppedProc`). Under a fast restart the live reference has already moved on, so a late exit from the old process gets judged against the new one; recording the killed process does not have that window. That is the only thing between them, and it happens to point the same way as the dates. Nothing here needs porting across, which I would have said if it did.

Your `exit` comment is better written than the one that is landing, for what it is worth — *"a second trigger here would race it into a double restart"* is the sentence that stops someone adding a restart call to that handler later.

You landed #720 today and it went in cleanly through two rounds. If you want another, these have no open PR against them and I can promise it because they are mine:

- **#731** — the confirmation deadline for `tapflow migrate net-filter` was sized from five observations, and the command now exits non-zero on a path that used to exit zero. It needs a real sample and a judgement about the exit code, which is a more interesting problem than it sounds.
- **#734** — the agent reads a heartbeat from a world-writable `/tmp` fallback and now makes a decision on it. Three possible fixes are written out in the issue; picking between them is the work.

daniarjabagin:
Understood, and thanks for reading it rather than just the dates. I looked at the issue's comments and assignee and not at the linked PRs in the sidebar, which is on me. The auto-comment will help.

Good point on the late exit under a fast restart; recording the killed process is the tighter shape.

I'll take a look at #734.

PR #745, opened 2026-09-03
fix(ios-agent): refuse a filter state file anyone on the Mac could have written

### LINKED ISSUE #734: The agent's heartbeat fallback trusts a world-writable /tmp path

daniarjabagin:
## Summary

Closes #734.

Took the first of the three options from the issue: a state file has to be root-owned to be believed. I keyed the rule on the directory rather than on the path, since that seemed the cleanest way to scope it. A file in a directory that anyone but its owner can write to is trusted only when root owns it and nobody else can write it. The protected path is unaffected because its directory is 0755, the `/tmp` fallback still exists for the case it was added for, and the tests keep injecting files into their private temp directories without needing a special case.

A few details worth a look:

- The check runs on the file descriptor that is then read, and the open uses `O_NOFOLLOW | O_NONBLOCK`, so what was judged is what gets believed, a symlink is refused outright, and a FIFO at the path returns immediately instead of blocking the thread.
- Liveness reads through the same method, so a forged file cannot hide a stopped provider either. That is the second acceptance point.
- A refused file is named once in the log, not on every 100ms poll.

Tests: a forged file in a world-writable directory is refused on the confirmation path with a positive control that differs only in the directory, one for the liveness watcher after the provider's file is removed, and one that races a FIFO at the fallback path against a timer. Under root the forged file is handed to `nobody` with `chownSync`, so the forgery cases run everywhere.

What I could not do here is measure how often the protected directory is genuinely unwritable, which the issue notes is what would decide between the options. This one keeps the fallback and makes it safe, so that measurement can come later without changing the shape.

## Checklist

- [x] No `any`
- [ ] Interface changes land in `agent-core` first (no interface change here)

## Related `.work/` docs

None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS simulator network-filter security by rejecting forged or improperly secured state files.
  * Prevented untrusted files in world- or group-writable directories from falsely indicating that network filtering is active or masking a stopped filter.
  * Added validation for file ownership, permissions, and file type before trusting filter state.
  * Added a warning when a state file is refused.


daniarjabagin:
Good call, that was a coverage gap rather than a limitation. The forged file is now chowned to `nobody` when the suite runs as root, with an assertion that it is not root-owned either way, and both tests run unconditionally.

jo-duchan:
One thing has to change before this can merge, and it is a flag.

`readIfProviderWrote` opens with `O_RDONLY | O_NOFOLLOW`, and opening a FIFO read-only blocks until a writer arrives. The `!file.isFile()` check that would reject it runs after the open, so it never gets there. Any local process can run `mkfifo /tmp/tapflow-netfilter-state.json`, and the state it needs is the one #734 already describes: a Mac where the protected file is absent because the filter is stopped or was never installed.

I reproduced it against the real class, with a FIFO as the second state file:

```ts
const open = join(dir, 'openfifo')
mkdirSync(open)
chmodSync(open, 0o1777)
const fifo = join(open, 'state.json')
execFileSync('/usr/bin/mkfifo', [fifo])
const net = make(undefined, 300, [join(dir, 'state.json'), fifo])
const raced = await Promise.race([
  net.setOffline(UDID, true).then(() => 'returned'),
  new Promise((r) => setTimeout(() => r('HUNG'), 4000)),
])
expect(raced, 'openSync on the FIFO blocked the event loop').toBe('returned')
```

It does not fail, it never returns. Neither the 4s race nor Vitest's own 20s timeout fires, because the synchronous open holds the worker thread and no JS timer can run; I killed it at 150s. In the agent that is the whole process, streaming and input and the relay connection with it. **Adding `constants.O_NONBLOCK` is the entire fix**: regular files are unaffected, the FIFO open returns immediately, and `isFile()` rejects it the way you meant it to. With that flag the test above passes and the suite is still 508/508.

`readFileSync(path)` blocked the same way before you touched it, so this is not something you introduced. I am asking for it here because this function is now where the read gets judged, and it already refuses non-regular files; only the ordering is wrong.

Four small ones. The directory check at `SimulatorNetwork.ts:1008` tests `mode & 0o002`, which lets a group-writable directory through, and `0o022` covers both bits; the comment above it states the rule as "a directory anyone can write to", so that sentence widens with the check. The same comment says a test's private temp directory is 0755, but `mkdtempSync` creates it 0700, and while the point holds either way the number does not. In `forgedState`, the message on `expect(statSync(path).uid, 'the forged file is owned by root').not.toBe(0)` says the opposite of what the assertion checks. And the description still says the two forgery tests skip under root, which `test(ios-agent): run the forged-state cases under root too` replaced with `chownSync`; the code is the better answer, so it is the description that needs to catch up.

The branch conflicts with `main` in `CHANGELOG.md` alone. The entry moves to the first `### Fixed` under `## [Unreleased]` as that section stands now.

I ran your branch before any of this: 508/508, `typecheck` clean, `lint` clean apart from one warning in a file you did not touch. Then I disabled the guard with `if (false)` and re-ran, and both new cases fail, so the positive control you paired them with is doing its job. The issue assumed the ownership rule would have to be scoped to the default fallback path, since the tests inject their own; keying on the directory mode means that problem never comes up. That reframing is the part I would not have arrived at, and it is why the diff is this small. Thanks for it.

daniarjabagin:
Thanks for reproducing the FIFO case, that would have been an ugly one to find in production. `O_NONBLOCK` is in, and your test is in too, with a comment saying why the flag is there so it does not get tidied away later. I confirmed it hangs without the flag before restoring it.

The four small ones are done as well: `0o022` on the directory check with the comment widened to match, `0700` for the temp directory, the assertion message now says what it checks, and the description below is updated for the `chownSync` approach.

Rebased onto `main`; the changelog entry sits under `[Unreleased]` → `### Fixed`. Local run: 508 pass plus the new FIFO case, one unrelated `.app.zip` failure that fails the same way on `main` here.

jo-duchan:
Three things left, and none of them is in the guard's logic.

The artifact failure I get here is not unrelated to the rebase, and it cannot happen on `main`. It is `netfilterArtifactFresh > is not lower than the one on main`: your tree carries extension version 1788368455 and `main` now carries 1788540210. That test compares the working tree against `origin/main`, so on `main` itself the two values are equal and it passes. `main` moved again after your rebase, because the netfilter work that landed since bumped `CFBundleVersion`. Your branch changes nothing under `ios-netfilter`, so rebasing takes `main`'s `shipped.json` whole, with no rebuild and no notarisation, and it settles the `CHANGELOG.md` conflict in the same pass.

The comment sentence needs one more go. It now reads "its directory is 0755 — and so is a test's private temp directory, 0700 from `mkdtemp`", where "and so is" points back at 0755 and the number then contradicts it. The property both directories share is that nobody but the owner can write to them, and saying that is shorter than the numbers anyway.

Then the directory widening went in with nothing measuring it. Reverting `0o022` to `0o002` on that check leaves **all 509 tests passing**. Neither shipped path can tell the two masks apart: `/Library/Application Support/tapflow` is 0755 and `/private/tmp` is 1777. The group bit only decides something for a directory like 0775, which is not an odd shape on macOS, where every local account is in `staff`. A temp directory chmodded to 0775 holding a file the suite wrote itself covers it, and that case goes back to failing if the mask ever narrows again.

Two older gaps in the same function are mine to close afterwards rather than yours. `O_NOFOLLOW` and the file-mode half of the condition both date from your first commit and survive being removed, and neither is something this round introduced. Nothing for you to do there.

Writing the assertion message as the failure it would describe, "so it would be trusted", is the form that argument wants, and it reads correctly from either end now. Thanks for turning these around as fast as you did.

daniarjabagin:
All three done.

Rebased onto `main` as it stands now, with nothing under `ios-netfilter` touched, so `shipped.json` is main's; `netfilterArtifactFresh` passes here along with the rest of the scripts suite. The changelog conflict settled in the same pass, entry first under `[Unreleased]`.

The 0775 case is in as its own test, with `forgedState` taking the mode. I checked it does what it is for: narrowing the mask back to `0o002` fails exactly that one test and nothing else.

The comment sentence now says the property instead of the numbers. Thanks for taking the two older gaps yourself.

jo-duchan:
The group-writable case is what I asked for, and I checked that it holds rather than that it runs: narrowing the mask back to `0o002` turns exactly one test red out of 510, and it is that one. That was the whole gap — the widening had nothing measuring it, so it could have gone back without a sound.

Your branch is seven commits behind `main` since #760 landed, but nothing conflicts.

Putting the reason inside the test is the part I would not have thought to ask for: **"it is the only case that fails if the mask narrows to `0o002`"** is the falsifiability argument written where the next person meets it, rather than in a review thread nobody reads twice. Parameterising `forgedState` with a mode instead of copying the fixture is what let the case be three lines.

Ready to merge from my side. Thanks for carrying this one through four rounds.

PR #772, opened 2026-09-09
fix(dashboard): unify unsupported decoder status

### LINKED ISSUE #748: The decoder-unsupported path explains itself in Korean on Android, and not at all on iOS
When no decoder can be picked, tapflow does not stream and there is nothing else it can do — so the message explaining that is the entire product experience for whoever lands there. Three things are wrong with it, all within a few lines of each other.

**The Android message is hardcoded Korean.** `AndroidViewer.tsx:588` renders "이 환경에서는 스트리밍을 표시할 수 없습니다. / Chrome/Edge 또는 HTTPS 환경에서 다시 시도해 주세요." to everyone. This is the widest of the three: it reaches every person on that path, not a subset, and an English-speaking self-hoster gets two lines they cannot read at the one moment the product needs to explain itself.

**iOS shows nothing at all.** `IOSViewer.tsx:125` is `onUnsupported: () => { /* iOS has no separate unsupported UI; the hook warns */ }`, and the warning is `console.warn` in `useDecoderStream.ts:73`. So the same environment gives an Android tester an explanation in the wrong language and an iOS tester a black rectangle with no explanation in any language. Whichever of the two behaviours is right, they should not disagree.

**Neither is announced.** The Android branch is a plain `<span>` — not a live region, no role, not focusable — and `SimulatorInfoCard`'s `role="status"` has no branch for it, because `getStatusText` returns `null` once the session is connected, joined, ready and installed, which it is: the session is healthy, only the picture is missing. A screen-reader user hears "Device screen" and nothing else. `role="alert"` on that div is the one-line answer and is not reliable — a live region generally has to exist before its content appears, and this element mounts with the message already in it. Routing `decoderUnsupported` through the status region, which is mounted early and `sr-only` while empty for exactly this shape of problem, is the fix that works.

**How reachable this is.** Both decoder tiers require WebGL2 (`pickDecoder.ts:47-49`), so it is not old browsers — WebGL2 has shipped since 2017, and Safari 15 in 2021. What is left is disabled hardware acceleration, a GPU on Chrome's driver blocklist, a VM without passthrough, and privacy-hardened configurations. Rare, but note that WebCodecs also needs a secure context, and tapflow's ordinary path is a LAN URL over plain HTTP — so on the path most teams use, WebGL2 is the only thing between a working stream and this screen.

Parent: #746

Farouk-Ben-Amor:
Closes #748

## What changed

- routes the unsupported-decoder state through the shared `SimulatorInfoCard`
- makes Android and iOS use the same English status message
- removes the Android-only Korean decoder message
- preserves the existing persistent `role="status"` region
- keeps higher-priority connection, boot, and install statuses ahead of the decoder status
-`break-all` was changed to `break-words` so normal status sentences wrap at word boundaries, while long unbroken error text can still wrap instead of overflowing.
- adds dedicated regression tests for `SimulatorInfoCard`

## Before

Android showed a hardcoded Korean unsupported-decoder message inside the device viewport, while iOS had no equivalent user-facing status.

**Android unsupported-decoder state:**

issue-748-before1 png

issue-748-before2 png

## After

The unsupported-decoder state is now surfaced through the shared status region used by both Android and iOS:

> Streaming is not supported in this environment.

**Shared status behavior:**

issue-748-after1 png

issue-748-after2 png

## Verification

- `pnpm --filter @tapflowio/dashboard test SimulatorInfoCard` — 3/3 passed
- `pnpm --filter @tapflowio/dashboard test` — 499/499 passed
- `pnpm --filter @tapflowio/dashboard typecheck` — passed
- `pnpm --filter @tapflowio/dashboard lint` — passed

The regression tests also verify their documented mutations:

- removing the decoder-unsupported status branch
- making the decoder message unconditional
- moving the decoder status above the connection-status precedence

## Bug Fixes

- Android and iOS viewers now consistently indicate when video streaming is unsupported in the current environment.
- Displays “Streaming is not supported in this environment.” in the shared status area.
- Replaces Android’s previous unsupported-stream message with consistent English text.
- iOS viewers now provide visible feedback instead of failing silently.
- Preserves the “Connecting…” status while a session is disconnected.
- Keeps the status area available for reliable viewer feedback.


jo-duchan:
Good first PR — thanks. Shared card, right register, decoder branch below connection/boot/install where it belongs.

You named a mutation per test, so I ran all three instead of taking the comments at their word. Each turns exactly one test red. I also made the `role="status"` wrapper conditional — the anti-pattern the issue names — and your second test caught it.

**Ignore the arm64 build failure, that one was ours.** `prebuild-install` reports a failed download as "no prebuilt binaries found", and there is a musl arm64 binary for the version we pin. Re-ran it, green.

The thing I'd want before merging is a test on the iOS side. Put your wiring back to `onUnsupported: () => {}` and all 499 dashboard tests still pass — the required prop only means *something* gets passed, and `false` compiles fine. `src/__tests__/DeviceViewer.network.test.tsx` is the harness for it: mocks `useRelay` and `pickDecoder`, renders `DeviceViewer` for real. It was itself made by copying `DeviceViewer.openUrl.test.tsx` and cutting what it didn't need.

Smaller: your second test rerenders `false → false`. Make it `false → true` and assert it's the same node. Right now a `key` on that wrapper that changes when the sentence appears passes all three tests, and that's the one thing the region gets mounted early to prevent.

Then the changeset — the one real CI failure. Careful with this one: a dashboard change names `@tapflowio/relay`, never `@tapflowio/dashboard`. The dashboard is private and ignored, but it ships inside the relay's `public/`. Naming both gets rejected at release time and nothing catches it earlier. Same job wants a line under `## [Unreleased]` in the root `CHANGELOG.md`.

One question: `break-all` → `break-words`. I think you're right, but it changes how a long `Install failed: …` wraps too — a sentence in the PR body would save the next person guessing.

Farouk-Ben-Amor:
Thanks @jo-duchan , I addressed the review points. 

- strengthened the persistent status-region test 
- added an iOS DeviceViewer wiring test 
- corrected the changeset to @tapflowio/relay 
- updated the PR body with the break-words rationale

jo-duchan:
Both of those close, and I checked rather than assumed: reverting `onUnsupported` to a no-op fails the new wiring test, and `key={statusText ?? 'empty'}` on the wrapper fails the strengthened one. 500/500 green, changeset names the relay.

Nice catch on `pickDecoder` — `network.test.tsx` only mocks `canDecodeH264`, which never reaches `ensureDecoder()`, and you found that instead of copying the harness and wondering why nothing fired.

I pushed one commit to your branch: the two new test files were missing a trailing newline (536 of the repo's 538 TS files have one, but nothing enforces it, so there was no way for you to know). Not worth another round on your side.

LGTM — merging once CI comes back.

Farouk-Ben-Amor:
Thanks @jo-duchan, glad the fix and the tests both held up. Nice to see it land in main.
looking forward to contributing more!

PR #777, opened 2026-09-10
fix(relay): use configured URL in reset emails

ML642:
## Summary

Use the configured public URL for member password-reset links instead of request headers, preventing Host-header-derived reset links.

## Checklist

- [x] No `any`

## Related `.work/` docs

- `2026-09-10-password-reset-public-url-plan.md`

## Summary by CodeRabbit

* **Bug Fixes**
  * Password-reset links now use the configured public URL instead of request-derived headers.
  * Links select the appropriate public address, including detected tunnel URLs, with a local fallback when needed.
  * A warning is logged when a password-reset link uses an insecure HTTP address.

* **Documentation**
  * Updated the unreleased changelog and release information to document the revised password-reset link behavior and URL selection order.

### DIFF
--- a/.changeset/secure-reset-links.md
+++ b/.changeset/secure-reset-links.md
@@ -2,4 +2,4 @@
 '@tapflowio/relay': patch
 ---
 
-Use the configured public URL for member password-reset links instead of request headers.
+Use the configured public URL for member password-reset links instead of request headers, and warn when the configured URL uses HTTP.
--- a/.changeset/secure-reset-links.md
+++ b/.changeset/secure-reset-links.md
@@ -1,0 +1,5 @@
+---
+'@tapflowio/relay': patch
+---
+
+Use the configured public URL for member password-reset links instead of request headers.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Member password-reset emails now use the configured public URL.** The reset endpoint previously constructed its link from the request's `Host` and `X-Forwarded-Proto` headers, which a proxy can pass through from an untrusted request. It now uses the same configured URL selection as invitations: public tunnel URL first, then the configured relay URL, then the local fallback.
+
 - **The dashboard is finally compressed on the deployment almost everyone runs** ([#260](https://github.com/jo-duchan/tapflow/issues/260), [#737](https://github.com/jo-duchan/tapflow/issues/737)). tapflow has been building Brotli copies of its dashboard assets for a while, and on a default install they were never sent. Browsers only offer Brotli on a secure origin, and the relay is plain HTTP unless you configure TLS — so over `http://<lan-box>:4000` the browser asks for gzip, the relay had only `.br` on disk, and every asset went out uncompressed. The build now writes `.gz` alongside `.br`, and the relay serves whichever the browser will take, preferring Brotli when both are offered. Nothing is compressed at request time: both files are built once, so this costs no CPU while you are streaming. Content-hashed files under `/assets/` are also marked `immutable` for a year, so a returning tab stops re-fetching them, while `index.html` is marked `no-cache` so an update is picked up on the next load rather than after a hard refresh.
 
 - **android-agent**: Log unexpected scrcpy server process exits with their exit code and signal as warnings, while expected exits from `stop()` log at debug level (#481).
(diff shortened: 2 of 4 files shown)


ML642:
This PR intentionally fixes a separate issue: Host Header Injection / password-reset poisoning. Reset links now derive their origin from trusted tapflow configuration rather than `Host` and `X-Forwarded-Proto`.

Requiring HTTPS here would fully address the cleartext-token risk, but it is a behavior change for existing self-hosted LAN installations, where HTTP is currently supported by default. The suggested patch also rejects HTTP unconditionally, despite the finding mentioning an explicit development-only policy.

I suggest tracking HTTPS-only password-reset links as a separate security migration:
1. warn when a reset link resolves to HTTP;
2. document HTTPS as required for secure email password resets;
3. introduce an explicit temporary insecure-development opt-in if needed;
4. make HTTPS mandatory in a future major release.

That keeps this PR focused on the confirmed Host-header vulnerability without silently breaking existing HTTP deployments.

jo-duchan:
Thanks for #777, and for keeping it to the Host-header fix when CodeRabbit pushed for HTTPS-only links. Enforcing that here would have broken every plain-HTTP LAN install, and your staged plan puts the warning before the breaking step, which is the right order. This is a heads-up, not a review yet.

#796 touches the same area and will likely merge first. It adds a check that allowlists the `Host` and `x-forwarded-proto` reads in `api/passwordReset.ts`, and it fails once your change removes them, with "no longer reads … (#777 has landed)". **Deleting that one `HEADER_ALLOWED` entry is the only change it asks of you.** CI tests the merge with `main`, so your next run after #796 merges will show it.

`buildInviteBaseUrl(cfg)` keeps its signature and result, so your import is unaffected. It also accepts an optional tunnel state now (`this.options.tunnel` in the relay). Passing it would let reset links use a Tailscale address the CLI detected, as invitations do after #794, but that is optional.

ML642:
fixes Host-header poisoning, but the HTTPS risk remains for deployments whose reset URL resolves to http:// (for example, a ws:// relay URL or local fallback). The reset token is still included in that URL and can be exposed in transit.
The current warning makes this visible but does not prevent it. HTTPS-only reset links should remain a separate, staged security migration to avoid breaking existing plain-HTTP LAN deployments.

jo-duchan:
Thanks for the update. The `HEADER_ALLOWED` entry is gone, and swapping the #777-specific stale test for a generic one plus a regression case for `passwordReset.ts` goes beyond what I asked.

**Needed before merge:** `CHANGELOG.md` now conflicts. `main` has its own `### Fixed` under `[Unreleased]`, so the entry belongs in that section rather than under a second heading.

**Up to you: passing the tunnel state.** I missed a case when I called it optional. With Tailscale, `tapflow start` leaves `tunnel.publicUrl` unset because the address is detected at runtime (#794). Without `this.options.tunnel`, reset links fall back to `relay.url` or `http://localhost:4000`, while invitations carry the Tailscale address. If you'd like to take it, it follows `handleInvite`:
- `RelayServer.ts`: `(req, res, params) => handleSendMemberReset(req, res, params, this.options.tunnel)`
- pass `tunnel?: TunnelRuntime` through `sendPasswordResetEmail` into `buildInviteBaseUrl(cfg, tunnel)`

A case next to "uses the tunnel URL the CLI detected (#794)" in `teammateUrls.server.test.ts` would pin it. If you'd rather not, just say so and I'll push it onto this branch before merging.

Minor: `judgeRelayFile` still appends ` (#777 has landed)` for `passwordReset.ts`, which can't happen anymore, so feel free to drop it. CodeRabbit's note on `code()` is about code from #796, and the file header explains that trade-off, so you can leave it.

ML642:
I took the runtime-tunnel follow-up in 123a2a60. Reset links now receive the RelayServer runtime tunnel state, and the new regression test verifies that a detected Tailscale URL is used for reset emails. I also removed the obsolete `(#777 has landed)` suffix.


The additional static-check changes only make its existing path and comment handling work on Windows/CRLF files; the original policy remains covered by regression tests.

jo-duchan (approved):
Checked 123a2a60 on top of current main. Dropping the tunnel from the route, or from `sendPasswordResetEmail`, fails only your new send-reset case, and the link falls back to `http://localhost:4000`. Putting the old `.*$` back in `code()` fails only the CRLF line you added.

The Windows path and CRLF fixes are fine to keep here. They're small, and this PR already touches that check.

Approving. You took the tunnel follow-up I had left open, and the Windows fixes came with a test of their own. Thanks for sticking with this one through the extra rounds.

PR #811, opened 2026-09-17
fix(relay,cli): stop treating tunnel traffic as local

jo-duchan:
## Summary

The relay does not ask connections that reach it over loopback to sign in, so the co-located agent and the CLI need no token. A tunnel client runs there too: rathole forwarded every visitor of the public URL to `127.0.0.1:<relay port>`, and `tailscale serve 4000` did the same for tailnet members, so those visitors inherited the exemption. `TAPFLOW_TRUSTED_PROXIES` could not close it, because rathole forwards raw TCP and adds no header.

The relay now opens a loopback-only **tunnel port** (`TAPFLOW_TUNNEL_PORT`, 4001 by default) when a tunnel is configured or the port is named. Requests accepted there are marked, and `resolveClientAddress` takes a required `viaTunnel` flag that makes them remote whatever their address. The CLI points rathole at that port, refuses a port it cannot use before rathole touches the VPS, stops rathole clients left behind by an exited tapflow process, and warns about the two Tailscale shapes no listener can fix. Reasoning is beside the code; the release note is in the changeset and the root `CHANGELOG.md`, including the `Migrate:` line for anyone serving `4000`.

## Checklist

- [x] No `any`
- [x] Interface changes land in `agent-core` first (n/a — no `agent-core` interface changed)

## Related `.work/` docs

`2026-09-18-tunnel-ingress-not-local-plan.md` — plan, invariant table, and the design review's dispositions.
`reviews/fix__tunnel-ingress-not-local.md` — the review record: one design channel before the code and two independent channels on the diff, 12 findings, all fixed, none deferred; 30 mutations run and all caught.

https://claude.ai/code/session_019C9Dy1tpbYA5BszS5QA6eQ

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated, loopback-only tunnel port for tunnel and proxy traffic.
  * The tunnel port defaults to `4001`, or `4002` when the relay uses `4001`.
  * Added port validation, collision detection, startup guidance, status warnings, and stale-client cleanup.

* **Security**
  * Tunnel traffic and remote connections require authentication.
  * Initial administrator setup is restricted to local relay access.

* **Documentation**
  * Updated self-hosting, configuration, security, and tunnel deployment guidance.


jo-duchan:
Five of the six are in `269976fd`.

The orphan check now needs both `node` and `tapflow` in the owner's command rather than either one. Measured on one Mac with 896 processes: 8 rows match either word, 1 matches both. The errors are not symmetric either — a client wrongly kept alive goes on forwarding to the relay port and says nothing, while one wrongly killed drops a tunnel loudly. Cleanup also runs from `setupServer` now, with a test that fails the binary download there and still expects the orphan stopped. The loopback wording, the 4002 fallback and `local.tunnelPort` went into the EN and KO pages together.

`isPortFree` I left as it is. Only `EADDRINUSE` counts as taken there on purpose, so `EACCES` on a privileged port reaches the operator in the relay's own words rather than as "already in use", and changing it changes what the relay port reports too. The case that reaches it is a tunnel port set below 1024, and it costs one VPS-side restart before the refusal.

PR #815, opened 2026-09-19
test(relay): guard Windows compatibility with CI job and archive prereqs

### LINKED ISSUE #710: ci(relay): the Windows compatibility from #709 is unguarded, and the suite still needs unzip on PATH
## What

#709 made the relay suite pass on Windows: the POSIX-only mode assertions are gated, ZIP fixtures come from a Node helper instead of a system `zip`, and each `builds.test.ts` SQLite handle closes before its temporary directory goes away. Two things it deliberately left alone, both raised while reviewing it.

**Nothing guards it.** The `test` job in `.github/workflows/ci.yml` is `runs-on: ubuntu-latest` with a matrix over Node 22 and 24 only, and no workflow in the repo names a Windows runner. So the next PR can reintroduce a `spawnSync('zip', …)` fixture or an ungated `0o600` assertion with every required check still green. The fix has the same property the bug had — it is invisible from CI, which is why #661 arrived as a bug report from a contributor rather than from the suite.

**The suite still needs `unzip` on `PATH`.** Production is unchanged and should stay that way here: `findAppDirInZip` calls `unzip -l` and `extractAppZipInfo` calls `unzip -p` (`packages/relay/src/api/builds.ts:89`, `:106`), and the tar path calls `tar -tzf` / `tar -xzOf`. #709's author had those binaries from Git for Windows but had to append them to `PATH` for the test process, and said so in the PR. A Windows contributor who does not know that gets `extractAppZipInfo` returning `null` and a failure that does not name the cause.

## What would close it

Two pieces, and they only work together.

1. Windows in CI. Either a `windows-latest` entry in the `test` matrix or, more likely the right size, one narrow job running `pnpm --filter @tapflowio/relay test`. This is the part that needs a decision rather than an edit: a full matrix leg doubles the test job, and either shape needs `unzip` and `tar` resolvable on the runner before it can be green, because the production dependency above applies there too.

2. Say what Windows needs. A prerequisites line in CONTRIBUTING, or an explicit skip with a message when `unzip` does not resolve, so the failure names what to install instead of looking like a broken test.

Adding 2 alone leaves the compatibility unguarded. Adding 1 alone makes it red on the first run.

Parent: #661

bferanmi806-sketch:
## Summary

Guard the #709 Windows relay compatibility in CI. Adds a narrow `windows-latest` relay job (`pnpm install --frozen-lockfile` to `pnpm --filter @tapflowio/relay test`, no build steps), makes `unzip`/`tar` resolvable to native processes while keeping System32 `tar` precedence, and wires the job into the aggregate `ci` gate as required-`success`. Adds a cross-platform archive-tool prerequisite that distinguishes missing executables from ones failing their version check, plus Windows prerequisites in CONTRIBUTING. Production extraction unchanged. Follow-up: adds `timeout-minutes: 15` to `test-relay-windows` and mocked `archivePrereqs` tests.

Closes #710. Adversarial review: `.work/reviews/feature__710-windows-relay-ci.md` (all findings cleared; the one minor finding deferred under code-freeze is now implemented, with non-ENOENT spawn errors reported as broken).

Validation (Windows): relay suite `Test Files 67 passed | 1 skipped (68)`, `Tests 817 passed | 1 skipped (818)`; missing `unzip` fails fast naming the tool (`spawnSync unzip ENOENT`); resolving-but-broken `unzip` reports the version-check failure distinctly (exit status); executables `unzip.exe` from Git `usr/bin`, `tar.exe` from `C:\WINDOWS\system32`; relay lint 0 errors (5 pre-existing warnings); relay typecheck pass. Follow-up validation: new `archivePrereqs.test.ts` 4 passed; targeted relay files 5 passed (`59 passed | 1 skipped`); full repo typecheck and lint via pre-commit hooks pass with 0 errors.

## Checklist

- [x] No `any`
- [x] Interface changes land in `agent-core` first (not applicable)

## Related `.work/` docs

- `.work/reviews/feature__710-windows-relay-ci.md`

## Summary by CodeRabbit

- **Tests**
  - Added required Windows coverage for the relay test suite in continuous integration.
  - Archive tests now verify required tools are available before running and provide clearer, platform-specific error messages when tools are missing or fail.
  - Improved cleanup handling so setup failures do not cause additional errors.
  - Archive creation failures now include actionable diagnostics.
  - The download-abandonment leak test remains limited to POSIX platforms.

- **Documentation**
  - Added Windows setup guidance for required archive utilities and environment configuration.


jo-duchan:
The `ci` wiring is right, and it's the part I expected to go wrong. `test-relay-windows` is in `needs` and asserted in the script body. The comment also explains why it doesn't get the skipped exception that `test-swift` does. That's what makes it a gate rather than just another job.

Two things before this goes in.

Give the new job a `timeout-minutes`. `test-swift` already has `timeout-minutes: 15` because the default is 360 minutes, and a hung run would hold the required check for six hours. The same applies here, especially since this suite spawns `unzip` and `tar`. One line.

What was the finding you skipped? The body says one was set aside "under PR code-freeze". `.work/` is gitignored, so I can't read the record. If it's small, it probably belongs here.

Optional: `assertArchiveTools` has two branches and no test of its own. Neither branch can be reached on CI, because the workflow step has already proved both tools resolve and run, so the manual check you mention is their only exercise. Mocking `child_process` for it should be about fifteen lines. `contributing/test-and-guard-coverage.md` is why I mention it.

I also checked that the job isn't green for the wrong reason: 816 tests across 67 files ran on the runner, with `unzip` resolving to Git's and `tar` to System32, which is what the PATH ordering was for.

One finding is worth calling out: the skip you added to `buildDownload.test.ts`. That test landed on 2026-09-11, eleven days after #709 merged, and reads `/dev/fd`, a POSIX-only assertion that drifted back in while every required check stayed green. That's the exact problem #710 was written to catch, and this change found it.

bferanmi806-sketch:
Done, and thanks for checking the gate logic.

Timeout: added `timeout-minutes: 15` to `test-relay-windows` in `.github/workflows/ci.yml`, with a short comment noting the 360 minute default and the required check. No other workflow logic changed. The `ci` gate, `needs`, and the no skipped exception for the Windows job are untouched.

Skipped finding: the deferred minor finding was the CodeRabbit note on `archivePrereqs.ts` to report only `ENOENT` as missing and other spawn errors as broken, including startup failures in the broken tool message. That is already implemented in this branch in `698790dd` (`probeTool` checks `error.code`, and the message reads `executables that fail to start or fail their version check`). I verified it against the current code and left the logic unchanged because it matches the review exactly. No `.work` file is present locally since that directory is gitignored, so this is based on the inline finding and the commit diff.

Mocked tests: added `packages/relay/src/__tests__/archivePrereqs.test.ts` mocking `child_process.spawnSync`. It covers success for both tools, `ENOENT` reported as missing on PATH, `EACCES` reported as broken and not missing, and nonzero version exit reported as broken with the exit status. Four tests, all passing.

Preserved: Windows CI gate wiring, PATH append ordering with System32 `tar` precedence, the `buildDownload` Windows skip, and the existing archive prerequisite fixes.

Validation on Windows host:
- New file: 4 passed.
- Targeted relay files (`builds`, `builds-tar-upload`, `upload-limits`, `buildDownload`, `archivePrereqs`): 5 files passed, 59 passed and 1 skipped (the Windows skipped fd test).
- Without Git `usr/bin` on PATH, the suite fails fast with `missing executables (not found on PATH): unzip (spawnSync unzip ENOENT)`, which is the intended message.
- With Git `usr/bin` appended, `unzip.exe` resolves there and `tar.exe` stays in System32.
- Relay lint: 0 errors (5 pre-existing warnings). Relay typecheck: pass. Full repo typecheck and lint via pre-commit hooks: pass with 0 errors.

Commit `fda394a5` is pushed to this PR branch.

jo-duchan:
This closes the half of #710 that needed a runner. The relay suite runs on `windows-latest` in 1m12s and `ci` requires it as `success`, so a reintroduced `spawnSync('zip')` fixture or an ungated POSIX assertion now has a leg that goes red on it.

I checked the one place Windows coverage narrows: the abandoned-download leak test skips there, and the mutation it exists for — dropping `res.on('close', () => rs.destroy())` — is still caught on the POSIX legs, so what the skip gives up is the fd-counting technique. I also checked that every suite touching an archive calls `assertArchiveTools`, not only the two the diff opens with: `builds`, `builds-tar-upload` and `upload-limits`.

Splitting ENOENT from the other spawn failures is past what #710 asked for and it is the right line — "not on PATH" and "resolves but will not start" send you to different places, and the message says which one you have. Guarding `afterAll` against a `beforeAll` that can now throw before the temp dirs exist is the second-order one; that normally surfaces weeks later as a teardown error nobody can place.

Ready to merge. Thanks — this is the follow-up I was hoping for when #709 landed.

